PR #1, opened 2021-11-17
Charmcraft init and added install functionality

MiaAltieri:
Did a charmcraft init and modified it to match structure of a machine charm
Removed actions.yaml & config.yaml since these are not implemented yet
Implemented on Install hook


jnsgruk (commented):
Looking good, thanks

src/charm.py
```diff
@@ -0,0 +1,74 @@
...
+    def _on_install(self, _):
+        """Install prerequisites for the application"""
+        # Install MongoDB
+        self.unit.status = MaintenanceStatus("installing mongodb")
+        # Get GPG key and add list file
+        gpg_url = "https://www.mongodb.org/static/pgp/server-5.0.asc"
+        repo = "https://repo.mongodb.org/apt/ubuntu"
+        archs = ["amd64", "arm64"]
+        lsb_release = "focal/mongodb-org/5.0"
+        software_archive = "multiverse"
```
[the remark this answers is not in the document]
MiaAltieri: I think if we do this then we will no longer need to parameterise the function and would no longer need to abstract the repo line ie.
```
            archs = "".join(["[ arch=", ",".join(archs), " ]"])
            line = " ".join(["deb", archs, repo, lsb_release, software_archive])
            repo = apt.DebianRepository.from_repo_line(line)
```
and would just be:
```
            line = " ".join(["deb [ arch=amd64,arm64 ]",repo," focal/mongodb-org/5.0 muliverse"])
            repo = apt.DebianRepository.from_repo_line(line)
```
and if we change this function to be called `_add_mongodb_org_repo` we might as well remove the constant and change it to:
```
            line = " ".join(["deb [ arch=amd64,arm64 ] https://repo.mongodb.org/apt/ubuntu focal/mongodb-org/5.0 muliverse"])
            repo = apt.DebianRepository.from_repo_line(line)
```
Should I go down that road or keep this function generalizable? I can't think of any other repos that would need to be added, so maybe making this function to be only for MongoDB would be a good idea. What do you think? 🙂

src/charm.py
```diff
@@ -0,0 +1,74 @@
...
+        self._get_packagage_external_repository(
+            gpg_url, repo, archs, lsb_release, software_archive
+        )
+        # install mongodb
+        self._install_apt_packages(["mongodb-org"])
+
+    def _get_packagage_external_repository(
```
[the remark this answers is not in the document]
MiaAltieri: see above comment 👆

marceloneppel (commented):
Great work, Mia! I've commented on just two changes that tox can alert you about the copyright notice. Other than that, LGTM.

src/charm.py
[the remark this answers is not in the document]
jnsgruk: Yeh, I'd just go with the simplest route for now and use `from_repo_line`

src/charm.py
```diff
@@ -0,0 +1,74 @@
+#!/usr/bin/env python3
+# Copyright 2021 Ubuntu
```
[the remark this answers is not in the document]
jnsgruk: In the same vain, can we make sure that the LICENSE header is present in all the files?

metadata.yaml
```diff
@@ -0,0 +1,15 @@
...
+name: mongodb
+display-name: MongoDB
+description: |
+  MongoDB is a general purpose distributed document database. This charm
+  deploys and operators MongoDB.
+summary: A MongoDB operator charm
+series:
+  - focal
+
+resources:
```
jnsgruk, line 15:
```suggestion
```

metadata.yaml
```diff
@@ -0,0 +1,15 @@
+# Copyright 2021 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+# For a complete list of supported options, see:
+# https://discourse.charmhub.io/t/charm-metadata-v2/3674/15
+name: mongodb
+display-name: MongoDB
+description: |
+  MongoDB is a general purpose distributed document database. This charm
+  deploys and operators MongoDB.
```
jnsgruk, line 10:
```suggestion
  deploys and operates MongoDB.
```

metadata.yaml
```diff
@@ -0,0 +1,15 @@
+# Copyright 2021 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+# For a complete list of supported options, see:
+# https://discourse.charmhub.io/t/charm-metadata-v2/3674/15
```
jnsgruk, line 5:
```suggestion
```

src/charm.py
```diff
@@ -0,0 +1,64 @@
+#!/usr/bin/env python3
+# Copyright 2021 Canonical Ltd.
+# See LICENSE file for licensing details.
+#
+# Learn more at: https://juju.is/docs/sdk
```
jnsgruk, line 5:
```suggestion
```

jnsgruk (commented):
Looking better already ;-) Couple of nits.

src/charm.py
```diff
@@ -0,0 +1,64 @@
...
+        """Install prerequisites for the application"""
+        self.unit.status = MaintenanceStatus("installing mongodb")
+        self._add_mongodb_org_repository()
+        self._install_apt_packages(["mongodb-org"])
+
+    def _add_mongodb_org_repository(self):
+        repositories = apt.RepositoryMapping()
+
+        # Get GPG key
+        key = urlopen("https://www.mongodb.org/static/pgp/server-5.0.asc").read().decode()
```
DnPlas, line 38: We may want some exception handling here

src/charm.py
```diff
@@ -0,0 +1,64 @@
...
+
+        # Add the repository if it doesn't already exist
+        if repo not in repositories:
```
DnPlas, line 41: I'm not sure about this condition, is `repo` declared somewhere else?

src/charm.py
```diff
@@ -0,0 +1,64 @@
...
+
+    def __init__(self, *args):
+        super().__init__(*args)
+        self.framework.observe(self.on.install, self._on_install)
+
+    def _on_install(self, _):
+        """Install prerequisites for the application"""
+        self.unit.status = MaintenanceStatus("installing mongodb")
+        self._add_mongodb_org_repository()
+        self._install_apt_packages(["mongodb-org"])
```
DnPlas, line 32: AFAIK, nothing sets `ActiveStatus` unless you do it explicitly. It could be in the event handler itself, otherwise the unit will remain in MaintenanceStatus.

src/charm.py
```diff
@@ -0,0 +1,64 @@
...
+            repo = apt.DebianRepository.from_repo_line(line)
+            # Import the repository's key
+            repo.import_key(key)
+            repositories.add(repo)
+
+    def _install_apt_packages(self, packages: list):
+        """Simple wrapper around 'apt-get install -y"""
+        try:
+            logger.debug("updating apt cache")
+            apt.update()
```
DnPlas, line 52: `apt.update()` could also raise an exception, we may want to catch that as well.

src/charm.py
DnPlas: I see you have `MONGO_GPG_URL`, maybe you can use it here or not have it declared at all, the linter will complain about that

src/charm.py
```diff
@@ -0,0 +1,64 @@
...
+
+    def _add_mongodb_org_repository(self):
+        repositories = apt.RepositoryMapping()
+
+        # Get GPG key
+        key = urlopen("https://www.mongodb.org/static/pgp/server-5.0.asc").read().decode()
+
+        # Add the repository if it doesn't already exist
+        if repo not in repositories:
+            line = "deb [ arch=amd64,arm64 ] https://repo.mongodb.org/apt/ubuntu focal/mongodb-org/5.0 muliverse"
```
DnPlas, line 42: You could use `MONGO_REPO` here, same case as my comment about `MONGO_GPG_URL`.

DnPlas (commented):
Thanks @MiaAltieri , just few comments.

src/charm.py
MiaAltieri: Hi! I looked at the code for `apt.update()` and it doesn't raise any exceptions in that function:

But  according to the documentation `check call` can raise the following exceptions: `TimeoutExpired`, `CalledProcessError`, and `SubprocessError` I'll catch these, unless you had something in mind :)

src/charm.py
MiaAltieri: 🤦‍♀️ thanks for catching this, fixed some comments and meant to check that the charm still built before EOD yesterday

src/charm.py
MiaAltieri: **Great Catch!** I'll put the charm in `WaitingStatus` since it isn't won't have started yet.

src/charm.py
MiaAltieri: I'll adopt your earlier sugggestion of deleting them

src/charm.py
```diff
@@ -0,0 +1,82 @@
...
+        # Get GPG key
+        try:
+            key = urlopen("https://www.mongodb.org/static/pgp/server-5.0.asc").read().decode()
+        except Exception as e:
+            logger.exception(f"failed to get GPG key", exc_info=e)
+            self.unit.status = BlockedStatus("Failed to get GPG key")
+            return
+
+        # Add the repository if it doesn't already exist
+        if "https://repo.mongodb.org/apt/ubuntu" not in repositories:
```
jnsgruk, line 43: Oh just one thing -- check that this is the right format... should this not be: `repo-mongodb-org-apt-ubuntu` or similar? You'll need to check what's in the repository mapping to be sure :)

DnPlas:
LGTM, once we resolve @jnsgruk 's conversations we can merge it.

tox.ini
```diff
@@ -0,0 +1,73 @@
...
+    coverage run --source={[vars]all_path} \
+        -m pytest --ignore={[vars]tst_path}integration -v --tb native -s {posargs}
+    coverage report
+
+[testenv:integration]
+description = Run integration tests
+deps =
+    pytest
+    juju
+    pytest-operator.git
```
marceloneppel, line 70:
```suggestion
    pytest-operator
```

marceloneppel, line 70: This can be changed to `pytest-operator` in a future PR (there is a recent change in the template to fix an error: https://github.com/canonical/operator-template/pull/6).

tox.ini
```diff
@@ -0,0 +1,73 @@
...
+    black --check --diff {[vars]all_path}
+
+[testenv:unit]
+description = Run unit tests
+deps =
+    pytest
+    coverage[toml]
+    -r{toxinidir}/requirements.txt
+commands =
+    coverage run --source={[vars]all_path} \
```
marceloneppel, line 61: Here you can change in a future PR to `coverage run --source={[vars]src_path} \`. It's another change that was made in the PR of the other comment.

marceloneppel (approved):
LGTM. There are two comments, but they can be addressed in a future PR.

PR #5, opened 2022-01-04
Added integration tests for charm start

MiaAltieri:
tests/integration/helpers.py from Marcelo's [postgres operator](https://github.com/canonical/postgresql-operator/blob/main/tests/integration/helpers.py)


tests/integration/test_charm.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+@pytest.mark.abort_on_fail
+async def test_status(ops_test: OpsTest) -> None:
+    """Verifies that the application and unit are active"""
+    await ops_test.model.wait_for_idle(
+        apps=[APP_NAME], status="active", timeout=1000
+    )
+    assert len(ops_test.model.applications[APP_NAME].units) == 1
+    assert ops_test.model.applications[APP_NAME].units[0].workload_status == "active"
```
WRFitch, line 35: You shouldn't need this line, since you're already checking the status is active in line 32.

tests/integration/test_charm.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+@pytest.mark.abort_on_fail
```
WRFitch, line 38: Generally I'd avoid abort_on_fail for every test that doesn't test essential functionality - it's necessary for building and deploying charms, but not for checking config files are correct, for example. 

My reasoning is that it's important to understand every failure, but when we know the essentials are broken, we can assume everything else is broken as well, and therefore we skip the less important tests. If the essentials are fine, we run all the tests so the developer has information about what's working and what isn't, and can make the necessary fixes without having to run these tests so often.

tests/data/mongod.conf
```diff
@@ -0,0 +1,15 @@
+net:
+  bindIp: localhost,10.126.63.237
```
WRFitch, line 2: I can't see this IP defined anywhere else - is this related to your local config?

tests/integration/test_charm.py
```diff
@@ -0,0 +1,88 @@
...
+            uri += ","
+        unit = ops_test.model.applications[APP_NAME].units[i]
+        uri += "{}:{}".format(unit.public_address, 27017)
+    uri += "/?replicaSet=rs0"
+
+    # connect to mongo replicaSet
+    client = MongoClient(uri)
+
+    # check mongo replicaset is ready
+    assert client.server_info()
```
marceloneppel, line 73: Maybe you can check the return against an expected content here too, to make it clear what is being checked (I believe it's a string).

tests/integration/test_charm.py
```diff
@@ -0,0 +1,88 @@
...
+        )
+        assert expected_mongodb_conf == unit_mongodb_conf_data
+
+
+@pytest.mark.abort_on_fail
+async def test_database_is_up_as_replica_set(ops_test: OpsTest) -> None:
+    """Tests that mongodb is running as a replica set for the application units"""
+    # construct URI for MongoDB replica set
+    uri = "mongodb://"
+    for i in range(0, len(ops_test.model.applications[APP_NAME].units)):
```
marceloneppel, line 62: Do you plan to deploy more than one unit in the integration tests right now in order to check the correct url construction and connection? I asked just because of this `for` logic. I understand that maybe you plan to implement it in a future PR.

tests/integration/test_charm.py
MiaAltieri: This makes a lot of sense to me!  Thanks for clearing this up for me Will! :D

tests/data/mongod.conf
MiaAltieri: Great question, this IP actually gets overwritten in [line 87](https://github.com/MiaAltieri/mongodb-operator/blob/71559c44d1a0611f7d33f9ed080365ee2c64e3f2/tests/integration/test_charm.py#L87) this is because this IP is meant to be the unit's IP address and we wouldn't know it ahead of time. I see how this is confusing. Maybe I could change it to:
```
net:
  bindIp: localhost,<placeholder_unit_ip>
```
Would that make things more clear?

tests/integration/test_charm.py
MiaAltieri: Hi @marceloneppel , great question! Yes I have several `for` loops that loop through the units in this file. My plan was to generalize the code and then implement it in a future PR. 

@jnsgruk correct me if I am wrong here, but my thought was to test multiple unit functionality after `juju add-unit` has been implemented.

tests/integration/test_charm.py
MiaAltieri: Yes! Actually I think that `client.server_info()` returns a string when the server is ready and otherwise it throws a `ServerSelectionTimeoutError` do you think I should just check that it doesn't throw a `ServerSelectionTimeoutError`?

tests/data/mongod.conf
WRFitch: Ah, I understand now. Setting it to explicitly be a placeholder would make this clearer, thanks!

tests/integration/helpers.py
```diff
@@ -0,0 +1,53 @@
...
+        await run_command_on_unit(unit, f"chmod {PERMISSION_MASK_FOR_SCP} {path}")
+
+    # Get the contents of the file.
+    temp_file = tempfile.NamedTemporaryFile()
+    await unit.scp_from(path, temp_file.name, scp_opts=["-v"])
+    data = temp_file.read().decode("utf-8")
+    temp_file.close()
+
+    # Change the file permissions back to the original mask.
+    if permissions != PERMISSION_MASK_FOR_SCP:
```
jnsgruk, line 35: This condition doesn't seem right? I think you need a flag somewhere to show that you modified the permissions just after line 26 above then check for that condition here?

tests/integration/helpers.py
```diff
@@ -0,0 +1,53 @@
+# Copyright 2021 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+
+import tempfile
+from typing import Optional
+
+PERMISSION_MASK_FOR_SCP = 644
```
jnsgruk, line 8:
```suggestion
_PERMISSION_MASK_FOR_SCP = 644
```

tests/integration/test_charm.py
jnsgruk: Yeh, I think it's good to keep this general, but it might be nicer to use pytest's `paramterize` feature and keep the actual test code relatively simple, while using `parameterize` to modify which unit the test runs on?

(Super simple example: https://github.com/canonical/interview-questions/blob/71dfb9d6d981be8b83d22c2f41b9ede93e758a7b/q01/tests/test_validation.py#L20)

tests/integration/test_charm.py
jnsgruk: This might help: https://docs.pytest.org/en/6.2.x/assert.html#assertions-about-expected-exceptions

jnsgruk (commented):
Looking good so far.

tests/integration/test_charm.py
marceloneppel: I was thinking about checking the content of the string (just to confirm that it returns the expected content based on the url that was passed to the client). But, an additional check like you and Jon mentioned would be very nice (for testing with an invalid url).

tests/integration/helpers.py
MiaAltieri: Sounds good to me, updating this

@marceloneppel , tagging you here since I used your code for this and we should keep it consistent. Maybe we should add this to a library instead since it seems quite useful and I could see other people needing :-)

tests/integration/helpers.py
MiaAltieri: @marceloneppel see above

tests/integration/test_charm.py
MiaAltieri: @marceloneppel I was thinking of doing something like this at first, but then I realised that the information returned by `server_info` gives information about the server that MongoDB is running on, such as the version of MongoDB, the version of git, what javascriptEngine is running, what the build environment is, etc. It doesn't really provide anything that is related to the deployment of MongoDB. 

**But** by returning this information and not throwing an error, it verifies that we were able to make a connection with MongoDB on both the units specified **and** the replicaset specified. This verifies two things:
1. an instance of MongoDB is running on the specified unit
2. the replicaset specified does exist **and** we can reach it

If we wanted to do further checks we could ask the client for information with something like `client.admin.command("replSetGetConfig")` in which we pass the command `replSetGetConfig` to the MongoDB client which we hopefully give us the config options for the replica set. And we could check that information to verify that specified replica set is there and that it runs for each unit. But I think this would be redundant since the fact we get a response from the server indicates these things already. 

I could go either way on this  🤷 , let me know if you want to do these extra checks. I trust your judgement 😄

tests/integration/helpers.py
marceloneppel: Thanks Mia! Yes, I agree that makes sense to create a library for that. We can discuss about it as probably there will be more charms testing this conditions.

tests/integration/test_charm.py
marceloneppel: Hi Mia! Yes. I agree that checking that information that you described is redundant. With all this info now, I believe we can keep the test as they are. Thanks for the details, Mia!

MiaAltieri: @jnsgruk Thanks for this link! It's quite helpful about asserting that an exception did occur. Since I am in the case of trying to assert that an exception did not occur. It makes things more complicated since we can't use the `with pytest.raises(ServerSelectionTimeoutError):` I think it might be the case that we would have to try and if an exception is thrown then we assert False with an explaination. Or am I missing something. Sorry if it is obvious I'm still experiencing some fatigue and brainfog

WRFitch (approved):

marceloneppel (approved):

jnsgruk (approved):

PR #9, opened 2022-02-04
Functionality to display primary replica (depends on PR #8)

MiaAltieri:
NOTE: This PR depends on #8 so it should not be merged before that PR is approved and Merged

Functionality to get the primary replica with juju action `get-primary` and primary status indicated as a message in `juju status`

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -1,4 +1,5 @@
 #!/usr/bin/env python3
+"""Charm code for MongoDB service."""
 # Copyright 2021 Canonical Ltd.
 # See LICENSE file for licensing details.
 import json
@@ -37,26 +38,18 @@ class MongodbOperatorCharm(ops.charm.CharmBase):
 
         self.framework.observe(self.on.install, self._on_install)
         self.framework.observe(self.on.config_changed, self._on_config_changed)
-        self.framework.observe(
-            self.on.mongodb_relation_joined, self._mongodb_relation_joined)
+        self.framework.observe(self.on.mongodb_relation_joined, self._mongodb_relation_joined)
         self.framework.observe(self.on.start, self._on_start)
         self.framework.observe(self.on.update_status, self._on_update_status)
-        self.framework.observe(self.on.get_primary_action,
-                               self._on_get_primary_action)
+        self.framework.observe(self.on.get_primary_action, self._on_get_primary_action)
 
     def _mongodb_relation_joined(self, event) -> None:
         if not self.unit.is_leader():
             return
 
         # do not initialise replica set until all peers have joined
-        if not self._number_of_expected_units == self._number_of_current_units:
-            logger.debug(
-                "planned units does not match joined units: %s != %s, will "
-                + "wait to initialise replica set until all planned units "
-                + "have joined",
-                str(self._number_of_expected_units),
-                str(self._number_of_current_units),
-            )
+        if not self._check_unit_count:
+            logger.debug("waiting to initialise replica set until all planned units have joined")
             return
 
         # do not initilise replica set until all peers have started mongod
@@ -64,10 +57,9 @@ class MongodbOperatorCharm(ops.charm.CharmBase):
             mongo_peer = self._single_mongo_replica(
                 str(self._peers.data[peer].get("private-address"))
             )
-            if not mongo_peer.is_ready(as_stand_alone=True):
+            if not mongo_peer.is_ready(standalone=True):
                 logger.debug(
-                    "unit: %s is not ready, cannot initilise replica set "
-                    + "until all units are ready, deferring on relation-joined",
+                    "unit: %s is not ready, cannot initilise replica set until all units are ready, deferring on relation-joined",
                     peer,
                 )
                 event.defer()
@@ -78,16 +70,13 @@ class MongodbOperatorCharm(ops.charm.CharmBase):
             try:
                 self._initialise_replica_set()
             except (ConnectionFailure, ConfigurationError):
-                self.unit.status = BlockedStatus(
-                    "failed to initialise replica set")
+                self.unit.status = BlockedStatus("failed to initialise replica set")
                 return
 
         # verify replica set is ready
         if not self._mongo.is_ready():
-            logger.debug(
-                "Replica set for units: %s, is not ready", self._unit_ips)
-            self.unit.status = BlockedStatus(
-                "failed to initialise replica set")
+            logger.debug("Replica set for units: %s, is not ready", self._unit_ips)
+            self.unit.status = BlockedStatus("failed to initialise replica set")
 
     def _on_install(self, _) -> None:
         """Handle the install event (fired on startup).
@@ -136,12 +125,11 @@ class MongodbOperatorCharm(ops.charm.CharmBase):
             logger.debug("opening tcp port")
             self._open_port_tcp(self._port)
         except subprocess.CalledProcessError:
-            self.unit.status = BlockedStatus(
-                "failed to open TCP port for MongoDB")
+            self.unit.status = BlockedStatus("failed to open TCP port for MongoDB")
             return
 
         # check if this unit's deployment of MongoDB is ready
-        if not self._mongo.is_ready(as_stand_alone=True):
+        if not self._mongo.is_ready(standalone=True):
             logger.debug("mongoDB not ready, deferring on start event")
             self.unit.status = WaitingStatus("waiting for MongoDB to start")
             event.defer()
@@ -155,21 +143,15 @@ class MongodbOperatorCharm(ops.charm.CharmBase):
             return
 
         # do not initialise replica set until all peers have joined
-        if not self._number_of_expected_units == self._number_of_current_units:
-            logger.debug(
-                "planned units does not match joined units: %s != %s,"
-                + "waiting to initialise replica set",
-                str(self._number_of_expected_units),
-                str(self._number_of_current_units),
-            )
+        if not self._check_unit_count:
+            logger.debug("waiting to initialise replica set until all planned units have joined")
             return
 
         # initialise replica set
         try:
             self._initialise_replica_set()
         except (ConnectionFailure, ConfigurationError):
-            self.unit.status = BlockedStatus(
-                "failed to initialise replica set")
+            self.unit.status = BlockedStatus("failed to initialise replica set")
 
     def _on_update_status(self, _):
         # connect to client for this single unit
@@ -184,7 +166,8 @@ class MongodbOperatorCharm(ops.charm.CharmBase):
     def _on_get_primary_action(self, event: ops.charm.ActionEvent):
         # check if current unit is the primary unit
         mongod_unit = self._single_mongo_replica(
-            str(self.model.get_binding(PEER).network.bind_address))
+            str(self.model.get_binding(PEER).network.bind_address)
+        )
 
         # if unit is primary display this inforamtion and exit
         if mongod_unit._is_primary:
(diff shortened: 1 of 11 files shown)


tests/integration/test_charm.py
```diff
@@ -75,8 +81,130 @@ async def test_database_is_up_as_replica_set(ops_test: OpsTest, unit_id: int) ->
...
+    # check if get-primary returns the correct primary unit regardless of
+    # which unit the action is run on
+    for unit in ops_test.model.applications[APP_NAME].units:
+        # use get-primary action to find primary
+        action = await unit.run_action("get-primary")
+        action = await action.wait()
+        identified_primary = action.results["replica-set-primary"]
+
+        # assert get-primary returned the right primary
+        assert identified_primary == expected_primary
```
marceloneppel, line 158: I believe this can trigger an error if I correctly understood the action code in the comment about it returning the primary just when called in the primary unit.

src/charm.py
```diff
@@ -79,29 +126,73 @@ def _on_start(self, event: ops.charm.StartEvent) -> None:
...
+    def _on_update_status(self, event):
+        # connect to client for this single unit
+        mongod_unit = self._single_mongo_replica(
+            str(self.model.get_binding(PEER).network.bind_address))
+
+        # if unit is primary then update status
+        if mongod_unit._is_primary:
+            self.unit.status = ActiveStatus("Replica set primary")
+
+    def _on_get_primary_action(self, event):
```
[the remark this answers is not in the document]
MiaAltieri: good eye!! thank you!

WRFitch (approved):
Some minor notes, but otherwise LGTM

src/charm.py
```diff
@@ -79,29 +126,73 @@ def _on_start(self, event: ops.charm.StartEvent) -> None:
...
+        # check if current unit is the primary unit
+        mongod_unit = self._single_mongo_replica(
+            str(self.model.get_binding(PEER).network.bind_address))
+
+        # if unit is primary display this inforamtion and exit
+        if mongod_unit._is_primary:
+            event.set_results({"replica-set-primary": self.unit.name})
             return
```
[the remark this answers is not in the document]
MiaAltieri: Yes, sorry if this code is confusing, the `README` is correct that "the action can be called in any unit to retrieve the primary". This return on line 184 just prevents the rest of the code from running. ie `# if unit is primary display this inforamtion and exit` and  `# loop through peers and check if one is the primary`

If the current unit isn't the primary, the lines 186-195 loop to check if a peer is a primary, and it if a peer is a primary that is outputted on line 194. Does that clear things up? 

BUT actually, I realize this code can be simplified using the `self.unit_ips` so I will simplify it as so:
```
for unit_ip in self._unit_ips():
            # set up a mongo client for this single replica
            mongod_unit = self._single_mongo_replica(unit_ip)

            # if unit is primary display this inforamtion and exit
            if mongod_unit._is_primary:
                event.set_results({"replica-set-primary": unit.name})
                return
```
thanks for the thorough reviews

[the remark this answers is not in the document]
marceloneppel: Sorry Mia, my bad. I incorrectly read the code and thougth  that the loop was other function. The code is clear. Sorry for the confusion.

tests/integration/test_charm.py
marceloneppel: As the comment above, I confused the flow of the code.

marceloneppel (approved):

src/charm.py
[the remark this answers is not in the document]
MiaAltieri: No its really good that you commented because you made me realized I could simplify my code!! I'm super grateful @marceloneppel

tests/integration/test_charm.py
```diff
@@ -75,8 +81,130 @@ async def test_database_is_up_as_replica_set(ops_test: OpsTest, unit_id: int) ->
...
+    """
+    # determine which unit is the primary
+    expected_primary = None
+    for unit in ops_test.model.applications[APP_NAME].units:
+        # connect to mongod
+        client = MongoClient(unit_uri(unit.public_address))
+
+        # check primary status
+        if client.is_primary:
+            expected_primary = unit.name
```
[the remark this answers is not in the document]
MiaAltieri: agreed

tests/integration/test_charm.py
```diff
@@ -75,8 +81,130 @@ async def test_database_is_up_as_replica_set(ops_test: OpsTest, unit_id: int) ->
...
+
+    # grab remaining IPS, must compare against leader_ip since it is not
+    # gaurenteed that the leader unit will be fully shutdown
+    ip_addresses = []
+    for unit in ops_test.model.applications[APP_NAME].units:
+        if not unit.public_address == leader_ip:
+            ip_addresses.append(unit.public_address)
+
+    # check that the replica set with the remaining units has a primary
+    replica_set_uri = "mongodb://{}:27017,{}:27017/replicaSet=rs0".format(
```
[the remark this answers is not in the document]
MiaAltieri: agreed! Ironically enough I already have a PORT constanst, I just didn't use it

src/charm.py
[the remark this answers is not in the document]
MiaAltieri: 🤦 Actually I can't do my above solution, since I need access to `unit` to get its name and there is no nice way to get all the units from the application and their IPs.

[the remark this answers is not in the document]
MiaAltieri: Even with 
```
                
    def _on_get_primary_action(self, event: ops.charm.ActionEvent):
        # get all units
        units = self._peers.units
        units.add(self.unit)

        # check each unit for primary status
        for unit, unit_ip in zip(units, self._unit_ips):
            # set up a mongo client for this single replica
            mongod_unit = self._single_mongo_replica(unit_ip)

            # if unit is primary display this inforamtion and exit
            if mongod_unit._is_primary:
                event.set_results({"replica-set-primary": unit.name})
                return
```
we will run into issues with maintaining the order of the `set` `units`, we could convert the `set` to a `list` but that leads to overhead so I'll keep this solution that I already have

jnsgruk:
Hmm, this feels like the relevant changes need rebasing on top of main now that #8 is merged?

src/charm.py
```diff
@@ -79,16 +70,13 @@ def _mongodb_relation_joined(self, event) -> None:
...
                 return
 
         # verify replica set is ready
         if not self._mongo.is_ready():
-            logger.debug(
-                "Replica set for units: %s, is not ready", self._unit_ips)
-            self.unit.status = BlockedStatus(
-                "failed to initialise replica set")
+            logger.debug("Replica set for units: %s, is not ready", self._unit_ips)
+            self.unit.status = BlockedStatus("failed to initialise replica set")
```
jnsgruk, line 79: Should this be Blocked, or Waiting? Feels like the case that warrants a blocked status is handled above -- blocked should only be used in the "HALP I NEED A HUMAN" case :)

src/charm.py
```diff
@@ -38,37 +38,28 @@ def __init__(self, *args):
...
-            logger.debug(
-                "planned units does not match joined units: %s != %s, will "
-                + "wait to initialise replica set until all planned units "
-                + "have joined",
-                str(self._number_of_expected_units),
-                str(self._number_of_current_units),
-            )
+        if not self._check_unit_count:
+            logger.debug("waiting to initialise replica set until all planned units have joined")
             return
```
jnsgruk, line 53: I don't recall if there is some nuance here - but in the next section you defer the event before returning - should you be deferring here?

src/charm.py
```diff
@@ -137,12 +125,11 @@ def _on_start(self, event: ops.charm.StartEvent) -> None:
             logger.debug("opening tcp port")
             self._open_port_tcp(self._port)
         except subprocess.CalledProcessError:
-            self.unit.status = BlockedStatus(
-                "failed to open TCP port for MongoDB")
+            self.unit.status = BlockedStatus("failed to open TCP port for MongoDB")
```
jnsgruk, line 128:
```suggestion
            self.unit.status = BlockedStatus("failed to open TCP port for MongoDB")
```

Not sure if this warrants a `BlockedStatus` - might just be worth logging with `logger.error`?

src/mongoserver.py
```diff
@@ -53,18 +52,19 @@ def check_server_info(self, client: MongoClient):
         """
         return client.server_info()
 
-    def is_ready(self, as_stand_alone=False) -> bool:
+    def is_ready(self, standalone=False) -> bool:
```
jnsgruk, line 55:
```suggestion
    def is_ready(self, standalone: bool=False) -> bool:
```

other methods seem to have type annotations, so might be worth being consistent here. Also, on a positive side if you indicate the type with an annotation, you can simplify the docstring and remove the type information :)

jnsgruk (commented):
Looking good, a few minor comments :)

src/charm.py
MiaAltieri: Yes! You are correct, Balbir and I discussed this in a meeting along with other changes that might want to be made to replication. We were thinking that they would be done in a the next PRs. Is it okay if I fix this in those PRs?

src/charm.py
MiaAltieri: Yeah, this was something that _really_ confused me for awhile. If we defer here the **same** event will get called for the unit that created the `RelationJoinedEvent` meaning that if planned units will still not match the joined units. (this is because the other `RelationJoinedEvents` for the other units will not be called, and thus this condition will never be true and will forever defer.)

MiaAltieri: Here is a longer explaination:

That's what I initially thought too! But actually having a defer here leads to a forever wait. For example:

Say you have 3 units (`A`,`B`,`C`) (and `A` is the leader). When the leader observes the first `relation-joined` event this is because **one** of the units `B` joined, **but `C` has not joined** thus:

`self.app.planned_units() == len(self._peers.units) + 1` = `3 = 2` 

so the condition is not satisfied. If we defer we will **keep waiting**, and we will **not** observe `C` joining. In order to observe `C` we will need to wait for the **second** `relation-joined` event. Once the second `relation-joined` even fires we will observe `C` and: 

`self.app.planned_units() == len(self._peers.units) + 1` = `3 = 3`

And then will will be ready.

TL;DR: We cannot defer since we need to allow all `relation-joined` events to fire so we can observe all the peers. 

I hope that makes sense 😅 its kind of confusing, I'm also super happy to talk about this on MM. This was something that tripped me up for _awhile_

src/charm.py
jnsgruk: Sounds good!

src/charm.py
jnsgruk: Nice, sounds good to me!

src/charm.py
```diff
@@ -156,21 +143,15 @@ def _on_start(self, event: ops.charm.StartEvent) -> None:
...
+            logger.debug("waiting to initialise replica set until all planned units have joined")
             return
 
         # initialise replica set
         try:
             self._initialise_replica_set()
         except (ConnectionFailure, ConfigurationError):
-            self.unit.status = BlockedStatus(
-                "failed to initialise replica set")
+            self.unit.status = BlockedStatus("failed to initialise replica set")
```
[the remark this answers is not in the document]
MiaAltieri: Hi Jon! I initially had this as part of the `_initialise_replica_set` function. But decided to pull it out so that I could return on the condition that it throws an exception on [line 74](https://github.com/canonical/mongodb-operator/blob/21dc4c51036006f84110f8bc97f943c5b8a99b43/src/charm.py#L74) before checking if the replica set is ready [lines 77-79](https://github.com/canonical/mongodb-operator/blob/21dc4c51036006f84110f8bc97f943c5b8a99b43/src/charm.py#L76-L79). I am happy to address this now, or in my subsequent PR(s) where I refactor the replication code. Let me know what you prefer. :)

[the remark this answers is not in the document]
jnsgruk: A subsequent PR as part of your refactoring is fine!

jnsgruk (approved):

PR #12, opened 2022-02-07
Change mongoserver (depends on PR #9)

MiaAltieri:
NOTE: This PR depends on https://github.com/canonical/mongodb-operator/pull/9 so it should not be merged before that PR is **Approved and Merged**

Changing `src/mongoserver.py` to `src/mongod_helpers.py`

See [JIRA Story](https://warthogs.atlassian.net/browse/DPE-45?atlOrigin=eyJpIjoiODJmZmRhNjQ2YmNlNDUyNGIwNzEzZThhNDRjNmNkZDAiLCJwIjoiaiJ9) for more details


src/charm.py
```diff
@@ -31,12 +31,62 @@ def __init__(self, *args):
...
+    def _mongodb_relation_joined(self, event) -> None:
+        if not self.unit.is_leader():
+            return
+
+        # do not initialise replica set until all peers have joined
+        if not self._number_of_expected_units == self._number_of_current_units:
+            logger.debug(
+                "planned units does not match joined units: %s != %s, will "
+                + "wait to initialise replica set until all planned units "
+                + "have joined",
```
balbirthomas, line 50: In this relation joined event a new Mongodb peer unit has joined so why not just add it to the replica set, without bothering to check if all planned units have joined ? In the future when additional peer units join the relation the charm will again get relation joined events for them too.

src/charm.py
```diff
@@ -31,12 +31,62 @@ def __init__(self, *args):
...
+        for peer in self._peers.units:
+            mongo_peer = self._single_mongo_replica(
+                str(self._peers.data[peer].get("private-address")))
+            if not mongo_peer.is_ready(as_stand_alone=True):
+                logger.debug(
+                    "unit: %s is not ready, cannot initilise replica set "
+                    + "until all units are ready, deferring on relation-joined",
+                    peer,
+                )
+                event.defer()
```
balbirthomas, line 66: I would be hesitant to defer a relation joined event. Since when new units join you are bound to get more relation joined events. Uptil now the only use case I have come across where a defer becomes essential is the start event.

src/charm.py
```diff
@@ -79,29 +126,73 @@ def _on_start(self, event: ops.charm.StartEvent) -> None:
...
-        logger.debug("initialising replica set")
+        # mongod is now active
+        self.unit.status = ActiveStatus()
+
+        # only leader should initialise the replica set
+        if not self.unit.is_leader():
+            return
+
+        # do not initialise replica set until all peers have joined
+        if not self._number_of_expected_units == self._number_of_current_units:
```
balbirthomas, line 151: I may be missing something but am surprised to see two places in the charm where there is a need to check expected/current units and "initialise" the replica set. Since AFAIU the MonogDB replica set needs to be initialized only once and so it could be done just once for example in the start event handler. There after there certainly will be the need to "reconfigure" the replica set every time a unit joins or departs. 

Also it may be simpler to compare the number of units in the replica set and the number of units currently available rather than the number of units currently available and the number of units expected.

PR #13, opened 2022-02-08
CI workflow

MiaAltieri:
CI workflow added to GitHub, modifications (and corrections) made to unit tests to facilitate the CI testing. Also some major reformatting was done for the linter

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,36 @@
+name: Tests
+on: [pull_request, push]
+
+jobs:
+  lint:
+    name: Lint
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Install dependencies
+        run: python3 -m pip install tox
+      - name: Run linters
+        run: tox -e lint
+  unit-test:
+    name: Unit tests
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Install dependencies
+        run: python -m pip install tox
+      - name: Run tests
+        run: tox -e unit
+  integration-test-microk8s:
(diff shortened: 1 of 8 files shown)


quiet-ranger (approved):
Most of the changes are harmless.  I only skimmed over the unit tests because we have a separate story to refactor them where needed.

.github/workflows/main.yml
```diff
@@ -0,0 +1,36 @@
+name: Tests
+on: [pull_request, push]
```
zmraul, line 2: I'm unaware If this has been discussed, but should the CI run on all branches, or specify only on main branch?

src/charm.py
```diff
@@ -1,21 +1,28 @@
...
-from mongoserver import MONGODB_PORT, MongoDB, ConnectionFailure, ConfigurationError
+from ops.model import (
+    ActiveStatus,
+    BlockedStatus,
+    MaintenanceStatus,
+    Relation,
+    WaitingStatus,
+)
+
+from mongoserver import MONGODB_PORT, ConfigurationError, ConnectionFailure, MongoDB
```
zeljkoagic, line 25: Not sure if our convention is to break over 80 lines; If yes, then we can multiline this import too.

zmraul (commented):
A small comment on the yml file. Otherwise I think it's fantastic to have CI up and running for the project!

.github/workflows/main.yml
jnsgruk: Good catch, Mia see https://github.com/canonical/operator-template/blob/d54942b30e9f1f0b2eb402d876f41bfd3a9efa62/.github/workflows/ci.yaml#L5

src/charm.py
jnsgruk: 99 lines for charms, per the styleguide :)

jnsgruk (approved):

PR #27, opened 2022-03-23
Add a check for if a replica is ready

MiaAltieri:
### Problem

As of now a the MongoDB charm checks `is_ready` by seeing if the mongo daemon is running (`mongod`) while this is important to do, it is also important that we check if replica is in the ready state (ie is the replica is ready to serve read and write requests.)

So essentially the problem is that we have no check to see if the replica is ready to serve read or write requests

### Solution

To resolve this we add a second function `is_replica_ready` which returns true if the replica is ready to serve read and write requests. This is done by checking if that replica is in the `SECONDARY`, `PRIMARY`, or `ARBITER` state.

Note: when in the `ARBITER` state we cannot accept reads or writes, but this replica will still be considered as ready as it is not in an error state or starting up or rolling back. 

### Context

There are two definitions of ready:
- `mongod` (the MongoDB daemon) is available to serve requests 
- The replicas in the replica set are ready to function ie are in one of the states `SECONDARY`, `PRIMARY`, or `ARBITER`

### Release Notes

Major Changes:
- `src/mongod_helpers.py` - old `is_ready` function has now been changed to `is_mongod_started` to indicate that this function should be used to check whether `mongod` is available
-  `src/mongod_helpers.py` - `is_replica_ready` now checks whether the replica is in one of the ready states (`SECONDARY`, `PRIMARY`, or `ARBITER`)
-  `src/mongod_helpers.py` - `check_replica_status` is added as a helper function to retrieve the replica status

Minor Changes:
- `src/charm.py` usage of `is_ready` updated to check either the replica status or the mongod status where appropriate
- `tests/unit/test_mongod_helpers` added new tests and updating old tests 
- `tests/unit/test_mongod_helpers` updating old tests for this new schema

### Notes to the reader

This is the first step in a larger process. I wanted to break up my PRs in a digestible way and this is my attempt to do so.

_What is the larger process you might ask?_ Well, the plan is to:
- Have unit status properly reflect the replica status as per the JIRA task [DPE-112](https://warthogs.atlassian.net/jira/software/c/projects/DPE/boards/390?modal=detail&selectedIssue=DPE-112&quickFilter=844)
- Using the replica status wait to add replicas the replica set until all replicas are ready as per the JIRA task [DPE-113](https://warthogs.atlassian.net/jira/software/c/projects/DPE/boards/390?modal=detail&selectedIssue=DPE-113&quickFilter=844)
- Using the replica status wait to remove replicas from the replica set until all replicas are ready as per the JIRA task [DPE-114](https://warthogs.atlassian.net/jira/software/c/projects/DPE/boards/390?modal=detail&selectedIssue=DPE-114&quickFilter=844)
- I see that this code gets re-used often, I will make a new PR for this to create a const for this:
```        
       exceptions = [
            ConnectionFailure("error message"),
            ConfigurationError("error message"),
            OperationFailure("error message"),
        ]
```

### DIFF
--- a/src/mongod_helpers.py
+++ b/src/mongod_helpers.py
@@ -53,18 +53,50 @@ class MongoDB:
         """
         return client.server_info()
 
-    def is_ready(self, standalone: bool = False) -> bool:
-        """Is the MongoDB server ready to services requests.
+    def check_replica_status(self) -> str:
+        """Retrieves the status of the current replica."""
+        # connect to mongod and retrieve replica set status
+        client = self.client(standalone=True)
+        try:
+            status = client.admin.command("replSetGetStatus")
+        except (ConnectionFailure, ConfigurationError, OperationFailure) as e:
+            logger.error("Failed to check the replica status, error: %s", e)
+            raise e
+        finally:
+            client.close()
 
-        Args:
-            standalone: an optional flag that indicates if the client should check if a single
-            instance of MongoDB or the entire replica set is ready
+        # look for our ip address in set of members
+        for member in status["members"]:
+            # get member ip without ":PORT"
+            if self._calling_unit_ip == member["name"].split(":")[0]:
+                return member["stateStr"]
+
+        return None
+
+    def is_replica_ready(self) -> bool:
+        """Is the MongoDB replica in the ready state.
+
+        Returns:
+            bool: True if services is ready False otherwise.
+        """
+        try:
+            replica_status = self.check_replica_status()
+        except (ConnectionFailure, ConfigurationError, OperationFailure):
+            return None
+
+        if replica_status == "PRIMARY" or replica_status == "SECONDARY" or replica_status == "ARBITER":
+            return True
+
+        return False
+
+    def is_mongod_ready(self) -> bool:
+        """Are mongod services available on a singe replica.
 
         Returns:
             bool: True if services is ready False otherwise.
         """
         ready = False
-        client = self.client(standalone)
+        client = self.client(standalone=True)
 
         try:
             self.check_server_info(client)
@@ -84,7 +116,7 @@ class MongoDB:
         is_replica_set = False
 
         # cannot be in replica set status if this server is not up
-        if not self.is_ready(standalone=True):
+        if not self.is_mongod_ready():
             return is_replica_set
 
         # access instance replica set configuration
--- a/src/mongod_helpers.py
+++ b/src/mongod_helpers.py
@@ -82,7 +82,7 @@ class MongoDB:
         try:
             replica_status = self.check_replica_status()
         except (ConnectionFailure, ConfigurationError, OperationFailure):
-            return None
+            return False
 
         if replica_status == "PRIMARY" or replica_status == "SECONDARY" or replica_status == "ARBITER":
             return True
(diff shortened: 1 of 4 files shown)


tests/unit/test_charm.py
```diff
@@ -186,7 +197,65 @@ def test_on_start_not_ready_defer(
...
+        is_mongod_ready,
+        is_replica_ready,
+        is_replica_set,
+    ):
+        """Tests that if a replica is not ready that the charm goes into waiting."""
+        self.harness.set_leader(True)
+        is_replica_set.return_value = True
+        service_running.return_value = True
+        is_mongod_ready.return_value = True
+        is_replica_ready.return_value = False
```
WRFitch, line 220: When you're patching a method, `return_value` can be set in the `@patch` statements, making these lines unnecessary for simple variables like booleans.

tests/unit/test_charm.py
```diff
@@ -510,17 +579,25 @@ def test_initialise_replica_failure_leads_to_waiting_state(
...
+        _open_port_tcp,
+        initialise_replica_set,
+        is_mongod_ready,
+        is_replica_ready,
     ):
         self.harness.set_leader(True)
         service_running.return_value = True
-        is_ready.return_value = True
+        is_mongod_ready.return_value = True
+        is_replica_ready.return_value = True
```
WRFitch, line 600: When you're patching a method, `return_value` can be set in the `@patch` statements, making these lines unnecessary for simple variables like booleans.

tests/unit/test_charm.py
```diff
@@ -186,7 +197,65 @@ def test_on_start_not_ready_defer(
...
+        _open_port_tcp,
+        is_mongod_ready,
+        check_replica_status,
+        is_replica_set,
+    ):
+        """Tests that failure to check replica state results in waiting status."""
+        self.harness.set_leader(True)
+        is_replica_set.return_value = True
+        service_running.return_value = True
+        is_mongod_ready.return_value = True
```
WRFitch, line 245: When you're patching a method, `return_value` can be set in the `@patch` statements, making these lines unnecessary for simple variables like booleans.

WRFitch (changes requested):
Looks good, I have a note about your method patching in the tests though.

tests/unit/test_charm.py
MiaAltieri: Great idea, since this is something that is seen in other areas of the unit tests I'll save this for my future PR where I clean up my unit tests.

src/mongod_helpers.py
```diff
@@ -53,18 +53,54 @@ def check_server_info(self, client: MongoClient):
...
-        """Is the MongoDB server ready to services requests.
+    def check_replica_status(self) -> str:
+        """Retrieves the status of the current replica."""
+        # connect to mongod and retrieve replica set status
+        client = self.client(standalone=True)
+        try:
+            status = client.admin.command("replSetGetStatus")
+        except (ConnectionFailure, ConfigurationError, OperationFailure) as e:
+            logger.error("Failed to check the replica status, error: %s", e)
+            raise e
```
shayancanonical, line 64: are we confident that these are the only exceptions that can occur when executing this command? also, should we be `raise`ing instead of `raise e` to preserve the original stack trace?

PR #29, opened 2022-03-24
DPE-113 wait for ready

MiaAltieri:
### Problem

[JIRA ticket](https://warthogs.atlassian.net/jira/software/c/projects/DPE/boards/390?modal=detail&selectedIssue=DPE-113&quickFilter=844)

As of now a the MongoDB charm adds a new replica to the replica set if that replica has `mongod` running, however an extra check is needed. A check to see if all of the current replica set hosts are ready. Without this check there can be an increased downtime or performance degradation (see more on why in the section Context)

### Solution

To resolve this we add a function `all_replicas_ready` which returns true if all replicas are ready to serve read and write requests. And this is checked before adding a replica to the replica set. 

### Context

It is important that all replicas be in the ready state before adding a new replica because during scale-up, each new member takes a backup from an alive member. e.g. during scale up from 3 to 6 all 3 current members will be loaded by the backup procedure. as result, they will not be able to serve (business) application queries. 

### Release Notes

Major Changes:
- `src/mongod_helpers.py` - `all_replicas_ready` function returns true if for every replica host each replica is ready to serve read/write requests

Minor Changes:
- `src/mongod_helpers.py` - now has field `_replica_set_hosts` this is different from the unit its, in that the unit ips are hosting MongoDB but not guaranteed to be hosting a replica

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -119,10 +119,18 @@ class MongodbOperatorCharm(ops.charm.CharmBase):
         calling_unit_ip = str(self._peers.data[calling_unit].get("private-address"))
         mongo_peer = self._single_mongo_replica(calling_unit_ip)
         if not mongo_peer.is_mongod_ready():
+            self.unit.status = WaitingStatus("waiting to reconfigure replica set")
             logger.debug(
                 "unit is not ready, cannot initialise replica set unit: %s is ready, deferring on relation-joined",
                 calling_unit.name,
             )
+            event.defer()
+            return
+
+        # only reconfigure if all current replicas of MongoDB application are in ready state
+        if not self._mongo.all_replicas_ready():
+            self.unit.status = WaitingStatus("waiting to reconfigure replica set")
+            logger.debug("waiting for all replica set hosts to be ready")
             event.defer()
             return
 
@@ -144,11 +152,13 @@ class MongodbOperatorCharm(ops.charm.CharmBase):
                 # update the set of replica set hosts
                 self._peers.data[self.app]["replica_set_hosts"] = json.dumps(self._unit_ips)
                 logger.debug("Replica set successfully reconfigured")
-                self.unit.status = ActiveStatus()
             except (ConnectionFailure, ConfigurationError, OperationFailure) as e:
                 logger.error("deferring reconfigure of replica set: %s", str(e))
                 self.unit.status = WaitingStatus("waiting to reconfigure replica set")
                 event.defer()
+                return
+
+        self.unit.status = ActiveStatus()
 
     def _on_install(self, _) -> None:
         """Handle the install event (fired on startup).
@@ -410,6 +420,7 @@ class MongodbOperatorCharm(ops.charm.CharmBase):
             "root_password": "password",
             "security_key": "",
             "unit_ips": self._unit_ips,
+            "replica_set_hosts": self._replica_set_hosts,
             "calling_unit_ip": str(self.model.get_binding(PEER).network.bind_address),
         }
         return config
--- a/src/charm.py
+++ b/src/charm.py
@@ -146,19 +146,20 @@ class MongodbOperatorCharm(ops.charm.CharmBase):
             event: The triggering relation event.
         """
         # check if reconfiguration is necessary
-        if self._need_replica_set_reconfiguration:
-            try:
-                self._mongo.reconfigure_replica_set()
-                # update the set of replica set hosts
-                self._peers.data[self.app]["replica_set_hosts"] = json.dumps(self._unit_ips)
-                logger.debug("Replica set successfully reconfigured")
-            except (ConnectionFailure, ConfigurationError, OperationFailure) as e:
-                logger.error("deferring reconfigure of replica set: %s", str(e))
-                self.unit.status = WaitingStatus("waiting to reconfigure replica set")
-                event.defer()
-                return
+        if not self._need_replica_set_reconfiguration:
+            self.unit.status = ActiveStatus()
+            return
 
-        self.unit.status = ActiveStatus()
+        try:
+            self._mongo.reconfigure_replica_set()
+            # update the set of replica set hosts
+            self._peers.data[self.app]["replica_set_hosts"] = json.dumps(self._unit_ips)
+            logger.debug("Replica set successfully reconfigured")
+            self.unit.status = ActiveStatus()
+        except (ConnectionFailure, ConfigurationError, OperationFailure) as e:
+            logger.error("deferring reconfigure of replica set: %s", str(e))
+            self.unit.status = WaitingStatus("waiting to reconfigure replica set")
+            event.defer()
 
     def _on_install(self, _) -> None:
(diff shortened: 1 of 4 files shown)


src/mongod_helpers.py
```diff
@@ -68,22 +74,41 @@ def check_replica_status(self) -> str:
...
                 return member["stateStr"]
 
         return None
 
-    def is_replica_ready(self) -> bool:
+    def all_replicas_ready(self) -> bool:
+        """Returns true if all replica hosts are ready."""
+        for unit_ip in self._replica_set_hosts:
+            if not self.is_replica_ready(unit_ip):
+                logger.debug("unit: %s is not ready", unit_ip)
```
WRFitch, line 87: Would it be useful to log all of the unready units?

src/charm.py
```diff
@@ -144,11 +152,13 @@ def _reconfigure(self, event: ops.charm.RelationEvent) -> None:
...
                 self._peers.data[self.app]["replica_set_hosts"] = json.dumps(self._unit_ips)
                 logger.debug("Replica set successfully reconfigured")
-                self.unit.status = ActiveStatus()
             except (ConnectionFailure, ConfigurationError, OperationFailure) as e:
                 logger.error("deferring reconfigure of replica set: %s", str(e))
                 self.unit.status = WaitingStatus("waiting to reconfigure replica set")
                 event.defer()
+                return
+
+        self.unit.status = ActiveStatus()
```
shayancanonical, line 161: nit: i usually try to avoid nested code structure if at all possible. in the above, would the following be better?

```
if not self._need_replica_set_reconfiguration:
    self.unit.status = ActiveStatus()
    return

try:
...
except:
...
```

tests/unit/test_charm.py
```diff
@@ -470,6 +468,68 @@ def test_mongodb_relation_joined_peer_ready(
...
+
+    @patch_network_get(private_address="1.1.1.1")
+    @patch("charm.MongodbOperatorCharm._single_mongo_replica")
+    @patch("charm.MongodbOperatorCharm._need_replica_set_reconfiguration", return_value=True)
+    @patch("mongod_helpers.MongoDB.reconfigure_replica_set")
+    @patch("mongod_helpers.MongoDB.all_replicas_ready", return_value=False)
+    def test_mongodb_relation_joined_all_replicas_not_ready(
+        self, all_replicas_ready, mongodb_reconfigure, need_reconfiguration, single_replica
+    ):
+        """Tests that when current replcia set hosts are not ready that we go into waiting.
```
WRFitch, line 512: This docstring is slightly clunky to read - consider using this: 
`"""Tests that we go into waiting when current ReplicaSet hosts are not ready"""`

WRFitch (commented):
Looks good, my complaints get more minor every time ;)

tests/unit/test_mongod_helpers.py
```diff
@@ -322,5 +323,32 @@ def test_check_replica_status_error(self, mock_client, client):
...
+    @patch("pymongo.MongoClient")
+    @patch("mongod_helpers.MongoDB.client")
+    def test_not_all_replicas_ready(self, is_replica_ready, mock_client, client):
+        """Verifies that all_replicas_ready returns False when not all replicas are ready."""
+        # standard presets
+        config = MONGO_CONFIG.copy()
+        mongo = MongoDB(config)
+
+        # verify result of all replicas ready
+        self.assertEqual(mongo.all_replicas_ready(), False)
```
shayancanonical, line 339: should we have a test for all replicas being ready?

marceloneppel (commented):
LGTM Mia! I commented about two possible small improvements.

src/charm.py
MiaAltieri: Yes! Much better, fully agree thanks for this suggestion Shayan

tests/unit/test_mongod_helpers.py
MiaAltieri: Hi Shayan :D 

Great point, this case actually gets covered in our integration tests and since we are adopting a new way of testing where unit tests are meant to test situations which are not covered in the integration tests.

src/mongod_helpers.py
```diff
@@ -68,22 +74,41 @@ def check_replica_status(self) -> str:
         # look for our ip address in set of members
         for member in status["members"]:
             # get member ip without ":PORT"
-            if self._calling_unit_ip == member["name"].split(":")[0]:
+            logger.debug(member["name"].split(":")[0])
+            if replica_ip == member["name"].split(":")[0]:
```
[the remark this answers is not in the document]
MiaAltieri: Agreed! Thanks for finding this :)

tests/unit/test_charm.py
```diff
@@ -470,6 +468,68 @@ def test_mongodb_relation_joined_peer_ready(
...
+        self.harness.set_leader(True)
+        single_replica.return_value.is_ready.return_value = True
+
+        for exception in PYMONGO_EXCEPTIONS:
+            # simulate failure to check replica status
+            check_replica_status.side_effect = exception
+
+            # simulate 2nd MongoDB unit
+            rel = self.harness.charm.model.get_relation("mongodb")
+            key_values = {"private-address": "127.4.5.6"}
```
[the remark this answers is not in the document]
MiaAltieri: 🤦 oops 

Yes, definitely!

src/mongod_helpers.py
MiaAltieri: Great catch Will! :D 

🤔 It would, but it might end up using more time / resources since we wouldn't return right away. (still `O(n)` with or without the return) so its wouldn't be a significant use of resources.

For clean-ness of code I think I will still opt for this solution unless you feel strongly :)

tests/unit/test_charm.py
MiaAltieri: Awesome, thanks for this recommendation! :D

WRFitch (approved):

src/charm.py
```diff
@@ -138,17 +146,21 @@ def _reconfigure(self, event: ops.charm.RelationEvent) -> None:
...
+            self._mongo.reconfigure_replica_set()
+            # update the set of replica set hosts
+            self._peers.data[self.app]["replica_set_hosts"] = json.dumps(self._unit_ips)
+            logger.debug("Replica set successfully reconfigured")
+            self.unit.status = ActiveStatus()
+        except (ConnectionFailure, ConfigurationError, OperationFailure) as e:
+            logger.error("deferring reconfigure of replica set: %s", str(e))
+            self.unit.status = WaitingStatus("waiting to reconfigure replica set")
+            event.defer()
+            return
```
shayancanonical, line 163: minor minor nit: may be these `event.defer()` and `return` are no longer necessary?

MiaAltieri: Great point! `return` is definitely no longer needed, but `event.defer()` is since we need to re-trigger the event, updating soon :D

shayancanonical (approved):
LGTM, left a minor nit

PR #30, opened 2022-03-30
DPE-113 wait for ready follow up PR

MiaAltieri:
### Problem

[JIRA ticket](https://warthogs.atlassian.net/jira/software/c/projects/DPE/boards/390?modal=detail&selectedIssue=DPE-114&quickFilter=844)

This is a follow up to this branch's previous [PR](https://github.com/canonical/mongodb-operator/pull/29), in which the previous PR checked if all replica set hosts were ready before adding a new unit. This PR checks if all replica set hosts are ready before removing a unit.

As of now a the MongoDB charm removes a replica from the replica set without checking much (other than if the replica is a primary and should be stepped down). However an extra check is needed. A check to see if all of the current replica set hosts are ready. Without this check there can be an increased downtime or performance degradation (see more on why in the section Context)

### Solution

To resolve this we use the boolean helper function `all_replicas_ready` to check if all replicas are ready to serve read and write requests. This check is used before removing a unit and if the conditions are not met, then we go into the waiting status. 

### Context

As is the same from the last PR:

It is important that all replicas be in the ready state before removing a new replica because during scale-down you inherentally decrease your avaialbilty of members. Decreasing members while some members are not yet ready can lead to downtime, or worse, you may loose data if a replica isn't fully backup. Similarily, if a new replica has been formed and is using the current unit to create its backup and we delete the current unit we will run into issues with downtime.

### Release Notes

Major Changes:
- `src/charm.py` - `_on_mongodb_remove_replica`  checks whether all replica hosts are ready before removing a replica. 

Minor Changes:
- `tests/unit/test_charm.py` - tests scenarios with removing units when not all replicas are ready.

### DIFF
--- a/tests/unit/test_charm.py
+++ b/tests/unit/test_charm.py
@@ -860,8 +860,8 @@ class TestCharm(unittest.TestCase):
     ):
         """Tests that we go into waiting when current ReplicaSet hosts are not ready.
 
-        Tests the scenario that if current the current hosts of MongoDB are not all ready that
-        the application waits before reconfiguring if the current unit is getting removed.
+        Tests the scenario that if the current hosts of MongoDB are not all ready that the
+        application waits before reconfiguring if the current unit is getting removed.
         """
         # preset values
         self.harness.set_leader(True)
--- a/tests/unit/test_charm.py
+++ b/tests/unit/test_charm.py
@@ -508,8 +508,8 @@ class TestCharm(unittest.TestCase):
     ):
         """Tests that we go into waiting when current ReplicaSet hosts are not ready.
 
-        Tests the scenario that a failure to check the current statuses of replica set hosts
-        results in WaitingStatus and no attempt to reconfigure is made.
+        Tests the scenario that if current replica set hosts are not ready, the leader goes into
+        WaitingStatus and no attempt to reconfigure is made.
         """
         # preset values
         self.harness.set_leader(True)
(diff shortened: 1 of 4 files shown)


tests/unit/test_charm.py
```diff
@@ -819,3 +819,59 @@ def test_storage_detaching_replica_removal_failure(self, _, remove_replica):
...
+    @patch("charm.MongodbOperatorCharm._single_mongo_replica")
+    @patch("charm.MongodbOperatorCharm._need_replica_set_reconfiguration", return_value=True)
+    @patch("mongod_helpers.MongoDB.reconfigure_replica_set")
+    @patch("mongod_helpers.MongoDB.all_replicas_ready", return_value=False)
+    def test_storage_detaching_all_replicas_not_ready(
+        self, all_replicas_ready, mongodb_reconfigure, need_reconfiguration, single_replica
+    ):
+        """Tests that we go into waiting when current ReplicaSet hosts are not ready.
+
+        Tests the scenario that if current the current hosts of MongoDB are not all ready that
```
shayancanonical, line 863: `Tests the scenario that if the current hosts of MongoDB`

shayancanonical (approved):
lgtm, left a minor comment

paulomach (approved):
LGTM

tests/unit/test_charm.py
MiaAltieri: 🤦 thanks Shayan for catching this

PR #32, opened 2022-04-07
Simplified repeated code in Integration tests

MiaAltieri:
Short PR descriptions as nothing major is changing here. (Changes are motivated by making the integration code cleaner and to minimise the size of the next PR).

Simplified repeated code by adding two helper functions:

`fetch_replica_set_members` - fetches members in the replica set 
`generate_replica_set_uri` - generates the URI used to connect to the MongoDB client

Updated retrieval of primary to use query to `mongod` rather than using `client.primary` as MongoDB client queries should rely on `mongod` rather than the helpers in `pymongo`. (As specified by managers and engineers).


WRFitch (approved):
LGTM!

tests/integration/test_charm.py
```diff
@@ -273,31 +249,55 @@ async def test_add_unit(ops_test: OpsTest) -> None:
...
+    return replica_set_uri
+
+
+def fetch_replica_set_members(replica_ips: List[str]):
+    """Fetches the IPs listed as replica set members in the MongoDB replica set configuration.
+
+    Args:
+        replica_ips: list of ips hosting the replica set.
+    """
+    print("replica_ips", replica_ips)
```
shayancanonical, line 286: delete print?

tests/integration/test_charm.py
```diff
@@ -273,31 +249,55 @@ async def test_add_unit(ops_test: OpsTest) -> None:
...
     # connect to replica set uri
-    replica_set_uri = "mongodb://{}:{},{}:{}/replicaSet=rs0".format(
-        ip_addresses[0], PORT, ip_addresses[1], PORT
-    )
-    client = MongoClient(replica_set_uri)
+    client = MongoClient(generate_replica_set_uri(replica_ips))
 
     # get ips from MongoDB replica set configuration
     rs_config = client.admin.command("replSetGetConfig")
+    print("rs_config", rs_config)
```
shayancanonical, line 292: delete print?

tests/integration/test_charm.py
```diff
@@ -321,12 +321,40 @@ def unit_uri(ip_address: str) -> str:
...
+
+    primary = None
+    # loop through all members in the replica set
+    for member in status["members"]:
+        # check replica's current state
+        if member["stateStr"] == "PRIMARY":
+            # get member ip without ":PORT"
+            primary = member["name"].split(":")[0]
+
+    return str(primary)
```
shayancanonical, line 349: if primary is `None`, this will return `str(None)`. this may cause issues on line 370

marceloneppel (approved):
Nice Mia. LGTM!

tests/integration/test_charm.py
MiaAltieri: yup yup yup good eye 👀

marcoppenheimer, line 349: I would move these helper functions to a `helpers.py` file within `tests/integration` and import them.

That way we can keep all methods called within `test_charm.py` standard `test_XYZ`

marcoppenheimer, line 349: I would move these helper functions to a `fixtures.py` file within `tests/integration` and import them.

That way we can keep all methods called within `test_charm.py` standard `test_XYZ`

tests/integration/test_charm.py
```diff
@@ -321,12 +321,40 @@ def unit_uri(ip_address: str) -> str:
     return "mongodb://{}:{}/".format(ip_address, PORT)
 
 
+def fetch_primary(replica_set_hosts: List[str]) -> str:
+    """Returns IP address of current replica set primary."""
+    # connect to MongoDB client
+    if len(replica_set_hosts) == 1:
+        replica_set_uri = replica_set_hosts[0] + ":" + str(PORT)
+        client = MongoClient(replica_set_uri, replicaset="rs0")
```
marcoppenheimer, line 329: Could this logic be moved into `generate_replica_set_uri`?

marcoppenheimer (approved):
Solid 👍🏽

tests/integration/test_charm.py
MiaAltieri: I don't see why not 😉

PR #33, opened 2022-04-11
added function for grabbing the unit ip

MiaAltieri:
Short PR description as this is a minor change. 

This change essentially adds a helper function for getting a unit’s IP address, it is helpful because it converts the address regardless of it being the current unit or the peer. It will also be helpful in a following PR

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -351,13 +351,15 @@ class MongodbOperatorCharm(ops.charm.CharmBase):
 
     def _unit_ip(self, unit: ops.model.Unit) -> str:
         """Returns the ip address of a given unit."""
+        # check if host is current host
         if unit == self.unit:
             return str(self.model.get_binding(PEER).network.bind_address)
-
-        if unit in self._peers.data:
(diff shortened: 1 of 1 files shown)


src/charm.py
```diff
@@ -351,6 +349,16 @@ def _single_mongo_replica(self, ip_address: str) -> MongoDB:
...
+    def _unit_ip(self, unit: ops.model.Unit) -> str:
+        """Returns the ip address of a given unit."""
+        if unit == self.unit:
+            return str(self.model.get_binding(PEER).network.bind_address)
+
+        if unit in self._peers.data:
+            return str(self._peers.data[unit].get("private-address"))
+
+        return None
+
```
marcoppenheimer, line 361: I quite like this encapsulation.
I would suggest changing the exit-points to be an:
```
if:
elif:
else: error
```

just in case returning 'None' wrecks things or upsets linters.

marcoppenheimer (approved):

WRFitch (approved):

PR #34, opened 2022-04-25
DPE-137-lock-on-removal

MiaAltieri:
### For the Reader
Things in `src/mongod_helpers.py` are somewhat messy, but have no fear! These will be updated in the following quarter when we merge the mongod helpers between the k8s and machine charm. 

NOTE: This branch called `DPE-137-lock-on-removal` actually does not use a lock, due to the issues related with deferring on storage and the fact that the current lock mechanism relies on relation events which are not reliable on unit departure.

### Problem

This PR addresses some of the complexities of removing units. Specifically the problems of:
1. Do not allow more than one unit to be removed at the same time by MongoDB while allowing mutliple units to be scaled down by Juju
2. Do not defer in storage detached 

### Solution

This PR enables a scale down feature of MongoDB such that it is possible to remove more than one unit of MongoDB (**as long as the user is not removing a majority**).

**NOTE** MongoDB does not allow multiple replicas to be removed at a time, so by "possible to remove more than one unit of MongoDB" we mean one can issue the command `juju remove-unit mongodb mongodb/0 mongodb/1 etc` and these units will be removed one at a time, while giving the user the experience of removing multiple units.

1. The solution uses versioning in config to allow only one unit to be removed at a time from mongod ([here](https://github.com/canonical/mongodb-operator/blob/f4c0cc0536752e5a0cc2ac3c2d2adb9848c416a2/src/mongod_helpers.py#L324)). This works by the fact that `mongod` cannot apply a configs with the same versions with different configurations. ie. if two units try to update the config at the same time when being removed, only one will be successful and one other **must** try again. 
2. Since Deferring isn't allowed `tenacity` retry function is used. In the case that we fail to remove a replica (cannot step down primary or cannot apply configuration) we will retry up to a period of 60s. ([here](https://github.com/canonical/mongodb-operator/blob/f4c0cc0536752e5a0cc2ac3c2d2adb9848c416a2/src/mongod_helpers.py#L290-L295))
3. If for some reason a unit cannot be removed from the replica set, then the leader unit will remove it from the config [here](https://github.com/canonical/mongodb-operator/blob/f4c0cc0536752e5a0cc2ac3c2d2adb9848c416a2/src/charm.py#L95)

### Context

1. Removing a Majority in MongoDB
Removing a majority of replicas at once can result in the inability to elect a new primary thus leading to downtime and potentially some work on the administrators side. 

2. Removing more than one unit in MongoDB is not something that is feasible in MongoDB. "Starting in MongoDB 4.4, [replSetReconfig](https://www.mongodb.com/docs/manual/reference/command/replSetReconfig/#mongodb-dbcommand-dbcmd.replSetReconfig) by default allows adding or removing no more than 1 [voting](https://www.mongodb.com/docs/manual/reference/replica-configuration/#mongodb-rsconf-rsconf.members-n-.votes) member at a time."

4. Deferring on Storage Detached Events
This is an operation that is not allowed.

5. Downtime in MongoDB
Downtime occurs when replicas are either busy copying data or there is no primary. Therefore it is important to eliminate these situations

### Release Notes

Major Changes:
- `src/charm.py` - 
     - `_on_mongodb_storage_detaching` - this function attempts to allow the unit departing to take care of itself (ie step down if it is the primary and remove itself from the config)
     - `_relation_departed` - this function allows the leader (if it is not departing) to remove units from mongodb if they were unsucessful in storage_detaching
     - `process_unremoved_units` - this function cross checks the juju hosts and the mongodb hosts to see if any units should be removed from the replica set config and then attempts to remove them

- `src/mongod_helpers.py` - 
     - `_is_any_removing` - checks if any unit is currently getting removed
     - `_is_primary` - checks if given unit the the primary
     - `remove_replset_member`- removes the given member from config and steps down it down as primary if necessary. It also checks that
         1. no other unit is currently removed (reduces downtime)
         2. there is not a race condition in removing units (will not allow multiple units to be removed at the same time)
     - `_hostname_from_hostport` - converts hostname to not use port


src/mongod_helpers.py
```diff
@@ -81,6 +99,33 @@ def check_replica_status(self, replica_ip) -> str:
...
+        # grab the replica set status
+        client = self.client()
+        try:
+            status = client.admin.command("replSetGetStatus")
+        except (ConnectionFailure, ConfigurationError, OperationFailure) as e:
+            raise e
+        finally:
+            client.close()
+
+        hosts = [member["name"].split(":")[0] for member in status["members"]]
```
delgod, line 113:
```suggestion
        hosts = [self._hostname_from_hostport(member["name"]) for member in status["members"]]
```

src/charm.py
```diff
@@ -55,57 +57,65 @@ def __init__(self, *args):
...
+            # it is not possible to defer in storage detached.
+            self._mongo.remove_replset_member(self._unit_ip(self.unit))
+        except NotReadyError:
+            logger.info(
+                "Failed to remove %s from replica set, another member is syncing", self.unit.name
+            )
+            event.defer()
+        except PyMongoError as e:
+            logger.info("Deferring reconfigure: error=%r", e)
             event.defer()
```
marceloneppel, line 82: I'm in doubt if these `defer` calls should be here. I'm a little bit confused with this plus the above comment and the PR description, both stating that isn't allowed to defer on storage detached event.

marceloneppel (commented):
LGTM. I have only one question about the `defer` calls.

src/charm.py
```diff
@@ -55,57 +57,65 @@ def __init__(self, *args):
...
+        self.framework.observe(self.on.mongodb_relation_departed, self._relation_departed)
 
-    def _update_hosts(self, _) -> None:
-        """Update replica set hosts."""
-        # allow leader to reset which hosts are being used
-        if self.unit.is_leader():
-            self._peers.data[self.app]["replica_set_hosts"] = json.dumps(self._unit_ips)
+    def _update_hosts(self, event) -> None:
+        """Update replica set hosts and remove any unremoved replicas from the config."""
+        if not self.unit.is_leader() and "replset_initialised" not in self._peers.data[self.app]:
```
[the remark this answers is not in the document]
MiaAltieri: 🤦‍♀️  😆  great catch, cant believe I did this! Thanks for seeing this Mykola :)

src/charm.py
MiaAltieri: You're right, this is outdated, I did something wrong with git here. Updating this right away. thank you for catching Marcelo :)

delgod (approved):

PR #36, opened 2022-05-13
Updating mongod to start with commandline args instead of config file

MiaAltieri:
### For the Reader
This is a big PR but dont fear! Most of the big diff is attributed to:
1. removing dead code that is no longer needed
2. updating a charm lib

Other notes
1. This is a small step in a bigger process where the process is:
update mongod to start with commands`->` start mongod with keyFile & set up password via commandline args`->` adopt k8s mongodb lib

3. NOTE: This change from starting mongodb with config file to starting with command line args motivates the need to remove associated `config_changed` hook, unit tests, integration tests, and test files. This ended up being a lot of code, so I apologise for the big diff here. If needed I can do this in a separate PR. But I think it's a good idea to remove unused code from a repo.

4. Apologies for updating operator lib in the process and not in a separate PR.

### Problem

Previously `mongod` which starts the MongoDB daemon starts by taking in a config file (`etc/mongodb.conf`). While this works, it is not ideal for future functionality where we want to allow the user to pass their own config file. Merging these two files would potentially be messy. It is instead better to start `mongod` by passing commandline args

### Solution

The solution is to pass args to the `mongod` process via its `mongod.service` file. 

### Context

1. `systemd` starts services via their service file ie `mongod.service`, so the necessary args should be set in that file
2. mongod requires access to its appropriate data folder `/data/db`
3. Unlike other databases, MongoDB only accepts one config file (this is our motivation for giving users sole domain over the config file)

### Release Notes

The main changes in this PR are passing args to `mongod` via the `mongod.service` file. This is done by copying the file over and then updating the commandline args as necessary. We also give `mongodb` permissions to access its appropriate data folder `data/db`

Now that the charm no longer uses the config file, all tests, hooks, data related to the `config_changed` hook are removed.

### Testing

Unit tests were properly updated and verify functionality. Integration tests verify that these changes do not effect the charms functionality. 

Tests for `config_changed` are removed as no longer necessary.


tests/unit/test_charm.py
```diff
@@ -44,56 +43,40 @@ def setUp(self, *unused):
...
+                self.harness.charm.unit.status, MaintenanceStatus("installing MongoDB")
+            )
+            _install.assert_called_once()
+            _add.assert_called_with(["mongodb-org"])
+            shutil.copy.assert_called_with(
+                "src/data/mongod.service", "/etc/systemd/system/mongod.service"
+            )
+            daemon_reload.assert_called()
+            getpwnam.assert_called()
+            chown.assert_called()
```
marcoppenheimer, line 67: These tests essentially just test the code execution, which might not be necessary as one would expect the core functionality is there. It's also harder to maintain if we ever make changes to `on_install`.

It might be better to test the following instead:


- Does the `src/data/mongod.service` file get updated correctly? It didn't add any funky newlines?
- Did `machine_ip` get inserted correctly?
- Does my code handle an invalid `src/data/mongod.service`?

Testing function calls is probably not needed.

src/charm.py
```diff
@@ -186,22 +191,45 @@ def _on_install(self, _) -> None:
         except (apt.InvalidSourceError, ValueError, apt.GPGKeyError, URLError):
             self.unit.status = BlockedStatus("couldn't install MongoDB")
 
+        # Construct the mongod startup commandline args for systemd
+        machine_ip = self._unit_ip(self.unit)
+        mongod_start_args = " ".join(
```
paulomach, line 196: Considering that the idea is allow users to pass custom config file in the future, will you be parsing the file and adding to this list or writing the file directly to the fs? 
In the latter case, will you completely dismiss the `mongo_start_args`? (because, cmd line args will superseed config file config, in case on conflicts)

paulomach (commented):
Add a quick view concerning future-proofing it.

src/data/mongod.service
```diff
@@ -0,0 +1,33 @@
...
+Description=MongoDB Database Server
+Documentation=https://docs.mongodb.org/manual
+After=network-online.target
+Wants=network-online.target
+
+[Service]
+User=mongodb
+Group=mongodb
+EnvironmentFile=-/etc/default/mongod
+ExecStart=/usr/bin/mongod --config /etc/mongod.conf
```
WRFitch, line 11: You could replace this with the following, read this file the way Marc suggests above, and then format this line as a string: 
```ini
ExecStart=%s
```

This is overkill, but you could also use the [`ConfigParser`](https://docs.python.org/3/library/configparser.html) library to read this file into a dict-like object and modify it the same way you normally do with a dict.

WRFitch (commented):
Marc's review already describes my concerns, but I had a suggestion about how the config file can be modified

src/charm.py
```diff
@@ -186,22 +191,45 @@ def _on_install(self, _) -> None:
...
+        # mongod.service file passes commandline args to mongod service
+        shutil.copy("src/data/mongod.service", "/etc/systemd/system/mongod.service")
+
+        with open("/etc/systemd/system/mongod.service", "r") as service_template:
+            mongodb_service = service_template.readlines()
+
+        mongodb_service[MONGO_EXEC_LINE] = mongod_start_args
+
+        with open("/etc/systemd/system/mongod.service", "w") as service_file:
+            service_file.writelines(mongodb_service)
```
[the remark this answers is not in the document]
MiaAltieri: First Comment: yes agreed, I like this, this sounds much better

Second Comment: So I need to do more research on `systemd` and where/when its service files get created. When the install hook executes there is _no_ `mongod.service` file on the machine so one isn't getting overwritten. (But I guess it prevents one from getting created so "in a way" it is overwriting the default file.)

A more ideal scenario would be to grab the default `mongod.service` file and then overwrite that one line. Let me see how I can grab that default file.

src/charm.py
```diff
@@ -211,9 +239,9 @@ def _on_start(self, event: ops.charm.StartEvent) -> None:
         """
         # start mongo service
         self.unit.status = MaintenanceStatus("starting MongoDB")
-        if not service_running("mongod.service"):
+        if not systemd.service_running("mongod.service"):
             logger.debug("starting mongod.service")
-            mongod_enabled = service_resume("mongod.service")
+            mongod_enabled = systemd.service_start("mongod.service")
```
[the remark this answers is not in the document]
MiaAltieri: Actually, after reading the [systemd.py](https://github.com/canonical/operator-libs-linux/blob/6a0ea2cac40d150ecf450960c7073eaa64ee5b7d/lib/charms/operator_libs_linux/v1/systemd.py#L123-L127) file, it seems that `systemd.service_start` either returns true or raises an error, so I think a `try` `except` block is more suited here

tests/unit/test_charm.py
```diff
@@ -44,56 +43,40 @@ def setUp(self, *unused):
...
-        # TODO change expected output based on config options,(once config options are implemented)
-        open_mock.assert_called_with("/etc/mongod.conf", "w")
-        open_mock.return_value.write.assert_has_calls(
-            [mock.call(arg) for arg in self.mongodb_config_args]
-        )
+        with mock.patch("builtins.open"):
+            self.harness.charm.on.install.emit()
+            self.assertEqual(
+                self.harness.charm.unit.status, MaintenanceStatus("installing MongoDB")
+            )
```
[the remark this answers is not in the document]
MiaAltieri: I 1000% agree, this was a relic from the past

tests/unit/test_charm.py
MiaAltieri: Yeah you're right, I just quickly edited this test. I believe we are now using unit tests to test conditions which are not tested by the integration test and the functionality of install is tested by the integration test so I believe this can be removed all together. Thoughts?

src/charm.py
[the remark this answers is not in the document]
marceloneppel: The default file is written to `/lib/systemd/system/mongod.service`.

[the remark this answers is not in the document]
MiaAltieri: @marceloneppel Interesting, thanks for sharing! It seems that "Drop-in files in `/etc` take precedence" over those in `/lib`. 

So this solution should work to create a new file while preserving the default file in `/lib/systemd/system/mongod.service`. :D

src/charm.py
```diff
@@ -40,6 +42,9 @@
     "deb [ arch=amd64,arm64 ] https://repo.mongodb.org/apt/ubuntu focal/mongodb-org/5.0 multiverse"
 )
 GPG_URL = "https://www.mongodb.org/static/pgp/server-5.0.asc"
+MONGO_EXEC_LINE = 10
```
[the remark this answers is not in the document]
MiaAltieri: opting for marcs solution. Thanks for pointing this out :)

src/charm.py
MiaAltieri: I am not sure what is the best option here, perhaps @delgod knows?

src/data/mongod.service
MiaAltieri: I am actually going to opt to read from `/lib/systemd/system/mongod.service` on the machine, modify it, then save it the modified version in `/etc/systemd/system/mongod.service`. So this file will no longer exist

src/data/mongod.service
```diff
@@ -0,0 +1,33 @@
...
+# processes/threads
+LimitNPROC=64000
+# locked memory
+LimitMEMLOCK=infinity
+# total threads (user+kernel)
+TasksMax=infinity
+TasksAccounting=false
+
+# Recommended limits for mongod as specified in
+# https://docs.mongodb.com/manual/reference/ulimit/#recommended-ulimit-settings
```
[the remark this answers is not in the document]
MiaAltieri: I removed this file all together, so no need for this anymore :)

src/charm.py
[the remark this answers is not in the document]
marcoppenheimer: (IMO I think it might be best to read from /lib/systemd/system/mongod.service, modify it, then save it the modified version in /etc/systemd/system/mongod.service)

Yes 👍🏽

src/charm.py
```diff
@@ -206,24 +205,26 @@ def _on_install(self, _) -> None:
...
-        # mongod.service file passes commandline args to mongod service
-        shutil.copy("src/data/mongod.service", "/etc/systemd/system/mongod.service")
-
-        with open("/etc/systemd/system/mongod.service", "r") as service_template:
+        with open("/lib/systemd/system/mongod.service", "r") as service_template:
             mongodb_service = service_template.readlines()
 
-        mongodb_service[MONGO_EXEC_LINE] = mongod_start_args
+        # replace start command with our parameterized one
+        for i in range(len(mongodb_service)):
```
WRFitch, line 212: you can use `enumerate` here: 
```suggestion
for index, line in enumerate(mongodb_service):
````

File: src/charm.py [hunk 10]
```diff
@@ -206,24 +205,26 @@ def _on_install(self, _) -> None:
             ]
         )
 
-        # mongod.service file passes commandline args to mongod service
-        shutil.copy("src/data/mongod.service", "/etc/systemd/system/mongod.service")
-
-        with open("/etc/systemd/system/mongod.service", "r") as service_template:
+        with open("/lib/systemd/system/mongod.service", "r") as service_template:
             mongodb_service = service_template.readlines()
```
WRFitch, line 209: nitpicking here, but could this be renamed `mongodb_service_file` or similar?

src/charm.py
```diff
@@ -206,24 +205,26 @@ def _on_install(self, _) -> None:
...
 
-        mongodb_service[MONGO_EXEC_LINE] = mongod_start_args
+        # replace start command with our parameterized one
+        for i in range(len(mongodb_service)):
+            if "ExecStart" in mongodb_service[i]:
+                mongodb_service[i] = mongod_start_args
 
+        # systemd gives files in /etc/systemd/system/ precedence over those in /lib/systemd/system/
```
WRFitch, line 216: It might be worth removing the file in /lib if it's not being used.

src/charm.py
```diff
@@ -186,22 +190,47 @@ def _on_install(self, _) -> None:
...
+        with open("/etc/systemd/system/mongod.service", "w") as service_file:
+            service_file.writelines(mongodb_service)
+
+        # mongod requires permissions to /data/db
+        mongodb_user = pwd.getpwnam(MONGO_USER)
+        os.chown(MONGO_DATA_DIR, mongodb_user.pw_uid, mongodb_user.pw_gid)
+
+        # changes to service files are only applied after reloading
+        systemd.daemon_reload()
+
```
delgod, line 227: do we need to handle errors here?
like writing errors, chown, reload errors

delgod (approved):

src/charm.py
MiaAltieri: Awesome! This is a great explaination @delgod thanks for making this clear for me :)

paulomach (approved):
LGTM, got my question answered

src/charm.py
MiaAltieri: resolving due to our Mattermost chat.

delgod (approved):

WRFitch (approved):

PR #38, opened 2022-06-10
Update to K8s Library

MiaAltieri:
### Problem

MongoDB Machine Charm uses different library code than the K8s charm.

### Solution

Migrate Machine Charm to k8s charm library. 

### Context

This is one of the "Subsequent PRs" described by the previous PR #37 

### Release Notes

Changed the first set of `self._mongo` usages to `MongoDBConnection` and updated unit tests appropriately. ie:
- `mongo.remove_replset_member`
- `mongo.primary`

This also required adding a `primary` helper function to the MongoDB library, as the Machine charm requires this functionality.

### Testing

Unit tests were updated (and some unit tests were removed since their scenarios are tested with integration tests.)

Integration tests were not changed but instead verify that migrating the the k8s lib does not change anything.

### Subsequent PRs
To minimise the size of this PR changes for the remaining usages of `self._mongo` will be updated in a future PR. This is due to the fact that the remain updates will not be as trivial as those in this PR and will require significant refactoring.


src/charm.py
```diff
@@ -180,12 +179,10 @@ def _on_mongodb_relation_handler(self, event: ops.charm.RelationEvent) -> None:
...
         logger.debug("unit: %s is ready to join the replica set", calling_unit.name)
-        self._reconfigure(event)
-
-    def _reconfigure(self, event: ops.charm.RelationEvent) -> None:
-        """Adds/removes RelationEvent triggering unit from the replica set.
+        self._add_replica(event)
 
-        Note: Removal funcationality will be implemented in the following PR.
+    def _add_replica(self, event: ops.charm.RelationEvent) -> None:
+        """Adds RelationEvent triggering unit from the replica set.
```
marceloneppel, line 185: Shouldn't it be `to the replica set`?

marceloneppel (approved):
LGTM!

src/charm.py
MiaAltieri: yes it should be! Great eye : )

WRFitch (approved):

PR #39, opened 2022-06-13
Finalise Migration to K8s MongoDB Library. 

MiaAltieri:
### Prelude 
A lot of code/files were removed which explains the big diff here. :) 

A **lot of the code that was added here is borrowed** from the [k8s-charm](https://github.com/delgod/mongodb-k8s-operator); which code is borrowed is outlined in the "Borrowed Code" section of this. 

### Problem

MongoDB Machine Charm uses different library code than the K8s charm.

### Solution

Migrate Machine Charm to k8s charm library. 

### Context

This is one of the "Subsequent PRs" described by the previous PR #37 

### Release Notes

1. Changed the remaining usages of `self._mongo` to `MongoDBConnection` and updated unit tests appropriately. 
2. Removed dead code related to old library `mongod_helpers`

### Testing

Unit tests were updated (and some unit tests were removed since their scenarios are tested with integration tests.)

Integration tests were not changed but instead verify that migrating the the k8s lib does not change anything.

### Borrowed Code
- code for `_on_mongodb_relation_handler` was borrowed from the k8s charm [reconfigure](https://github.com/delgod/mongodb-k8s-operator/blob/c6f23b282b72cd0bf39bee03a09b0d0ae7494b74/src/charm.py#L151)
- unit tests for `_on_mongodb_relation_handler` was borrowed from [the k8s charm](https://github.com/delgod/mongodb-k8s-operator/blob/main/tests/unit/test_charm.py)


tests/unit/test_charm.py
```diff
@@ -316,163 +307,104 @@ def test_unit_ips(self):
...
+                    # simulate 2nd MongoDB unit joining
+                    self.harness.add_relation_unit(rel.id, "mongodb/1")
+                    self.harness.update_relation_data(rel.id, "mongodb/1", PEER_ADDR)
 
-            # verify we go into waiting and don't reconfigure
-            self.assertEqual(
-                self.harness.charm.unit.status, WaitingStatus("waiting to reconfigure replica set")
-            )
-            mongodb_reconfigure.assert_not_called()
+                if departed:
```
marceloneppel, line 372: One question: why do you need to have this if/else block separated from the one in L364? (the condition is the same).

marceloneppel (approved):
LGTM! I have just one question in the unit tests.

tests/unit/test_charm.py
```diff
@@ -316,163 +307,104 @@ def test_unit_ips(self):
...
-                )
-
-            else:
-                mongodb_reconfigure.assert_not_called()
+        # verify we go into waiting and don't reconfigure
+        self.assertTrue(isinstance(self.harness.charm.unit.status, WaitingStatus))
+        connection.return_value.__enter__.return_value.add_replset_member.assert_not_called()
 
     @patch_network_get(private_address="1.1.1.1")
+    @patch("ops.framework.EventBase.defer")
```
zmraul, line 343: Not completely related to the PR.

With imports from the charm module you could potentially keep it here:

```python
from ops.framework import EventBase
. . .
# tests
. . .
@patch.object(EventBase, "defer")
def test_relation_joined_get_members_failure(defer):
```

Or in places that you are using `@patch("ops.model.ActiveStatus")`. Since `ActiveStatus` is being imported at the top already, you can just do:
```python
@patch.object(ActiveStatus)
def test(active_status):
```

tests/unit/test_charm.py
MiaAltieri: You're so smart!

They shouldn't be separated :)

src/charm.py
```diff
@@ -160,49 +158,33 @@ def _on_mongodb_relation_handler(self, event: ops.charm.RelationEvent) -> None:
...
+                            event.defer()
+                            return
+                    mongo.add_replset_member(member)
+            except NotReadyError:
+                self.unit.status = WaitingStatus("waiting to reconfigure replica set")
+                logger.info("Deferring reconfigure: another member doing sync right now")
+                event.defer()
+            except PyMongoError as e:
+                self.unit.status = WaitingStatus("waiting to reconfigure replica set")
+                logger.info("Deferring reconfigure: error=%r", e)
```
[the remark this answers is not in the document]
MiaAltieri: I agree! Thanks for pointing this out :)

PR #42, opened 2022-06-15
Relation Infrastructure

MiaAltieri:
### Prelude 
This is a big PR since we are adding a ton of new files! 

NOTE: **almost ALL of the code that was added here is borrowed** from the [k8s-charm](https://github.com/delgod/mongodb-k8s-operator) and [data-platform-libs](https://github.com/canonical/data-platform-libs); which code is borrowed is outlined in the "FAQ" section of this. 

My thought was to separate this into two PRs, so that it would be easier to tell which code is new and which is borrowed:
1. big one adding files/infra (this PR)
2. new code related to the testing of the relation (next PR)

### Problem

Relations are not implemented.

### Solution

This solution makes use of `lib/charms/mongodb_libs/v0/mongodb_provider.py` by setting up the necessary infrastructure for relations. (As well as the testing infrastructure for relating MongoDB to an application). 

This solution does not test the functionality of the relation. That will come in the next PR. For now the test only verifies that the MongoDB charm can be related to an application via `juju relate`

### Release Notes

Changes in this PR:
- added provider code to `src/charm.py`
- set up testing infra under `tests/integration/relation_tests`

### Testing

Infrastructure for setting up relation integration tests was added. Testing of relations will be done in the next PR. 

### FAQ
What code from this PR is borrowed from the [MongoDB K8s Charm](https://github.com/delgod/mongodb-k8s-operator)?
- `src/charm.py` - changes in this file were directly copied from the k8s charm 

What code from this PR is borrowed from [Data Platform Lib](https://github.com/canonical/data-platform-libs)?
- `tests/integration/relation_tests` - everything from here was copied from data-platform-libs, with minor adjustments made to `conftest.py`, ` test_charm.py` (now ` test_charm_relations.py`) and `helpers.py`

### Subsequent PR
Setting up the necessary infra for this PR resulted in a large diff. To reduce the size of this PR the relation integration tests will be saved for the next PR. These tests will test

1. Relation results in access to MongoDB database
4. Application can write/read/update data
5. After application departs from the relation the application can no longer access the data from 2.


tests/integration/relation_tests/application-charm/data_platform_libs/v0/database_provides.py
```diff
@@ -0,0 +1,306 @@
...
+        Args:
+            relation_id: the identifier for a particular relation.
+            tls_ca: TLS certification authority.
+        """
+        self._update_relation_data(relation_id, {"tls_ca": tls_ca})
+
+    def set_uris(self, relation_id: int, uris: str) -> None:
+        """Set the database connection URIs in the application relation databag.
+
+        MongoDB, Redis, OpenSearch and Kafka only.
```
grobbie, line 291: remark but it might apply to others in the future, and I guess it also applies to ZooKeeper? Maybe it would be better to say "Set the database connection URIs in the application relation databag.

Applies to applications making use of the connection URI approach, for example MongoDB, Redis, OpenSearch and Kafka."

metadata.yaml
```diff
@@ -9,6 +9,12 @@ summary: A MongoDB operator charm
 series:
   - focal
 
+provides:
+  database:
+    interface: database-client
```
marceloneppel, line 14: This should be changed to be specific to MongoDB.

```suggestion
    interface: mongodb-client
```

config.yaml
```diff
@@ -0,0 +1,10 @@
+# Copyright 2022 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+options:
+  auto-delete:
+    type: boolean
+    description: |
+        When a relation is removed, auto-delete ensures that any relevant databases
+        associated with the relation are also removed
+    default: true
```
marceloneppel, line 10: I think it's better to change the default value to `false` to avoid the database being deleted when it's not wanted if some operator incorrectly removes the relation.

```suggestion
    default: false
```

Mykola change it on https://github.com/delgod/mongodb-k8s-operator/pull/13/files.

tests/integration/relation_tests/application-charm/data_platform_libs/v0/database_provides.py
```diff
@@ -0,0 +1,306 @@
+# Copyright 2022 Canonical Ltd.
```
marceloneppel, line 1: This file can be removed as you're using the mongodb provider library.

tests/integration/relation_tests/application-charm/metadata.yaml
```diff
@@ -0,0 +1,12 @@
...
+name: application
+description: |
+  Data platform libs application charm used in integration tests.
+summary: |
+  Data platform libs application meant to be used
+  only for testing of the libs in this repository.
+
+requires:
+  database:
+    interface: database-client
```
marceloneppel, line 12: To match the other change, this should also be changed:

```suggestion
    interface: mongodb-client
```

tests/integration/relation_tests/application-charm/data_platform_libs/v0/database_requires.py
```diff
@@ -0,0 +1,312 @@
+# Copyright 2022 Canonical Ltd.
```
marceloneppel, line 1: This file can be removed as you also have it in a different folder (`tests/integration/relation_tests/application-charm/lib/charms/data_platform_libs/v0/database_requires.py`, which is the right path).

marceloneppel (changes requested):

tests/integration/relation_tests/application-charm/lib/charms/data_platform_libs/v0/database_requires.py
```diff
@@ -0,0 +1,312 @@
...
+                for all relation instances (indexed by the relation id).
+        """
+        data = {}
+        for relation in self.relations:
+            data[relation.id] = {
+                key: value for key, value in relation.data[relation.app].items() if key != "data"
+            }
+        return data
+
+    def _update_relation_data(self, relation_id: int, data: dict) -> None:
```
grobbie, line 253: Hey, I don't know jack about Python really, but I do know Java. In Java, this method would typically be prefixed with `private` to denote it's scope, which is pretty evidently limited to the class. In Python, I understand that declaring private scope is done by prefixing a double underscore to the method name. Just wondering why we are not using that convention for this stuff?

grobbie: maybe something to do with the `self.framework.observe()` stuff triggering this event? If so feel free to ignore this comment!

marceloneppel (approved):

tests/integration/relation_tests/application-charm/lib/charms/data_platform_libs/v0/database_requires.py
MiaAltieri: Hi @grobbie ! I can't take credit for this awesome code, it was merely ported from [database-libs](). So I may not be answering this question perfectly. But I think on the DPE team we have been using a single underscore to denote private functions. 

I will admit that I don't know much about the knitty-gritty syntax of Python since I am a somewhat new professional, but I think it might be a contested point that sometimes private functions use `__` and sometimes they use `_`. [This stackoverflow post](https://stackoverflow.com/questions/1301346/what-is-the-meaning-of-single-and-double-underscore-before-an-object-name#:~:text=Single%20underscore%20at%20the%20beginning,not%20part%20of%20the%20API.) explains the difference between the two, but I have seen that double underscore can be used for private (but again on the DPE team we've been using a single underscore)

tests/integration/relation_tests/application-charm/data_platform_libs/v0/database_provides.py
MiaAltieri: Hi @grobbie ! I can't take credit for this awesome code, it was merely ported from [database-libs](https://github.com/canonical/data-platform-libs), which is the source of this code.

Note that in the future we hope use library fetching so that the file doesn't have to exist in this repository, but instead gets fetched and then deleted in the integration tests.

I can make the change here, but in order for it to take effect for all charms I will have to make this change in the source code for this file (ie in the [data platform libs](https://github.com/canonical/data-platform-libs)) for it to take effect for everyone. Sound good?

PR #48, opened 2022-07-07
Legacy follow up

MiaAltieri:
NOTE: please do not review until the previous PR has been merged

### Problem

Legacy Relations are supported but do not handle edge cases (ie new relation joining while the legacy relation interface is being used and vice versa, this is significant because each interface does/doesn't use auth)

### Solution

Addresses the basic concerns of legacy relations edge cases with respect to auth enablement/disablement

### Context

1. When disabling auth in MongoDB it is not possible to keep cluster auth, hence on disable of both auth and cluster auth

### Release Notes

1. In legacy relation code add checks for 
    1. Is there an app that requires auth, if so go into blocked
    2. Has auth already been turned off? If so no need to stop and start again
2. In new relation provider code add check for 
    1. Is auth turned off and a new app wants to join that uses auth?  
3. Add a check for when a new unit joins that it appropriately joins with/without auth

### Testing

Integration tests verify
1. charm goes into blocked if trying to use both new relation interface and legacy relation interface
2. units added during scaling correctly disable auth if legacy interface is being used

### Future PRs
1. An organisational PR to find best places for:
    1. Legacy relation <— this is dependent on the needs of the K8s charm
    2. Mutual code between legacy code and charm.py


lib/charms/mongodb_libs/v0/mongodb_provider.py
```diff
@@ -159,25 +180,37 @@ def _on_legacy_relation_created(self, event):
...
+        if self._auth_enabled:
+            try:
+                logger.debug("Disabling authentication.")
+                self._stop_mongod_service()
+                self._update_mongod_service(auth=False)
+                self._start_mongod_service()
+            except systemd.SystemdError:
+                self.charm.unit.status = BlockedStatus("couldn't restart MongoDB")
+                return
+
```
marcoppenheimer, line 208: Could you generalise `oversee_users` and call that here?

lib/charms/mongodb_libs/v0/mongodb_provider.py
```diff
@@ -187,10 +220,35 @@ def _on_legacy_relation_joined(self, event):
...
+        # the relation data for all units, hence all units setting the relation data and not just
+        # the leader
         relation = self.model.get_relation(REL_NAME, event.relation.id)
         relation.data[self.charm.unit].update(updates)
 
+    def get_users_from_legacy_relations(self):
+        """Return usernames for legacy relation."""
+        relations = self.model.relations[LEGACY_REL_NAME]
+        logger.debug(relations)
+        return set([self._get_username_from_relation_id(relation.id) for relation in relations])
```
marcoppenheimer, line 233: Just a fun tip, you can use set-comprehension here:

`{self._get_username_from_relation_id(relation.id) for relation in relations}`

lib/charms/mongodb_libs/v0/mongodb_provider.py
```diff
@@ -187,10 +220,35 @@ def _on_legacy_relation_joined(self, event):
...
+            return False
+
+        with open("/lib/systemd/system/mongod.service", "r") as mongodb_service_file:
+            mongodb_service = mongodb_service_file.readlines()
+
+        for index, line in enumerate(mongodb_service):
+            # ExecStart contains the line with the arguments to start mongod service.
+            if "ExecStart" in line:
+                if "--auth" in line:
+                    return True
```
marcoppenheimer, line 248: Am I right in thinking you don't use index here?

marcoppenheimer: you can also do:

`if "ExecStart" and "--auth" in line:`

tests/integration/relation_tests/legacy_relations/test_charm_legacy_relations.py
```diff
@@ -119,3 +125,86 @@ async def test_legacy_db_ops(ops_test: OpsTest) -> None:
...
+
+    # add legacy relation
+    await ops_test.model.add_relation(GRAYLOG_APP_NAME, ANOTHER_DATABASE_APP_NAME)
+    await ops_test.model.wait_for_idle(
+        apps=[ANOTHER_DATABASE_APP_NAME], status="blocked", timeout=1000
+    )
+
+    assert (
+        ops_test.model.applications[ANOTHER_DATABASE_APP_NAME].units[0].workload_status_message
+        == "cannot have both legacy and new relations"
```
marcoppenheimer, line 197: Nit, also personal preference. Might be better to avoid testing for specific status messages as these are likely to change. 

Maybe look for `isinstance`, and `workload_status_message is not None` or something.

marcoppenheimer (approved):
Looks really good, bunch of Pythony rubbish comments, nothing major.

lib/charms/mongodb_libs/v0/mongodb_provider.py
marceloneppel: I guess `if "ExecStart" and "--auth" in line:` always evaluates `"ExecStart"` as True, because of the string content.

lib/charms/mongodb_libs/v0/mongodb_provider.py
MiaAltieri: For now I am going to keep these separate, I think it will make things simpler when it comes time to remove everything related to legacy relations. Also `oversee_users` has a bunch of new relation code that would be challenging/annoying for me to generalise

lib/charms/mongodb_libs/v0/mongodb_provider.py
MiaAltieri: why didn't i see this earlier 🤦

tests/integration/relation_tests/legacy_relations/test_charm_legacy_relations.py
MiaAltieri: good point, I'll remove this, but keep the blocked check

lib/charms/mongodb_libs/v0/mongodb_provider.py
zmraul: Small change, `if "ExecStart" and "--auth" in line` will evaluate `ExecStart` as always true since it's a non empty string:
```
if "ExecStart" in line and "--auth" in line:
```
 or 
```
if any(i in line for i in ("ExecStart", "--auth"))
```

src/charm.py
```diff
@@ -204,27 +204,42 @@ def _on_install(self, event) -> None:
...
+            "--bind_ip",
+            f"localhost,{machine_ip}",
+            # part of replicaset
+            "--replSet",
+            f"{self.app.name}",
+        ]
+
+        # if a new unit is joining a cluster with a legacy relation it should start without auth
+        requires_auth = len(self.client_relations.get_users_from_legacy_relations()) == 0
+        if requires_auth:
```
zmraul, line 226: I think you can leave this as `requires_auth =self.client_relations.get_users_from_legacy_relations()`
An empty set will evaluate to False

lib/charms/mongodb_libs/v0/mongodb_provider.py
zmraul: Heh, race conditions on github comments :)

MiaAltieri: opting for `if "ExecStart" in line and "--auth" in line:` thanks for the help :)

src/charm.py
```diff
@@ -204,27 +204,41 @@ def _on_install(self, event) -> None:
...
+            # bind to localhost and external interfaces
+            "--bind_ip",
+            f"localhost,{machine_ip}",
+            # part of replicaset
+            "--replSet",
+            f"{self.app.name}",
+        ]
+
+        # if a new unit is joining a cluster with a legacy relation it should start without auth
+        if self.client_relations.get_users_from_legacy_relations():
```
marceloneppel, line 225: Shouldn't it be `if not self.client_relations.get_users_from_legacy_relations():`? The `auth` flag is being added inside this if block if there are active legacy relations.

MiaAltieri: YES great catch!

marceloneppel (approved):
LGTM!

zmraul (approved):

PR #49, opened 2022-07-07
Feature/auto charmhub publish

paulomach:
Auto publish to charmhub latest/edge github workflow


.github/workflows/release.yaml
```diff
@@ -0,0 +1,79 @@
+name: Release to latest/edge
```
MiaAltieri, line 1: This looks great! Thanks for doing this Paulo! Quick question! Should this be merged after legacy relations are finished?

PR #53, opened 2022-07-15
Format MongoDB lib code

MiaAltieri:
## Problem

Previously `tox` ignored all lib code, but since it "owns" the `mongodb` lib code it should format that accordingly. 

## Solution

Update `tox` to pay attention and format the `mongodb` lib code


WRFitch (approved):
LGTM!

PR #60, opened 2022-07-22
HA-Tests now accept a provided cluster

MiaAltieri:
New functionality in integration tests allows for the  user to do: `tox -e ha-integration -- --model=<model_name>` and the ha-tests will be run on their provided cluster.

The functionality is flexible in that:
1. The cluster being tested with does not need to be named exactly `mongodb`
2. The cluster being tested does not need to have a specified number of units

### Release Notes

`tests/integration/ha_tests/test_ha.py` 
- updated to use the deployment name rather than the hardcoded APP_NAME
- in scale down minority it computes the minority units

`tests/integration/ha_tests/helpers.py`
- functions updated to use the deployment name rather than the hardcoded APP_NAME
- `cluster_name` identifies the deployments name


tests/integration/ha_tests/helpers.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2021 Canonical Ltd.
+# # Copyright 2021 Canonical Ltd.
```
MiaAltieri, line 1: remove extra #

tests/integration/ha_tests/test_ha.py
```diff
@@ -72,32 +81,43 @@ async def test_scale_down_capablities(ops_test: OpsTest) -> None:
...
-        leader_unit.name,
-        non_leader_unit.name,
-    )
+    for _ in range(minority_count):
+        unit_to_remove = avail_units.pop()
+        deleted_unit_ips.append(unit_to_remove.public_address)
+        units_to_remove.append(unit_to_remove.name)
+
+    # destroy units simulatenously
+    # pass a lambda function here.
```
MiaAltieri, line 109: `# pass a lambda function here.`?? remove

marceloneppel (approved):
LGTM! The implementation is so interesting. I am looking forward to do the same in the PostgreSQL charms.

tests/integration/ha_tests/helpers.py
```diff
@@ -160,3 +145,43 @@ async def retrieve_entries(ops_test, app, db_name, collection_name, query_field)
...
+    provided_cluster = await cluster_name(ops_test)
+    if not provided_cluster:
+        return UNIT_IDS
+    unit_ids = [
+        unit.name.split("/")[1] for unit in ops_test.model.applications[provided_cluster].units
+    ]
+    return unit_ids
+
+
+async def cluster_name(ops_test: OpsTest) -> str:
```
paulomach, line 172: I found it a bit confusing calling this `cluster_name`, since what you are returning is the name the application was given on deployment for a given model. On contexts other than mongo, the cluster name can be a configuration of the workload itself.
How about calling this just `app_name`?

paulomach (approved):
I gave it just a small suggestion, otherwise LGTM

tests/integration/ha_tests/helpers.py
MiaAltieri: Love that! Will do, this is a really good suggestion!!

jnsgruk:
Nice work here, it would be good to augment CONTRIBUTING.me with some instructions on how to run the various test suites and their different modes.

If we standardise this methodology across the team, the text can likely be reused :)

PR #61, opened 2022-07-22
File re-org for integration testing + updates to tox

MiaAltieri:
Content in files `tests/integration/*` is not new. Please do not review their changes, this PR just splits up HA and charm tests and adds to the tox file. Only review of the `tox.ini` file is needed.


tox.ini
```diff
@@ -64,13 +64,32 @@ commands =
...
+[testenv:charm-integration]
+description = Run charm integration tests
 deps =
     pytest
     juju
     pytest-operator
     -r{toxinidir}/requirements.txt
 commands =
-    pytest -v --tb native --ignore={[vars]tst_path}unit --log-cli-level=INFO -s {posargs} --durations=0 --asyncio-mode=auto
+    pytest -v --tb native  {[vars]tst_path}integration/test_charm.py --log-cli-level=INFO -s {posargs} --durations=0 --asyncio-mode=auto
```
Mehdi-Bendriss, line 75: Optional - could be worth using `--ignore` in the future instead of pointing to a test file, in case new the integration test scope grows and new files get added in addition to the `test_charm.py`

MiaAltieri: True 🤔

PR #66, opened 2022-07-29
Test storage re use

MiaAltieri:
### Solution

Tests that when a unit `unit B` gets removed (without removing its storage `storage B`) and a new unit `unit C` re-uses `storage B`, that MongoDB correctly re-uses that storage. (ie the mongo daemon (`mongod`) doesn't just copy all the data from the primary to `unit C` but instead only copies the necessary data to `unit C` based on the re-used `storage B`)

### Specialised Knowledge

1. Storage re-attachment cannot be done on rootfs file systems, see [this convo](https://chat.canonical.com/canonical/pl/eirmfogfx3rmufmom9thjx6pwr)
2. python-libjuju does not have functionality for storage based operations see [issue](https://github.com/juju/python-libjuju/issues/695) and [issue](https://github.com/juju/python-libjuju/issues/694)
3. To verify that MongoDB properly re-uses storage log messages for startup are used:
- If member transitions to STARTUP2 from STARTUP then it is syncing/getting data from primary.
- If member transitions to STARTUP2 from REMOVED then it is re-using the storage we provided.

### Release Notes

1. Created storage helper functions in `tests/integration/ha_tests/helpers.py`
2. Tested storage re-use in `tests/integration/ha_tests/test_ha.py`

### Testing

Since GitHub CI for this charm uses lxd for integration testing rootfs storage is used so this test is not run on the GitHub CI, therefore it must be tested somewhere else. To verify the test works:
1. Create an AWS/GCP/OpenStack cloud. Networking for AWS/GCP/OpenStack instances must be enabled.
4. Create a storage pool: `juju create-storage-pool mias-ebs ebs volume-type=standard`
5. Deploy 1 or more units of application with said storage pool: `juju deploy ./*charm --storage mongodb=mias-ebs,7G,1`
6. Run tests in the model where you deployed the app `tox -e ha-integration -- --model=default`


tests/integration/ha_tests/helpers.py
```diff
@@ -1,6 +1,8 @@
-# Copyright 2021 Canonical Ltd.
+# # Copyright 2021 Canonical Ltd.
```
WRFitch, line 1: typo here

tests/integration/ha_tests/helpers.py
```diff
@@ -185,3 +175,105 @@ async def app_name(ops_test: OpsTest) -> str:
...
+
+
+def storage_type(ops_test, app):
+    """Retrieves type of storage associated with an application.
+
+    Note: this function exists as a temporary solution until this issue is resolved:
+    https://github.com/juju/python-libjuju/issues/694
+    """
+    model_name = ops_test.model.info.name
+    proc = subprocess.check_output(f"juju storage --model={model_name}".split())
```
WRFitch, line 187: there's actually an ops_test function for running juju code that can make this a little cleaner:
```suggestion
    get_storage = await ops_test.juju("storage", f"--model={model_name}")
    result = get_storage[1]
```

You could also add the --format=json flag in there, load it into a python dict and then access what you want programmatically. This suggestion might be overcomplicating it, though.

tests/integration/ha_tests/helpers.py
```diff
@@ -185,3 +175,105 @@ async def app_name(ops_test: OpsTest) -> str:
...
+    for line in proc.splitlines():
+        if "Storage" in line:
+            continue
+
+        if len(line) == 0:
+            continue
+
+        if "detached" in line:
+            continue
+
```
WRFitch, line 223: I understand this is a temp solution, but you could write a helper function to get_unit_storage(), return line if line==unit_name, and then index the output to get ids and storage types.

tests/integration/ha_tests/test_ha.py
```diff
@@ -40,6 +40,38 @@ async def test_build_and_deploy(ops_test: OpsTest) -> None:
...
+        await ops_test.model.applications[app].add_unit(count=1)
+        await ops_test.model.wait_for_idle(apps=[app], status="active", timeout=1000)
+
+    # remove a unit and attach it's storage to a new unit
+    unit = ops_test.model.applications[app].units[0]
+    unit_storage_id = storage_id(ops_test, unit.name)
+    expected_units = len(ops_test.model.applications[app].units) - 1
+    await ops_test.model.destroy_unit(unit.name)
+    await ops_test.model.wait_for_idle(apps=[app], status="active", timeout=1000)
+    assert len(ops_test.model.applications[app].units) == expected_units
```
WRFitch, line 67:
```suggestion
    await ops_test.model.wait_for_idle(apps=[app], status="active", timeout=1000, wait_for_exact_units=expected_units)
```

WRFitch (commented):
Looks great! I have a couple of minor notes, though:

tests/integration/ha_tests/helpers.py
zmraul: +1 to using `ops_test.juju()`

tests/integration/ha_tests/test_ha.py
zmraul: This can also be applied to L88, L148 and L247

zmraul (approved):
It looks good. I second the suggestions Will made :)

tests/integration/ha_tests/helpers.py
paulomach: I'll chime in here and argue *in favor* of the current implementation for four reasons:
1. `ops_test.juju` just wraps an async way to call a sub process
2. the output of `--format=json` is way more verbose than the default one, which is very straightforward to parse
3. this is a (temporary?) workaround
4. it's done and it works

tests/integration/ha_tests/helpers.py
paulomach: I'll argue as above

tests/integration/ha_tests/test_ha.py
```diff
@@ -2,7 +2,6 @@
 # Copyright 2021 Canonical Ltd.
```
paulomach, line 2: bump year?

tests/integration/ha_tests/test_ha.py
```diff
@@ -40,6 +40,38 @@ async def test_build_and_deploy(ops_test: OpsTest) -> None:
...
+    primary, but instead computes a diff between current storage and primary storage.)
+    """
+    app = await app_name(ops_test)
+    if storage_type(ops_test, app) == "rootfs":
+        pytest.skip(
+            "re-use of storage can only be used on deployments with persistent storage not on rootfs deployments"
+        )
+
+    # removing the only replica can be disastrous
+    if len(ops_test.model.applications[app].units) < 2:
```
paulomach, line 57: hmmm, have to replicate this on my stuff ;)

paulomach (approved):
Great work Mia, LGTM

Issue #65, opened 2022-07-28
resolve failing charmhub publishing


MiaAltieri:

MiaAltieri:
Charmhub is failing to publish since tox.ini no longer has the `integration` workflow, see [here](https://github.com/canonical/mongodb-operator/actions/runs/2759294471)

MiaAltieri:
proposed solution: #68

PR #68, opened 2022-07-29
Fixing Charmhub Failure

MiaAltieri:
See Issue #65  
Charmhub is failing to publish since tox.ini no longer has the integration workflow, see [here](https://github.com/canonical/mongodb-operator/actions/runs/2759294471)

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -36,7 +36,8 @@ commands =
 description = Check code against coding style standards
 deps =
     black
-    flake8
+    # flake8 version 5.0.0 & 5.0.1 has issues with `flake8.options.config`
+    flake8==4.0.1
     flake8-docstrings
     flake8-copyright
     flake8-builtins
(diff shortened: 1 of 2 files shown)


tox.ini
```diff
@@ -93,3 +93,13 @@ deps =
...
+
+[testenv:integration]
+description = Run all integration tests
+deps =
+    pytest
+    juju
+    pytest-operator
+    -r{toxinidir}/requirements.txt
+commands =
+    pytest -v --tb native --ignore={[vars]tst_path}unit --log-cli-level=INFO -s {posargs} --durations=0 --asyncio-mode=auto
```
marceloneppel, line 105: Paulo had a good suggestion about movin the asyncio mode to `pyproject.toml`, so it is not needed to specify it in each environment that we have in `tox.ini`.

```toml
[tool.pytest.ini_options]
minversion = "6.0"
log_cli_level = "INFO"
asyncio_mode = "auto"
```

This is optional.

marceloneppel (approved):
LGTM!

WRFitch (approved):

tox.ini
MiaAltieri: great idea! I like this :)

PR #70, opened 2022-07-29
Bug fix: _init_user crash when called twice

MiaAltieri:
This PR contains the following:
- Fix handling the case where `_init_user` is called multiple times due to a failure in the `oversee_users` method
- Unit test for said fix

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -483,7 +483,11 @@ class MongodbOperatorCharm(ops.charm.CharmBase):
     @property
     def app_data(self) -> Dict:
         """Peer relation data object."""
-        return self.model.get_relation(PEER).data[self.app]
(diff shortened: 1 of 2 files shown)


src/charm.py
```diff
@@ -483,7 +483,11 @@ def mongodb_config(self) -> MongoDBConfiguration:
     @property
     def app_data(self) -> Dict:
         """Peer relation data object."""
-        return self.model.get_relation(PEER).data[self.app]
+        relation = self.model.get_relation(PEER)
+        if relation is None:
```
marcoppenheimer, line 487:
```suggestion
        if not relation:
```

marcoppenheimer (approved):

marceloneppel (approved):
LGTM!

Issue #75, opened 2022-08-05
Update status should report a message for non-primary units.


MiaAltieri:
## Problem:
TL;DR: The update status hook of the MongoDB charm incorrectly reports which unit is the "Replica Set Primary". **It reports the old primaries and the current primaries.**

You can see this on a real deployment
```
juju status 
ubuntu@mongo-ha:~/mongodb-operator$ juju status
Model  Controller  Cloud/Region         Version  SLA          Timestamp
neato  overlord    localhost/localhost  2.9.32   unsupported  13:15:07Z

App      Version  Status  Scale  Charm    Channel  Rev  Exposed  Message
mongodb           active      3  mongodb             0  no

Unit        Workload  Agent  Machine  Public address  Ports      Message
mongodb/0*  active    idle   0        10.30.220.92    27017/tcp    Replica set primary
mongodb/1   active    idle   1        10.30.220.211   27017/tcp
mongodb/2   active    idle   2        10.30.220.123   27017/tcp
```
and for some reason `mongodb/0` is no longer the primary and `mongodb/1` is the new primary, then `juju status` will show 
```
juju status 
ubuntu@mongo-ha:~/mongodb-operator$ juju status
Model  Controller  Cloud/Region         Version  SLA          Timestamp
neato  overlord    localhost/localhost  2.9.32   unsupported  13:15:07Z

App      Version  Status  Scale  Charm    Channel  Rev  Exposed  Message
mongodb           active      3  mongodb             0  no

Unit        Workload  Agent  Machine  Public address  Ports      Message
mongodb/0*  active    idle   0        10.30.220.92    27017/tcp    Replica set primary  <----- old status message
mongodb/1   active    idle   1        10.30.220.211   27017/tcp      Replica set primary
mongodb/2   active    idle   2        10.30.220.123   27017/tcp
```

This is **wrong** we would instead expect to see is:
```
juju status 
ubuntu@mongo-ha:~/mongodb-operator$ juju status
Model  Controller  Cloud/Region         Version  SLA          Timestamp
neato  overlord    localhost/localhost  2.9.32   unsupported  13:15:07Z

App      Version  Status  Scale  Charm    Channel  Rev  Exposed  Message
mongodb           active      3  mongodb             0  no

Unit        Workload  Agent  Machine  Public address  Ports      Message
mongodb/0*  active    idle   0        10.30.220.92    27017/tcp    
[...]

PR #76, opened 2022-08-05
Status update for non primary units

gokhancetinkaya:


gokhancetinkaya (commented):
#75

MiaAltieri (approved):
looks go to me, thanks for doing this :)

src/charm.py
```diff
@@ -277,6 +277,8 @@ def _on_update_status(self, _):
         # if unit is primary then update status
         if self._primary == self.unit.name:
             self.unit.status = ActiveStatus("Replica set primary")
+        else:
+            self.unit.status = ActiveStatus("")
```
paulomach, line 281: nit, if no message is set, just:
```suggestion
            self.unit.status = ActiveStatus()
```

paulomach (approved):
minor nit, lgtm

MiaAltieri:
you can fix these failing tests by merging the main branch into yours. The new juju version causes issues retrieving the results of an action

tests/integration/test_charm.py
```diff
@@ -120,3 +120,26 @@ async def test_get_primary_action(ops_test: OpsTest) -> None:
 
         # assert get-primary returned the right primary
         assert identified_primary == expected_primary
+
+
+async def test_exactly_one_primary_reported_by_juju(ops_test: OpsTest) -> None:
+    primary = []
```
MiaAltieri, line 126: since this variable is defined outside the scope of the function `check`, it will increase to a length of 2 when a new primary gets elected and hence the call to `check()` on line 142 will fail.

Perhaps instead you can put `primary` within the scope of `check` and then return the variable `primary`

tests/integration/test_charm.py
```diff
@@ -120,3 +120,26 @@ async def test_get_primary_action(ops_test: OpsTest) -> None:
...
+    
+    def check():
```
MiaAltieri, line 128:
```suggestion
    def check_juju_status_primaries():
```

tests/integration/test_charm.py
```diff
@@ -120,3 +120,26 @@ async def test_get_primary_action(ops_test: OpsTest) -> None:
...
+        for unit in ops_test.model.applications[app].units:
+            if self.unit.status = "Replica set primary"
+            primary.append(unit)
+        assert len(primary) == 1, "Multiple replica set primaries detected"
+
+    check()
+
+    # kill the replica set primary
+    await kill_unit_process(ops_test, primary[0], kill_code="SIGKILL")
```
MiaAltieri, line 137: you will need to import ` kill_unit_process` from the file which it belongs to

MiaAltieri (changes requested):
Hi great work here! you'll need to make some updates to your tests so they pass.

You can run the tests on your machine with:
`tox -e fmt; tox -e lint` <--- cleans the file then runs the linter
`tox -e charm-integration` <--- runs the tests in the file `test/integration/test_charm.py`

Its best to run these tests before pushing to GitHub to verify they run on your machine first. [here](https://github.com/canonical/mongodb-operator/blob/main/CONTRIBUTING.md) is a doc that goes over the steps to to test the charm :)

tests/integration/test_charm.py
```diff
@@ -120,3 +124,57 @@ async def test_get_primary_action(ops_test: OpsTest) -> None:
...
+
+        async with ops_test.fast_forward():
+            time.sleep(20)
+        
+        for unit in ops_test.model.applications[app].units:
+            unit_messages[unit.entity_id] = unit.workload_status_message
+
+        return(unit_messages)
+
+    def single_primary(unit_messages):
```
MiaAltieri, line 145:
```suggestion
    def juju_reports_one_primary(unit_messages):
```

tests/integration/test_charm.py
```diff
@@ -120,3 +124,57 @@ async def test_get_primary_action(ops_test: OpsTest) -> None:
...
+        """Confirms there is only one replica set primary unit"""
```
MiaAltieri, line 146:
```suggestion
        """Confirms there is only one replica set primary unit."""
```

tests/integration/test_charm.py
```diff
@@ -120,3 +124,57 @@ async def test_get_primary_action(ops_test: OpsTest) -> None:
...
+        count = 0
+        for value in unit_messages:
+            if unit_messages[value] == "Replica set primary":
+                count += 1
+
+        if count > 1:
+            assert count == 1, "Multiple replica set primary units detected"
+        elif count < 1:
+            assert count == 1, "No replica set primary units detected"
```
MiaAltieri, line 155: Thorough, but it's enough to do:
``
```suggestion
        assert count == 1, f"juju is expected to report one primary not {count} primaries"
```

tests/integration/test_charm.py
```diff
@@ -120,3 +124,57 @@ async def test_get_primary_action(ops_test: OpsTest) -> None:
...
+
+    # collect unit status messages
+    unit_messages = await get_unit_messages()
+
+    # confirm there is only one replica set primary unit
+    single_primary(unit_messages)
+
+    # kill the replica set primary unit
```
MiaAltieri, line 163:
```suggestion
    # kill the mongod process on the replica set primary unit to force a re-election.
```

tests/integration/test_charm.py
```diff
@@ -120,3 +124,57 @@ async def test_get_primary_action(ops_test: OpsTest) -> None:
...
+        app = await app_name(ops_test)
+        unit_messages = {}
+
+        async with ops_test.fast_forward():
+            time.sleep(20)
+        
+        for unit in ops_test.model.applications[app].units:
+            unit_messages[unit.entity_id] = unit.workload_status_message
+
+        return(unit_messages)
```
MiaAltieri, line 143: According to [the docs](https://github.com/charmed-kubernetes/pytest-operator/blob/main/docs/reference.md):
`ops_test.fast_forward():` "Temporarily speed up update-status firing rate for the current model. Returns an async context manager that temporarily sets update-status firing rate to fast_interval. If provided, when the context exits the update-status firing rate will be set to slow_interval. Otherwise, it will be set to the previous value."

Since it is used as a [context manager](https://book.pythontips.com/en/latest/context_managers.html) here it will only fast forward things that are indented underneath it. (What you have here works, but I think this might look a little nicer). Try this instead:
```suggestion
        async with ops_test.fast_forward(fast_interval=f"{STATUS_INTERVAL}s"):
            time.sleep(STATUS_INTERVAL)
            for unit in ops_test.model.applications[app].units:
                unit_messages[unit.entity_id] = unit.workload_status_message

        return(unit_messages)
```
with STATUS_INTERVAL as a constant at the top of the time STATUS_INTERVAL=5

MiaAltieri (changes requested):

MiaAltieri (approved):

PR #77, opened 2022-08-08
Self healing tests pt1

MiaAltieri:
### Release Notes

This PR adds and tests self-healing.
1. Self healing was added via adding restart options to the service file, this is seen in `lib/charms/mongodb_libs/v0/machine_helpers.py`
2. This performs three self healing tests "kill DB process", "freeze DB process", and "graceful restart of one instance."

### Specialised Knowledge

1. `MongoDBClient` will hang on insert unless `socketTimeoutMS=5000` is specified.
2. `systemd` restarts the service via options in the Service File. We specify `Restart=always`, as to restart the DB process anytime it gets shutdown (ie SIGTERM), not only on failure (ie SIGKILL)
3. Pausing and unpausing the primary can lead to duplicate writes, hence in this test we ensure we don't miss any writes, but we do not check if there are no duplicates.

### Release Notes

1. Added self healing functionality in `tests/integration/ha_tests/helpers.py`
2. Tests added in `tests/integration/ha_tests/test_ha.py`

### Testing

Tests 
1. kill DB process
2. freeze DB process
3. graceful restart of one instance

### DIFF
--- a/tests/integration/ha_tests/helpers.py
+++ b/tests/integration/ha_tests/helpers.py
@@ -1,6 +1,7 @@
-# Copyright 2021 Canonical Ltd.
+# Copyright 2022 Canonical Ltd.
 # See LICENSE file for licensing details.
 
+import subprocess
 from pathlib import Path
 from typing import List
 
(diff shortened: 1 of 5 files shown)


tests/integration/ha_tests/helpers.py
```diff
@@ -378,3 +453,62 @@ async def insert_focal_to_cluster(ops_test: OpsTest) -> None:
...
+        await ops_test.model.wait_for_idle(apps=[app], status="active", timeout=1000)
+
+    # find the DB process on the unit and kill it
+    find_pid = f" run --unit {unit_name} ps aux | grep {DB_PROCESS}"
+    output = await ops_test.juju(*find_pid.split())
+    pid = output[1].split()[1]
+    kill_cmd = f"run --unit {unit_name} -- kill -s {kill_code} {pid}"
+    output = await ops_test.juju(*kill_cmd.split())
+
+    if not output[0] == 0:
```
shayancanonical, line 473: if output[0] != 0 seems cleaner here

shayancanonical (approved):

tests/integration/ha_tests/helpers.py
```diff
@@ -378,3 +453,62 @@ async def insert_focal_to_cluster(ops_test: OpsTest) -> None:
...
+    app = await app_name(ops_test)
+    if len(ops_test.model.applications[app].units) < 2:
+        await ops_test.model.applications[app].add_unit(count=1)
+        await ops_test.model.wait_for_idle(apps=[app], status="active", timeout=1000)
+
+    # find the DB process on the unit and kill it
+    find_pid = f" run --unit {unit_name} ps aux | grep {DB_PROCESS}"
+    output = await ops_test.juju(*find_pid.split())
+    pid = output[1].split()[1]
+    kill_cmd = f"run --unit {unit_name} -- kill -s {kill_code} {pid}"
```
paulomach, line 470: You can simplify this by using `pkill`:

```suggestion
    kill_cmd = f"run --unit {unit_name} -- pkill -s {kill_code} -f {DB_PROCESS}"
```

tests/integration/ha_tests/helpers.py
```diff
@@ -378,3 +453,62 @@ async def insert_focal_to_cluster(ops_test: OpsTest) -> None:
...
+    output = await ops_test.juju(*kill_cmd.split())
```
paulomach, line 471: This is more of a style thing, but you can use multiple assignment to retrieve the `return_code` like:
```suggestion
    return_code, _, _  = await ops_test.juju(*kill_cmd.split())
```
i.e. `return_code` is the same as `output[0]`

paulomach (approved):
great work. Add comments about code style only.
The tests (the main thing) are super

tests/integration/ha_tests/helpers.py
```diff
@@ -110,23 +124,57 @@ async def fetch_primary(replica_set_hosts: List[str], ops_test: OpsTest) -> str:
     return primary
 
 
+async def count_primaries(ops_test: OpsTest) -> str:
+    """Returns IP address of current replica set primary."""
```
marceloneppel, line 128:
```suggestion
    """Returns the number primary of the current replica set."""
```

marceloneppel (approved):
LGTM!

tests/integration/ha_tests/helpers.py
```diff
@@ -262,7 +310,34 @@ async def count_writes(ops_test: OpsTest) -> int:
     client = MongoClient(connection_string)
     db = client["new-db"]
     test_collection = db["test_collection"]
-    return sum(1 for _ in test_collection.find())
+    count = sum(1 for _ in test_collection.find())
```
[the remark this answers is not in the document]
MiaAltieri: oooooh did not know about this! You're awesome

PR #78, opened 2022-08-09
Stricter log checks

MiaAltieri:
## Problem:

Checks for `db_step_down` and `storage_reuse` use the log messages from MongoDB to verify operations. However they don't take into account the time of these log messages. i.e. the log message could have been the result of an earlier operation not the one we performed for the tests. This is especially a problem if a pre-provided cluster is given for testing, since it could already stepped down a member or re-used storage previously. 

## Solution:

Ensure operations such as `step_down` and `storage_reuse` are the direct result of an operation performed during testing. This is done via time stamping

### DIFF
--- a/tests/integration/ha_tests/helpers.py
+++ b/tests/integration/ha_tests/helpers.py
@@ -1,6 +1,7 @@
-# Copyright 2021 Canonical Ltd.
+# Copyright 2022 Canonical Ltd.
 # See LICENSE file for licensing details.
 
+import subprocess
 from pathlib import Path
 from typing import List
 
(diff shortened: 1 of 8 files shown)


tests/integration/ha_tests/helpers.py
```diff
@@ -504,7 +512,28 @@ async def db_step_down(ops_test: OpsTest, unit_ip: str):
...
+    """Converts a string time representation to an integer time representation."""
+    # parse time representation, provided in this format: 'YYYY-MM-DDTHH:MM:SS.MMM+00:00'
+    y_m_d, h_m_s_m = time_as_str.split("T")
+    year, month, day = y_m_d.split("-")
+    hour, minute, second = h_m_s_m.split(".")[0].split(":")
+    microsecond = h_m_s_m.split(".")[1].split("+")[0]
+
+    d = datetime.datetime(
+        int(year), int(month), int(day), int(hour), int(minute), int(second), int(microsecond)
+    )
```
zmraul, line 537: You can get your `datetime` object with `strptime`:
```python
from datetime import datetime

a = "2022-08-09T09:15:28.131+00:00"   # Mongo log

d = datetime.strptime(b, "%Y-%m-%dT%H:%M:%S.%f%z")
```
You can check the reference for format codes in [the docs](https://docs.python.org/3/library/datetime.html#strftime-and-strptime-format-codes), in case I made some mistake :)

MiaAltieri: soooo much nicer!! 🤩🤩🤩 thank you so much for sharing :)

shayancanonical (approved):
lgtm, but i agree with raul's comment about formatting time

zmraul (approved):

PR #79, opened 2022-08-09
HA-tests for full cluster crash and restart

MiaAltieri:
### Summary
Tests full cluster crash and full cluster restart.

### Specialised Knowledge

1. `MongoDBClient` can occasionally get out of sync with its vector clock leading to HMAC issues, therefore it is best practice to periodically recreate clients.
2. These tests test how the cluster behaves when members start up _after no members had their DB process running_, therefor it is important to verify that at some point all members are down. To ensure that at some point all members are down, we change the restart delay in the service file.

### Testing

Tests 
1. full cluster crash
3. full cluster restart

### Removal of step down verification
Up to this point step down has been verified by the log messages from mongodb. However these logs rotate quite quickly and thus at points they rotate quicker than we are able to verify. As of now a solution for this issue has not been determined and should be resolved in a future PR. Hence the commenting out of this verification step.

### DIFF
--- a/tests/integration/ha_tests/helpers.py
+++ b/tests/integration/ha_tests/helpers.py
@@ -500,22 +500,29 @@ async def mongod_ready(ops_test, unit_ip) -> bool:
     return True
 
 
-async def db_step_down(ops_test: OpsTest, unit_ip: str):
+async def db_step_down(ops_test: OpsTest, old_primary: str):
     app = await app_name(ops_test)
     password = await get_password(ops_test, app)
-    client = MongoClient(unit_uri(unit_ip, password, app), directConnection=True)
-    log = client.admin.command("getLog", "global")
-    client.close()
 
-    for item in log["log"]:
-        item = json.loads(item)
+    # loop through all units that aren't the old primary
+    for unit in ops_test.model.applications[app].units:
+        client = MongoClient(unit_uri(unit.public_adddress, password, app), directConnection=True)
+        log = client.admin.command("getLog", "global")
+        client.close()
 
-        if "msg" not in item:
+        if unit.public_address == old_primary:
             continue
 
-        # this message indicates that the previous primary performed a repl step down operation.
-        if item["msg"] == "Starting an election due to step up request":
-            return True
+        for item in log["log"]:
+            item = json.loads(item)
+
+            if "msg" not in item:
+                continue
+
+            # this message indicates that the previous primary performed a repl step down
+            # operation.
+            if item["msg"] == "Starting an election due to step up request":
+                return True
 
     return False
 
--- a/tests/integration/ha_tests/helpers.py
+++ b/tests/integration/ha_tests/helpers.py
@@ -26,9 +26,19 @@ METADATA = yaml.safe_load(Path("./metadata.yaml").read_text())
 PORT = 27017
 APP_NAME = METADATA["name"]
 DB_PROCESS = "/usr/bin/mongod"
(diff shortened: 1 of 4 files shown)


tests/integration/ha_tests/helpers.py
```diff
@@ -498,32 +508,118 @@ async def mongod_ready(ops_test, unit_ip) -> bool:
...
+    return True
+
+
+async def update_restart_delay(ops_test: OpsTest, unit, delay: int):
+    """Updates the restart delay in the DB service file.
+
+    When the DB service fails it will now wait for `dalay` number of seconds.
+    """
+    # load the service file from the unit and update it with the new delay
+    await unit.scp_from(source=MONGOD_SERVICE_DEFAULT_PATH, destination=TMP_SERVICE_PATH)
```
marceloneppel, line 574: If you choose to change the code to use `NamedTemporaryFile` (see my other comment) you can do something like below:
```python
with NamedTemporaryFile(mode="w") as f:
    await unit.scp_from(source=MONGOD_SERVICE_DEFAULT_PATH, destination=f.name)
    # Then run other steps like changing the file and uploading it again to the unit.
```

marceloneppel (commented):
LGTM! Just waiting for the tests to pass. I really liked the well documented helper functions and tests :smiley:

tests/integration/ha_tests/test_ha.py
```diff
@@ -459,3 +478,100 @@ async def test_restart_db_process(ops_test, continuous_writes):
...
+        ]
+    )
+
+    # This test serves to verify behavior when all replicas are down at the same time that when
+    # they come back online they operate as expected. This check verfies that we meet the criterea
+    # of all replicas being down at the same time.
+    assert await all_db_processes_down(ops_test), "Not all units down at the same time."
+
+    # sleep for twice the median election time and the restart delay
+    time.sleep(MEDIAN_REELECTION_TIME * 2 + RESTART_DELAY)
```
marcoppenheimer, line 505: 👍🏽

tests/integration/ha_tests/test_ha.py
```diff
@@ -459,3 +476,100 @@ async def test_restart_db_process(ops_test, continuous_writes):
...
+    more_writes = await count_writes(ops_test)
+    assert more_writes > writes, "writes not continuing to DB"
+
+    # verify presence of primary, replica set member configuration, and number of primaries
+    await verify_replica_set_configuration(ops_test)
+
+    # verify that no writes to the db were missed
+    total_expected_writes = await stop_continous_writes(ops_test)
+    actual_writes = await count_writes(ops_test)
+    assert total_expected_writes["number"] == actual_writes, "writes to the db were missed."
```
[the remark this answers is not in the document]
marcoppenheimer: Out of curiosity, does Mongo allow for duplicate writes as a KV store?
I suppose it might if you `insert()`, but I wonder if normally people `update()`?

marcoppenheimer (approved):
This is a really solid range of tests, very thorough and very readable.

tests/integration/ha_tests/test_ha.py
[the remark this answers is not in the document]
MiaAltieri: @marceloneppel hmm I don't think so but not 100% sure

@marcoppenheimer  good point, Mykola and I think that is a good fix for now :)

tests/integration/ha_tests/helpers.py
MiaAltieri: wasn't quite able to get this to work so I will keep this as is for now

marceloneppel (approved):

PR #82, opened 2022-08-17
Update logging for `mongod`

MiaAltieri:
## For the Reader
This was an old PR ticket that got closed due to complications around implementation and needing more research. After some discussion and research it was decided that this is the right implementation strategy.

## Context: 
[This JIRA ticket](https://warthogs.atlassian.net/browse/DPE-677?atlOrigin=eyJpIjoiNDM1Nzg1MTRjMTBiNDM4OThjNjJmYTcwMWNjMWUwOWUiLCJwIjoiaiJ9) refers to verifying that when a primary replica experiences a *planned removal* that it performs a "step down" operation before leaving the replica set.

## Problem 
The mongod logs show a replica leaving the replica set. But the logs rotate too quickly, meaning we can "miss" logging of this step down. Previously we added the flag `--logappend` to mongod's start args which effectively never rotates logs. The problem with this is that the log file becomes too large to parse during testing, leading ultimately to the tests hanging and failing. 

## Proposed Solution 
A couple of months ago I opened [this PR](https://github.com/canonical/mongodb-operator/pull/82) which used [this fixture](https://github.com/canonical/mongodb-operator/blob/bc9801687086df9a269df79040a01f258d18937a/tests/integration/ha_tests/test_ha.py#L70) to enable `logappend` for a test and then disable it at the end of the test. This solution solved two issues:
1. since logs were not rotating for the tests we were gauranteed to not miss the logging of the step down command
2. since `logappend` was only being used for the duration of a test , tests would not hang and fail due to a huge log file size

# =====================================
# Previous PR message
# =====================================
## Summary
Previously tests that relied on checking the log messages from the daemon with the `getLog` command were failing. Instead of checking this, we check the log file `mongodb.log` and set up the charm to write to this directory.

## Specialised Knowledge
1. It is not a good idea for check for certain log messages (ie for testing) via the `getLog` command. This is because this log rotates so you are not guaranteed that your sought after log message will be in those messages (ie log could have been rotated by the time you look)
2. It is important that we put our log file somewhere where it will persist after a unit has been removed. Therefore we put it in the charms storage `/data/db/`
4. The option `--logappend` allows the log file to grow rather than rotate, thus ensuring the presence of past log messages.


lib/charms/mongodb_libs/v0/machine_helpers.py
```diff
@@ -148,6 +148,9 @@ def generate_service_args(auth: bool, machine_ip: str, replset: str) -> str:
         # part of replicaset
         "--replSet",
         f"{replset}",
+        # logging should be sent to /data/db as to ensure persistence on machine removal.
+        # logappend ensures we have all logs instead of rotating through logs
+        "--logpath=/data/db/mongodb.log --logappend",
```
marceloneppel, line 153: I think it is important to do some rotation after some time, by using something like [db.adminCommand( { logRotate : "server" } )
](https://www.mongodb.com/docs/manual/tutorial/rotate-log-files/#rotate-the-log-file-1) at some point in the code and remobing old log files, to not fulfill the entire storage. This can be done in another PR.

marceloneppel (approved):
Nice approach, Mia! I left just one comment for another PR, relating to concerns about fulfilling the storage with the logs.

tests/integration/ha_tests/test_ha.py
```diff
@@ -474,95 +473,95 @@ async def test_restart_db_process(ops_test, continuous_writes):
...
-    assert more_writes > writes, "writes not continuing to DB"
-
-    # verify presence of primary, replica set member configuration, and number of primaries
-    await verify_replica_set_configuration(ops_test)
-
-    # verify that no writes to the db were missed
-    total_expected_writes = await stop_continous_writes(ops_test)
-    actual_writes = await count_writes(ops_test)
-    assert total_expected_writes["number"] == actual_writes, "writes to the db were missed."
+# async def test_full_cluster_crash(ops_test: OpsTest, continuous_writes, reset_restart_delay):
```
WRFitch, line 476: Why have these tests been commented out?

MiaAltieri: oops accident

marcoppenheimer (approved):

WRFitch (approved):

tests/integration/ha_tests/test_ha.py
```diff
@@ -64,6 +66,23 @@ async def reset_restart_delay(ops_test: OpsTest):
         await update_restart_delay(ops_test, unit, ORIGINAL_RESTART_DELAY)
 
 
+@pytest.fixture()
+async def change_logging(ops_test: OpsTest):
```
marceloneppel, line 70: Amazing idea for the tests!

marceloneppel (approved):
Looks awesome! This approach of enabling logging only during tests is amazing!

WRFitch (approved):

marceloneppel (approved):

marcoppenheimer (approved):

Issue #59, opened 2022-07-21
update relation data on db unit joining/leaving set


MiaAltieri:
Right now when a unit gets removed from/added to the replica set the related application is not made aware of this. This can lead to some negative consequences, i.e. an application not being able to connect to the replica set.

To resolve this:
1. either [the relation handler](https://github.com/canonical/mongodb-operator/blob/0896eb8e7901c0cb85ead45227b5b79b7fc75341/lib/charms/mongodb_libs/v0/mongodb_provider.py#L70) or [oversee_users](https://github.com/canonical/mongodb-operator/blob/0896eb8e7901c0cb85ead45227b5b79b7fc75341/lib/charms/mongodb_libs/v0/mongodb_provider.py#L122) should update the relation data fields `uri` and `hosts`. (similar to how [set_relation](https://github.com/canonical/mongodb-operator/blob/0896eb8e7901c0cb85ead45227b5b79b7fc75341/lib/charms/mongodb_libs/v0/mongodb_provider.py#L220) does perhaps in a function called `update_relation_hosts`)
2. Add necessary unit tests
3. Integration test this. Add a test to [test_charm_relations](https://github.com/canonical/mongodb-operator/blob/main/tests/integration/relation_tests/new_relations/test_charm_relations.py) This test should:
- scale the mongodb application by 2 units
- remove the original two units (the ones that were deployed at [the beginning](https://github.com/canonical/mongodb-operator/blob/0896eb8e7901c0cb85ead45227b5b79b7fc75341/tests/integration/relation_tests/new_relations/test_charm_relations.py#L40-L44) of the test)
- have the application perform CRUD operations, similar to [this test](https://github.com/canonical/mongodb-operator/blob/0896eb8e7901c0cb85ead45227b5b79b7fc75341/tests/integration/relation_tests/new_relations/test_charm_relations.py#L54)

averma-canonical:
Hi @MiaAltieri , can I close this?

averma-canonical:
Closing after ack from Mia.

PR #83, opened 2022-08-19
[UPDATED] App relation data refresh feature on mongo cluster scale up and down.

averma-canonical:
This pull request contains changes to update the application metadata (specifically endpoints and uris) as the mongodb cluster scales up and down. 

A detailed description of the feature requirements can be found [here](https://github.com/canonical/mongodb-operator/issues/59). 

An integration test covering both the scale up and down scenario has been added as well.

In addition to refactoring input, would request the reviewers to suggest more test case scenarios which make sense to add for this feature.

[UPDATE] 
Hi all, 
Kindly take a look at the feature and share your thoughts on error scenarios handling, especially during leader election cases.

### DIFF
--- a/lib/charms/mongodb_libs/v0/mongodb_provider.py
+++ b/lib/charms/mongodb_libs/v0/mongodb_provider.py
@@ -144,7 +144,7 @@ class MongoDBProvider(Object):
                 mongo.drop_user(username)
 
             for username in relation_users - database_users:
-                config = self._get_config(username)
+                config = self._get_config(username, None)
                 if config.database is None:
                     # We need to wait for moment when provider library
                     # set the database name into the relation.
@@ -154,7 +154,7 @@ class MongoDBProvider(Object):
                 self._set_relation(config)
 
             for username in relation_users.intersection(database_users):
-                config = self._get_config(username)
+                config = self._get_config(username, None)
                 logger.info("Update relation user: %s on %s", config.username, config.database)
                 mongo.update_user(config)
                 logger.info("Updating relation data according to diff")
@@ -205,14 +205,17 @@ class MongoDBProvider(Object):
         # Return the diff with all possible changes.
         return Diff(added, changed, deleted)
 
-    def _get_config(self, username: str) -> MongoDBConfiguration:
+    def _get_config(self, username: str, password: str) -> MongoDBConfiguration:
         """Construct config object for future user creation."""
         relation = self._get_relation_from_username(username)
+        if not password:
+            password = generate_password()
+
         return MongoDBConfiguration(
             replset=self.charm.app.name,
             database=self._get_database_from_relation(relation),
             username=username,
-            password=generate_password(),
+            password=password,
             hosts=self.charm.mongodb_config.hosts,
             roles=self._get_roles_from_relation(relation),
         )
--- a/src/charm.py
+++ b/src/charm.py
@@ -107,9 +107,6 @@ class MongodbOperatorCharm(ops.charm.CharmBase):
 
         self._update_hosts(event)
 
-        if "db_initialised" not in self.app_data:
-            return
-
         try:
             self.update_app_relation_data()
         except PyMongoError as e:
--- a/src/charm.py
+++ b/src/charm.py
@@ -58,6 +58,8 @@ MONGO_DATA_DIR = "/data/db"
 # We expect the MongoDB container to use the default ports
 MONGODB_PORT = 27017
 
+REL_NAME = "database"
+
 
 class MongodbOperatorCharm(ops.charm.CharmBase):
     """Charm the service."""
@@ -69,7 +71,7 @@ class MongodbOperatorCharm(ops.charm.CharmBase):
         self.framework.observe(self.on.install, self._on_install)
         self.framework.observe(self.on.config_changed, self._on_config_changed)
         self.framework.observe(self.on.start, self._on_start)
-        self.framework.observe(self.on[PEER].relation_joined, self._on_mongodb_relation_handler)
+        self.framework.observe(self.on[PEER].relation_joined, self._on_mongodb_relation_joined)
         self.framework.observe(self.on[PEER].relation_changed, self._on_mongodb_relation_handler)
 
         self.framework.observe(self.on.update_status, self._on_update_status)
@@ -105,6 +107,16 @@ class MongodbOperatorCharm(ops.charm.CharmBase):
 
         self._update_hosts(event)
 
+        if "db_initialised" not in self.app_data:
+            return
+
+        try:
+            self.update_app_relation_data()
+        except PyMongoError as e:
+            logger.error("Deferring on updating app relation data since: error: %r", e)
+            event.defer()
+            return
+
     def _update_hosts(self, event) -> None:
         """Update replica set hosts and remove any unremoved replicas from the config."""
         if "db_initialised" not in self.app_data:
@@ -132,15 +144,43 @@ class MongodbOperatorCharm(ops.charm.CharmBase):
         except PyMongoError as e:
             logger.error("Failed to remove %s from replica set, error=%r", self.unit.name, e)
 
+    def update_app_relation_data(self) -> None:
+        """Helper function to update application relation data."""
+        if "db_initialised" not in self.app_data:
+            return
+
+        database_users = set()
+
+        with MongoDBConnection(self.mongodb_config) as mongo:
+            database_users = mongo.get_users()
+
+        for relation in self.model.relations[REL_NAME]:
+            username = self.client_relations._get_username_from_relation_id(relation.id)
+            password = relation.data[self.app]["password"]
+            config = self.client_relations._get_config(username, password)
+            if username in database_users:
+                data = relation.data[self.app]
+                data["endpoints"] = ",".join(config.hosts)
+                data["uris"] = config.uri
+                relation.data[self.app].update(data)
+
     def _relation_departed(self, event: ops.charm.RelationDepartedEvent) -> None:
         """Remove peer from replica set if it wasn't able to remove itself.
 
         Args:
             event: The triggering relation departed event.
         """
-        # allow leader to update hosts if it isn't leaving
-        if self.unit.is_leader() and not event.departing_unit == self.unit:
-            self._update_hosts(event)
+        # allow leader to update relation data and hosts if it isn't leaving
+        if not self.unit.is_leader() or event.departing_unit == self.unit:
+            return
+        self._update_hosts(event)
+
+        try:
+            self.update_app_relation_data()
+        except PyMongoError as e:
+            logger.error("Deferring on updating app relation data since: error: %r", e)
+            event.defer()
+            return
 
     def process_unremoved_units(self, event) -> None:
         """Removes replica set members that are no longer running as a juju hosts."""
@@ -157,6 +197,23 @@ class MongodbOperatorCharm(ops.charm.CharmBase):
                 logger.error("Deferring process_unremoved_units: error=%r", e)
                 event.defer()
 
+    def _on_mongodb_relation_joined(self, event: ops.charm.RelationJoinedEvent) -> None:
+        """Add peer to replica set.
+
+        Args:
+            event: The triggering relation joined event.
+        """
+        if not self.unit.is_leader():
+            return
+
+        self._on_mongodb_relation_handler(event)
+        try:
+            self.update_app_relation_data()
+        except PyMongoError as e:
+            logger.error("Deferring on updating app relation data since: error: %r", e)
+            event.defer()
+            return
+
     def _on_mongodb_relation_handler(self, event: ops.charm.RelationEvent) -> None:
         """Adds the unit as a replica to the MongoDB replica set.
 
@@ -166,6 +223,7 @@ class MongodbOperatorCharm(ops.charm.CharmBase):
         # only leader should configure replica set and app-changed-events can trigger the relation
         # changed hook resulting in no JUJU_REMOTE_UNIT if this is the case we should return
         # further reconfiguration can be successful only if a replica set is initialised.
+
         if not (self.unit.is_leader() and event.unit) or "db_initialised" not in self.app_data:
             return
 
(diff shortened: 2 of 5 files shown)


MiaAltieri (changes requested):
Abhishek this looks AWESOME!! Really thank you so much for doing all of this so quickly! 🤩 

As for unit tests, perhaps some black box testing of `update_relation_data ` can be added

src/charm.py
```diff
@@ -157,6 +173,17 @@ def process_unremoved_units(self, event) -> None:
...
                 event.defer()
 
+    def _on_mongodb_relation_joined(self, event: ops.charm.RelationJoinedEvent) -> None:
+        """Add peer to replica set.
+
+        Args:
+            event: The triggering relation joined event.
+        """
+        if self.unit.is_leader():
+            self.update_relation_data()
```
MiaAltieri, line 183: it would be good to also do this for `relation-departed-events` (or you can do that in a separate follow up PR, if you do this in a separate PR it would be best to update your integration tests as to only test the adding units scenario :) )

MiaAltieri (changes requested):

tests/integration/relation_tests/new_relations/test_charm_relations.py
```diff
@@ -94,6 +94,40 @@ async def test_database_relation_with_charm_libraries(ops_test: OpsTest):
...
+    )
+    await ops_test.model.wait_for_idle(apps=[DATABASE_APP_NAME], status="active", timeout=1000)
+
+    endpoints_str = await get_application_relation_data(
+        ops_test, APPLICATION_APP_NAME, FIRST_DATABASE_RELATION_NAME, "endpoints"
+    )
+    endpoints = endpoints_str.split(",")
+    assert len(endpoints) == 2
+
+
```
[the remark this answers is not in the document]
averma-canonical: Hi Mia! 

Implemented all your suggestions.

Test 3 mentioned above i.e. CRUD is failing when leader is removed and gets re-elected. Need to discuss.

Deferring _on_relation_event since: error=ServerSelectionTimeoutError('No replica set members match selector "Primary()", Timeout: 1.0s, Topology Description: <TopologyDescription id: 630326ec6ee140b507b402f1, topology_type: ReplicaSetNoPrimary, servers: [<ServerDescription (\'10.114.175.35\', 27017) server_type: RSGhost, rtt: 0.0008979849517345428>]>')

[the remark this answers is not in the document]
averma-canonical: pymongo.errors.OperationFailure: Authentication failed., full error: {'ok': 0.0, 'errmsg': 'Authentication failed.', 'code': 18, 'codeName': 'AuthenticationFailed', '$clusterTime': {'clusterTime': Timestamp(1661153593, 1), 'signature': {'hash': b'\x00\x87r0q-\xead\x8dI\xc5\x00h_\xbb\xee\x0bT4q', 'keyId': 7134599397790187525}}, 'operationTime': Timestamp(1661153593, 1)}

src/charm.py
```diff
@@ -104,6 +106,7 @@ def _on_leader_elected(self, event) -> None:
             self._generate_passwords()
 
         self._update_hosts(event)
+        self.update_app_relation_data(check_mongo_users=False)
```
MiaAltieri, line 109: maybe something like this will help

`MEDIAN_REELECTION_TIME=12`

```suggestion
        time.sleep(MEDIAN_REELECTION_TIME*2)
        self.update_app_relation_data(check_mongo_users=True)
```

tests/integration/relation_tests/new_relations/test_charm_relations.py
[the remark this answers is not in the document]
averma-canonical: Thanks! Implemented all these suggestions.

src/charm.py
averma-canonical: Thanks! Done. This works, just that the notion of sleep makes the charm integration tests a little flaky for CI gates.

src/charm.py
```diff
@@ -105,6 +110,11 @@ def _on_leader_elected(self, event) -> None:
 
         self._update_hosts(event)
 
+        if "db_initialised" not in self.app_data:
+            return
+        time.sleep(MEDIAN_REELECTION_TIME * 2)
+        self.update_app_relation_data()
```
MiaAltieri, line 116: I don't know what I was thinking when I suggested this, `event.defer()` would be much better here
```suggestion
try:
       self.update_app_relation_data(check_mongo_users=True)
except OperationFailure as e:
       # when a unit departs the replica set it is possible that unit was a leader, thus resulting in mongod electing a new leader. If mongod hasn't elected a 
       # new leader yet an Authentication error will occur and we should retry updating the application relation data.
       if e.code == 18:
             logger.debug("Waiting for replicaset to elect a new primary before updating relation data")
             event.defer()
       else:
              raise
```

MiaAltieri (changes requested):

lib/charms/mongodb_libs/v0/mongodb_provider.py
```diff
@@ -205,14 +205,17 @@ def _diff(self, event: RelationChangedEvent) -> Diff:
         # Return the diff with all possible changes.
         return Diff(added, changed, deleted)
 
-    def _get_config(self, username: str) -> MongoDBConfiguration:
+    def _get_config(self, username: str, password: str) -> MongoDBConfiguration:
         """Construct config object for future user creation."""
         relation = self._get_relation_from_username(username)
+        if password is None:
```
MiaAltieri, line 211: nit:
```suggestion
        if not password:
```

src/charm.py
```diff
@@ -132,15 +153,36 @@ def _on_mongodb_storage_detaching(self, event: ops.charm.StorageDetachingEvent)
...
         """Remove peer from replica set if it wasn't able to remove itself.
 
         Args:
             event: The triggering relation departed event.
         """
-        # allow leader to update hosts if it isn't leaving
+        # allow leader to update relation data and hosts if it isn't leaving
         if self.unit.is_leader() and not event.departing_unit == self.unit:
             self._update_hosts(event)
+            self.update_app_relation_data()
```
MiaAltieri, line 185:
```suggestion
        # allow leader to update relation data and hosts if it isn't leaving
        if not self.unit.is_leader() or event.departing_unit == self.unit:
            return 
       
       self._update_hosts(event)
        try:
            self.update_app_relation_data()
        except PyMongoError as e:
            logger.error("Deferring on updating app relation data since: error: %r", e)
            event.defer()
            return
```

src/charm.py
```diff
@@ -157,6 +199,17 @@ def process_unremoved_units(self, event) -> None:
...
+        """Add peer to replica set.
+
+        Args:
+            event: The triggering relation joined event.
+        """
+        self._on_mongodb_relation_handler(event)
+
+        if self.unit.is_leader():
+            self.update_app_relation_data()
+
```
MiaAltieri, line 212:
```suggestion
        # allow leader to update relation data and hosts if it isn't leaving
        if not self.unit.is_leader():
            return 
       
       self._on_mongodb_relation_handler(event)
        try:
            self.update_app_relation_data()
        except PyMongoError as e:
            logger.error("Deferring on updating app relation data since: error: %r", e)
            event.defer()
            return

```

src/charm.py
```diff
@@ -105,6 +108,24 @@ def _on_leader_elected(self, event) -> None:
...
+            # thus resulting in mongod electing a new leader. If mongod hasn't elected a
+            # new leader yet an Authentication error will occur and we should retry updating
+            # the application relation data.
+            if e.code == 18:
+                logger.debug(
+                    "Waiting for replicaset to elect a new primary before updating relation data"
+                )
+                event.defer()
+            else:
+                raise
```
MiaAltieri, line 127: Now that I think about it, it would be best to catch all PyMongoErrors. Sorry about that Abhishek.
```suggestion
        except PyMongoError as e:
            logger.error("Deferring on updating app relation data since: error: %r", e)
            event.defer()
            return
```

tests/integration/relation_tests/new_relations/test_charm_relations.py
```diff
@@ -94,6 +96,104 @@ async def test_database_relation_with_charm_libraries(ops_test: OpsTest):
...
+        connection_string,
+        directConnection=False,
+        connect=False,
+        serverSelectionTimeoutMS=1000,
+        connectTimeoutMS=2000,
+    )
+
+    db = client[database]
+    test_collection = db["test_app_collection"]
+    ubuntu = {"release_name": "Focal Fossa", "version": 20.04, "LTS": True}
```
MiaAltieri, line 177: seems like the data below is already written from the previous test. To avoid any false positives from the previous test, it would be best to clear the DB
```suggestion
    # clear collection  writes from previous test
    test_collection = db["test_app_collection"]
    test_collection.drop()
    test_collection = db["test_app_collection"]
    
    ubuntu = {"release_name": "Focal Fossa", "version": 20.04, "LTS": True}
```

MiaAltieri (changes requested):
Looks awesome! Minor suggestions regarding error checking

tests/integration/relation_tests/new_relations/test_charm_relations.py
averma-canonical: Done

MiaAltieri (approved):

src/charm.py
```diff
@@ -105,6 +107,16 @@ def _on_leader_elected(self, event) -> None:
 
         self._update_hosts(event)
 
+        if "db_initialised" not in self.app_data:
```
paulomach, line 110: since you are testing this on `update_app_relation_data` and not testing on `_relation_departed`, you could remove it from here for code consistency

src/charm.py
```diff
@@ -132,15 +144,43 @@ def _on_mongodb_storage_detaching(self, event: ops.charm.StorageDetachingEvent)
...
-        if self.unit.is_leader() and not event.departing_unit == self.unit:
-            self._update_hosts(event)
+        # allow leader to update relation data and hosts if it isn't leaving
+        if not self.unit.is_leader() or event.departing_unit == self.unit:
+            return
+        self._update_hosts(event)
+
+        try:
+            self.update_app_relation_data()
+        except PyMongoError as e:
```
paulomach, line 180: I wonder what kind of errors will be catched here, and I defer should be the treatment for all, i.e. couldn't be the case of this is masking a blocking condition?

src/charm.py
```diff
@@ -157,6 +197,23 @@ def process_unremoved_units(self, event) -> None:
...
+        """
+        if not self.unit.is_leader():
+            return
+
+        self._on_mongodb_relation_handler(event)
+        try:
+            self.update_app_relation_data()
+        except PyMongoError as e:
+            logger.error("Deferring on updating app relation data since: error: %r", e)
+            event.defer()
```
paulomach, line 214: same as l180

paulomach (approved):
Left some broad questions, but otherwise looks solid.

src/charm.py
averma-canonical: makes sense. thanks Paulo!

src/charm.py
averma-canonical: Let me try to find out the exact error so we can be specific about this.

src/charm.py
MiaAltieri: Actually I think it should be kept here, when the charm starts up for the first time it executes the `on_leader_elected` hook before the start hook and so the DB won't have been started i.e.  "db_initialised" won't be in self.app_data and doing something like `update_app_relation_data` could cause problems since the functions in `update_app_relation_data` use communication with `mongod`

src/charm.py
```diff
@@ -132,15 +144,43 @@ def _on_mongodb_storage_detaching(self, event: ops.charm.StorageDetachingEvent)
...
             event: The triggering relation departed event.
         """
-        # allow leader to update hosts if it isn't leaving
-        if self.unit.is_leader() and not event.departing_unit == self.unit:
-            self._update_hosts(event)
+        # allow leader to update relation data and hosts if it isn't leaving
+        if not self.unit.is_leader() or event.departing_unit == self.unit:
+            return
+        self._update_hosts(event)
+
```
MiaAltieri, line 177:
```suggestion
if "db_initialised" not in self.app_data:
     return
```

src/charm.py
MiaAltieri: Hmm I think this might be a longer conversation, up to this point the charm has been catching all `PyMongoError`'s and deferring when possible. I think the idea is that if a `PyMongoError` occurs two things can happen:

1. We can defer, 
     - Pros: maybe `mongod` needed to do something (ie reconfigure, re-elect a primary) and the defer gives `mongod` time to do whatever operation is required, and when the hook gets fired again it will succeed on the second time.
     - Cons: maybe something more troubling is going on and this defers forever, then the user will have to manually debug the situation. (in this case we have the `logger.debug` message which gives context to the error)
     
2. Don't defer and let it error
     - Pros: we don't infinitely defer
     - Cons: we don't give `mongod` a chance to fix itself and other hooks might be able to fire further breaking the program and the user will still have to manually debug the situation. (it might be easier to manually debug if they see the raised error here)

I think it is best to leave this as it is since option 1 gives us some possibility of resolving the issue. But I am still new to system design and user design of charms. I think it might be useful here to get our local MongoDB/Charm expert @delgod 's opinion on what we should do here.

src/charm.py
MiaAltieri: see above

src/charm.py
delgod: Errors that we catching here are 1) or connection errors, 2) or a no-primary error, 3) or something similar.
MongoDB eventually recovers itself from such errors, so we just need to retry later.
I believe it is best to defer in such scenarios without blocking the whole charm.

src/charm.py
delgod: @MiaAltieri is absolutely right

averma-canonical: Just discussed with Mia, the functionality mentioned is already there. Will simply remove redundant check in leader election handler.

averma-canonical: ack

Issue #84, opened 2022-08-23
move functionality for secrets to `get_secret` and `set_secret`


MiaAltieri:
see [here](https://github.com/delgod/mongodb-k8s-operator/pull/26)

MiaAltieri:
Resolved in #88

PR #88, opened 2022-09-09
create secret implementation

MiaAltieri:
This PR aims to move secret implementation logic inside the separate functions. This way when juju secrets becomes available it will be much easier to update the charm to use juju secrets. (Ie without the logic of charm itself.)

Note: we have two types of secrets. Some secrets are shared between units, and some should be accessible and unique only for a particular unit.

This PR will be useful for TLS secrets as well

This addresses issue #84


marceloneppel (approved):
Looks great!

zmraul (approved):

PR #90, opened 2022-09-19
Tls lib + readme

MiaAltieri:
Adding TLS library code + README.md udates. These files have already undergone a thorough review process for the k8s charm and **do not need review.**

Requesting to merge into TLS-feature-branch rather than into Main, so that all of TLS (libs, impl, unit-test, int-tests) can be merged at once and pushed to charmhub.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,6 +68,23 @@ We have also added support for the database legacy relation from the [original v
 juju relate mongodb graylog
 ```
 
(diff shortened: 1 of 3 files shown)


averma-canonical (approved):

welpaolo (approved):
lgtm

PR #94, opened 2022-09-21
TLS Integration tests test

MiaAltieri:
### For the reader 
Requesting to merge into `TLS-feature-branch` rather than into `main`, so that all of TLS (libs, implementation, unit-test, int-tests) can be merged at once and pushed to charmhub. TLS integration tests coming in next PR

### Release Notes
Tests for enabling and disabling of TLS

### Future PRs
1. unit tests
2. integration tests for `set-tls-key`

### DIFF
--- a/tests/integration/tls_tests/test_tls.py
+++ b/tests/integration/tls_tests/test_tls.py
@@ -1,0 +1,49 @@
+#!/usr/bin/env python3
+# Copyright 2022 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+import pytest
+from pytest_operator.plugin import OpsTest
+
+from tests.integration.tls_tests.helpers import check_tls
+
+TLS_CERTIFICATES_APP_NAME = "tls-certificates-operator"
+DATABASE_APP_NAME = "mongodb"
+
+
+# @pytest.mark.abort_on_fail
+# async def test_build_and_deploy(ops_test: OpsTest) -> None:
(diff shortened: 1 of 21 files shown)


tests/integration/tls_tests/test_tls.py
```diff
@@ -0,0 +1,49 @@
...
+    # Relate it to the PostgreSQL to enable TLS.
+    await ops_test.model.relate(DATABASE_APP_NAME, TLS_CERTIFICATES_APP_NAME)
+    await ops_test.model.wait_for_idle(status="active", timeout=1000)
+
+    # Wait for all units enabling TLS.
+    for unit in ops_test.model.applications[DATABASE_APP_NAME].units:
+        assert await check_tls(ops_test, unit, enabled=True)
+
+
+async def test_disable_tls(ops_test: OpsTest) -> None:
```
WRFitch, line 39: Would it be worth sending & receiving some data, and verifying that TLS is applied?

WRFitch (approved):

tests/integration/tls_tests/test_tls.py
MiaAltieri: `check_tls` sends a command to mongod and then receives the results of the command to verifies that TLS is applied.

I could send data, but I think this is sufficient

marceloneppel (approved):
LGTM!

PR #95, opened 2022-09-21
Tls unit tests

MiaAltieri:
### For the reader 
Requesting to merge into `TLS-feature-branch` rather than into `main`, so that all of TLS (libs, implementation, unit-test, int-tests) can be merged at once and pushed to charmhub. Final TLS integration tests coming in next PR

### Release Notes
- [lib/charms/mongodb_libs/v0/mongodb_tls.py](https://github.com/canonical/mongodb-operator/compare/TLS-feat-branch...TLS-unit-tests?expand=1#diff-3a76d5563d1547ccbbb6f5038e8aa23bd008fddcb4bb555a2e738a35214cb717) - minor corrections
- [tests/unit/test_tls_lib.py](https://github.com/canonical/mongodb-operator/compare/TLS-feat-branch...TLS-unit-tests?expand=1#diff-2fe86ffbee888702d647689e6b0a9a3a688648e0eebc889736a49f51eed4ee37) - added unit tests


tests/unit/test_tls_lib.py
```diff
@@ -0,0 +1,311 @@
...
+        self.harness.set_leader(True)
+        rel_id = self.relate_to_tls_certificates_operator()
+        app_rsa_key = self.harness.charm.app_peer_data["key"]
+        app_csr = self.harness.charm.app_peer_data["csr"]
+
+        self.harness.set_leader(False)
+        self.harness.remove_relation(rel_id)
+        self.assertEqual(self.harness.charm.unit_peer_data.get("ca", None), None)
+        self.assertEqual(self.harness.charm.unit_peer_data.get("cert", None), None)
+        self.assertEqual(self.harness.charm.unit_peer_data.get("chain", None), None)
```
WRFitch, line 107: You could use `self.assertIsNone` here, and you can also use self.charm instead of self.harness.charm.

tests/unit/test_tls_lib.py
```diff
@@ -0,0 +1,311 @@
...
+
+        self.harness.remove_relation(rel_id)
+
+        # internal certificates and external certificates should be removed
+        self.assertEqual(self.harness.charm.unit_peer_data.get("ca", None), None)
+        self.assertEqual(self.harness.charm.unit_peer_data.get("cert", None), None)
+        self.assertEqual(self.harness.charm.unit_peer_data.get("chain", None), None)
+        self.assertEqual(self.harness.charm.app_peer_data.get("ca", None), None)
+        self.assertEqual(self.harness.charm.app_peer_data.get("cert", None), None)
+        self.assertEqual(self.harness.charm.app_peer_data.get("chain", None), None)
```
WRFitch, line 133: see above re: self.charm, and self.assertIsNone

WRFitch (approved):

tests/unit/test_tls_lib.py
MiaAltieri: TWIL thanks @WRFitch

tests/unit/test_tls_lib.py
```diff
@@ -0,0 +1,311 @@
...
+        )
+        self.verify_external_rsa_csr()
+
+    @patch_network_get(private_address="1.1.1.1")
+    def test_tls_relation_joined_leader(self):
+        """Test that leader units set both external and internal certificates."""
+        self.harness.set_leader(True)
+        self.relate_to_tls_certificates_operator()
+        self.verify_internal_rsa_csr()
+        self.verify_external_rsa_csr()
```
marcoppenheimer, line 91: Opinion, I would split these tests into separate tests. 
Mostly for the reason that if a test fails, I'd rather know what was wrong from the test name and fix that, than have to debug it. 
It would add loads of lines of bloat for setting up the test scenario though, so (not necessary though) maybe you could add these tests into their own test class, each as methods, with a shared `setUp`?

tests/unit/test_tls_lib.py
```diff
@@ -0,0 +1,311 @@
...
+        app_csr = self.harness.charm.app_peer_data.get("csr", None)
+        if specific_rsa:
+            self.assertEqual(app_rsa_key, expected_rsa)
+        else:
+            self.assertEqual(app_rsa_key.split("\n")[0], "-----BEGIN RSA PRIVATE KEY-----")
+
+        if specific_csr:
+            self.assertEqual(app_csr, expected_csr)
+        else:
+            self.assertEqual(app_csr.split("\n")[0], "-----BEGIN CERTIFICATE REQUEST-----")
```
marcoppenheimer, line 311: Huge nit, these could be the one method right? Something like:

```python
def verify_rsa_csr(
    self,
    specific_rsa=False,
    expected_rsa=None,
    specific_csr=False,
    expected_csr=None,
    object: Union[Unit, Application]
):
    rsa_key = self.harness.charm.model.get_relation(PEER).data[object].get("key", None)
    ...
```

tests/unit/test_tls_lib.py
```diff
@@ -0,0 +1,311 @@
...
+        action_event.params = {}
+
+        # generated rsa key test - leader
+        self.harness.charm.tls._on_set_tls_private_key(action_event)
+        self.verify_internal_rsa_csr()
+        self.verify_external_rsa_csr()
+
+        # provided rsa key test - leader
+        set_app_rsa_key = self.harness.charm.app_peer_data["key"]
+        action_event.params = {"internal-key": set_app_rsa_key}
```
marcoppenheimer, line 44: Is this expected behavior? Where we can `set_tls_private_key` with the same key as it already had, and that works?

I haven't read the other code though, is `key` something different?

marcoppenheimer: I think I'm just not clear what's being tested here, as we're getting `key` from relation data, updating key, then checking that the `internal_rsa_csr` is there, which it is because it was there already? I'm sure I'm missing something though.

marcoppenheimer (commented):
I left some nits and questions, but overall a REALLY clean set of tests, exhaustive and very tidy.

I particularly like the `self.substrate` (which I assume is for having the same code between `k8s` and `vm` charms?) and the merging into a feature branch to smooth PR review on big changes, definitely going to nick both of them!

tests/unit/test_tls_lib.py
```diff
@@ -0,0 +1,311 @@
...
+        self.addCleanup(self.harness.cleanup)
+
+    @patch_network_get(private_address="1.1.1.1")
+    def test_on_set_tls_private_key(self):
+        """Tests setting of TLS private key in external/internal certificate scenarios.
+
+        Note: this implicitly tests: _request_certificate & _parse_tls_file
+        """
+        # Tests for leader unit (ie internal certificates and external certificates)
+        self.harness.set_leader(True)
```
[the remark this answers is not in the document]
MiaAltieri: good eye thank you :)

tests/unit/test_tls_lib.py
MiaAltieri: I totally agree, great suggestion, I found a way around the additional bloat :)

tests/unit/test_tls_lib.py
MiaAltieri: it could be, but I like it better like this. If this is a make-or-break nit I am happy to refactor otherwise I will leave as is : )

tests/unit/test_tls_lib.py
MiaAltieri: Hi Marc great questions here, I only used `set_app_rsa_key = self.harness.charm.app_peer_data["key"]` as it was in the correct format that is needed by the `set-key` action so this test was just meant to test that setting the key to a specified value works, but I get what you're saying about it being the same I can change it. `set-key` for leaders will reset both the `internal` and `external` keys so if you don't pass in a value it gets generated randomly.

tests/unit/test_tls_lib.py
marcoppenheimer: Nah, it's a big nit.

lib/charms/mongodb_libs/v0/mongodb_tls.py
```diff
@@ -74,6 +74,7 @@ def _on_set_tls_private_key(self, event: ActionEvent) -> None:
             self._request_certificate("app", event.params.get("internal-key", None))
             logger.debug("Successfully set TLS private key.")
         except ValueError as e:
+            print("FAAILING WITH", e)
```
marcoppenheimer, line 77: Remove?

tests/unit/data/key.pem
```diff
@@ -0,0 +1,39 @@
+-----BEGIN RSA PRIVATE KEY-----
```
marcoppenheimer, line 1: Nice.

marcoppenheimer (approved):

marceloneppel (approved):
Awesome coverage!

tests/unit/test_tls_lib.py
```diff
@@ -0,0 +1,322 @@
...
+
+RELATION_NAME = "certificates"
+
+
+class TestMongoTLS(unittest.TestCase):
+    @patch_network_get(private_address="1.1.1.1")
+    def setUp(self):
+        self.harness = Harness(MongodbOperatorCharm)
+        self.harness.begin()
+        self.harness.add_relation("database-peers", "mongodb-peers")
```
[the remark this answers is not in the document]
MiaAltieri: good catch this is supposed to be the database relation so it shoould be 
`self.harness.add_relation("database-peers", "mongodb-peers")`

PR #96, opened 2022-09-22
added test for setting private TLS key and updated README for setting…

MiaAltieri:
### For the reader 
Requesting to merge into `TLS-feature-branch` rather than into `main`, so that all of TLS (libs, implementation, unit-test, int-tests) can be merged at once and pushed to charmhub. Final TLS integration tests coming in next PR

### Release Notes
1. Updated the README to include instructions on how to rotate the TLS private keys
2. Added tests for rotating TLS private keys, test ensures new certs are created, that the service is restarted, and units still have TLS enabled.


tests/integration/tls_tests/helpers.py
```diff
@@ -12,6 +13,10 @@
 PORT = 27017
 
 
+class ProcessError(Exception):
```
WRFitch, line 16: You could import this from `multiprocessing`

tests/integration/tls_tests/test_tls.py
```diff
@@ -36,6 +46,69 @@ async def test_enable_tls(ops_test: OpsTest) -> None:
...
+    )
+
+    # set external key
+    leader_unit = await find_unit(ops_test, leader=True)
+    await leader_unit.run_action(
+        action_name="set-tls-private-key", params="=".join(["internal-key", key_contents])
+    )
+
+    # sleep for long enough to process new certificate request and update certs
+    sleep(30)
```
WRFitch, line 85: Is there something you can wait for here? Maybe action completion, or a change in status using `wait_for_idle`?

WRFitch (approved):

tests/integration/tls_tests/test_tls.py
```diff
@@ -36,6 +46,69 @@ async def test_enable_tls(ops_test: OpsTest) -> None:
...
+            ops_test, unit.name, DB_SERVICE
+        )
+
+    with open(PRIVATE_KEY_PATH) as f:
+        key_contents = f.readlines()
+        key_contents = "".join(key_contents)
+
+    # set internal key
+    leader_unit = await find_unit(ops_test, leader=True)
+    leader_unit = await find_unit(ops_test, leader=True)
```
shayancanonical, line 73: duplicates: line 72 and 73

tests/integration/tls_tests/test_tls.py
```diff
@@ -36,6 +46,69 @@ async def test_enable_tls(ops_test: OpsTest) -> None:
...
+
+        # Once the certificate requests are processed and updated the mongod.service should be
+        # restarted
+        assert (
+            new_mongod_service_time > original_tls_times[unit.name]["mongod_service"]
+        ), f"mongod service for {unit.name} was not restarted."
+
+    # Verify that TLS is functioning on all units.
+    for unit in ops_test.model.applications[DATABASE_APP_NAME].units:
+        assert await check_tls(ops_test, unit, enabled=True)
```
shayancanonical, line 109: may be to confirm, it'd be nice to check that tls is disabled before setting the certificates?

MiaAltieri: TLS is already enabled at this point (in the previous test) so it shouldn't be disabled before setting the cert.

This check verifies that resetting it didn't break anything, i.e. TLS is still enabled. By verifying the service restarted (above) and the new files were created (also above) this verifies that it we infact changed TLS cert and restarted mongod with the new TLS cert and that this new cert works

tests/integration/tls_tests/test_tls.py
MiaAltieri: Great question @WRFitch :) Unfortunately I cannot wait for the action to complete since setting the keys doesn't immediately result in the charm enabling TLS. Changing the keys generates a certificate signing requests (CSR), which isn't processed until the CA processes this request. Then the TLS charm will issue a certificate-available-event which after that event is issued we receive the event on the DB charm and restart TLS with the correct certs. (For this reason I cannot wait for idle either 😢 )

I could use retry over a period of 30secs but this would be 5-7 more lines of code and essentially do the same thing so I will keep this as sleep if that is okay.

tests/integration/tls_tests/test_tls.py
MiaAltieri: 🤦‍♀️ thanks

tests/integration/tls_tests/test_tls.py
WRFitch: Great, thanks for the explanation :)

shayancanonical (approved):

PR #98, opened 2022-09-27
fix to regex + update tests +fix README

MiaAltieri:
### Problem

TLS set-action worked for the integration tests but not via the commandline

### Solution

TLS regex had to be fixed and integration tests were corrected to follow format of commandline action

(also README was updated as it was incorrectly formatted)


tests/integration/tls_tests/helpers.py
```diff
@@ -53,6 +53,7 @@ async def check_tls(ops_test: OpsTest, unit: ops.model.Unit, enabled: bool) -> b
                 return_code, _, _ = await ops_test.juju(*check_tls_cmd.split())
                 tls_enabled = return_code == 0
                 if enabled != tls_enabled:
+                    print(f"TLS is{' not' if not tls_enabled else ''} enabled on {unit.name}")
```
MiaAltieri, line 56: remove

PR #99, opened 2022-09-28
Regex fix

MiaAltieri:
### Problem

TLS set-action worked for the integration tests but not via the commandline

### Solution

TLS regex had to be fixed, integration & unit tests were corrected to follow format of commandline action. README was also updated accordingly


tests/integration/tls_tests/test_tls.py
```diff
@@ -67,18 +65,14 @@ async def test_rotate_tls_key(ops_test: OpsTest) -> None:
...
     # set external and internal key using auto-generated key for each unit
     for unit in ops_test.model.applications[DATABASE_APP_NAME].units:
         action = await unit.run_action(action_name="set-tls-private-key")
         action = await action.wait()
         assert action.status == "completed", "setting external and internal key failed."
 
-    # sleep for long enough to process new certificate request and update certs
-    sleep(30)
+    # wait for certificate to be available and processed.
+    await ops_test.model.wait_for_idle()
```
WRFitch, line 75: Should this be pointing at a specific application?

WRFitch (approved):

tests/integration/tls_tests/test_tls.py
MiaAltieri: this waits for all applications in the model to be idle , so its the same as `applications=[APP_1, APP_1]`

PR #102, opened 2022-10-04
Resolving Bug for failure to reset TLS keys 

MiaAltieri:
### Problem

Setting the TLS private internal key was flakey. It would fail with _both_ auto generated private internal keys and provided internal keys

### Solution

The bug was caused by [this line](https://github.com/canonical/mongodb-operator/blob/2824c0eab45e0dd9ab2585f1c3d19aa947d14a9a/lib/charms/mongodb_libs/v0/mongodb_tls.py#L163-L164), no matter the status of a unit leader/non-leader, internal certificates should be updated (ie saved to the machine and mongod should be restarted with the new private key.)

To understand why this leads to flakey tests consider these two scenarios:
1. Working case: the app cert data gets updated with the internal cert before the (non-leader) unit receives an `_on_certificate_available` event , then it restarts with both internal and external certs
2. Broken case: the app cert data does NOT get updated with the internal cert before the (non-leader) unit receives an `_on_certificate_available` event , then it restarts with only the external certs. When it receives a second `_on_certificate_available` with the app cert it skips because of the return and doesn't have the correct internal cert.

### Release Notes

1. Removed bug causing line
2. Integration tests updated
3. Unit tests updated
4. README had poor formatting so it was updated as well


tests/integration/tls_tests/test_tls.py
```diff
@@ -14,7 +12,7 @@
 
 TLS_CERTIFICATES_APP_NAME = "tls-certificates-operator"
 DATABASE_APP_NAME = "mongodb"
-PRIVATE_KEY_PATH = "tests/integration/tls_tests/data/key.txt"
+PRIVATE_KEY_PATH = "tests/integration/tls_tests/data"
```
WRFitch, line 15: Maybe this variable could be renamed to something more descriptive?

tests/integration/tls_tests/test_tls.py
```diff
@@ -122,22 +118,33 @@ async def test_set_tls_key(ops_test: OpsTest) -> None:
...
-        key_contents = f.readlines()
-        key_contents = "".join(key_contents)
+    with open(f"{PRIVATE_KEY_PATH}/internal-key.pem") as f:
+        internal_key_contents = f.readlines()
+        internal_key_contents = "".join(internal_key_contents)
 
     # set external and internal key for each unit
-    for unit in ops_test.model.applications[DATABASE_APP_NAME].units:
-        key_settings = [f"internal-key={key_contents}", f"external-key={key_contents}"]
+    for unid_id in range(len(ops_test.model.applications[DATABASE_APP_NAME].units)):
```
WRFitch, line 126:
```suggestion
    for unit_id in range(len(ops_test.model.applications[DATABASE_APP_NAME].units)):
```

WRFitch (approved):

tests/integration/tls_tests/test_tls.py
```diff
@@ -67,18 +68,15 @@ async def test_rotate_tls_key(ops_test: OpsTest) -> None:
...
     for unit in ops_test.model.applications[DATABASE_APP_NAME].units:
         action = await unit.run_action(action_name="set-tls-private-key")
         action = await action.wait()
         assert action.status == "completed", "setting external and internal key failed."
 
-    # sleep for long enough to process new certificate request and update certs
-    sleep(30)
+    # wait for certificate to be available and processed. Can get receive two certificate
+    # available events and restart twice so we do not wait for idle here
+    time.sleep(60)
```
welpaolo, line 79: Is ok to use time.sleep() in the tests? In some other components I was using the tenacity library to avoid it.

welpaolo (approved):

tests/integration/tls_tests/test_tls.py
MiaAltieri: Great question, time.sleep() in tests is okay when it is justfied, but typically you will want to use Tenacity :)

PR #105, opened 2022-10-06
Network healing test

MiaAltieri:
### For the Reader
This is the final PR for network self healing it covers the necessary integration tests and updates the README

It is important to note that as of now Juju incorrectly reports IP addresses after they change, and for the time being a fix using the function `get_unit_ip` is the work around. Until [this bug](https://github.com/juju/python-libjuju/issues/738) is resolved this will need to be used.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,12 @@ juju run-action mongodb/<unit_number> get-primary --wait
 
 Similarly, the primary replica is displayed as a status message in `juju status`, however one should note that this hook gets called on regular time intervals and the primary may be outdated if the status hook has not been called recently. 
 
+Further we highly suggest configuring the status hook to run frequently. In addition to reporting the primary, secondaries, and other statuses, the status hook performs self healing in the case of a network cut. To change the frequence of the update status hook do:
+```shell
+juju model-config update-status-hook-interval=<time(s/m/h)>
+```
+You can read more about status hooks [here](https://juju.is/docs/sdk/update-status-event).
+
 ### Replication
 #### Adding Replicas
 To add more replicas one can use the `juju add-unit` functionality i.e.
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ Further we highly suggest configuring the status hook to run frequently. In addi
 ```shell
 juju model-config update-status-hook-interval=<time(s/m/h)>
 ```
-You can read more about status hooks [here](https://juju.is/docs/sdk/update-status-event).
+Note that this hook executes a read query to MongoDB. On a production level server this should be configured to occur at a frequency that doesn't overload the server with read requests. Similarly the hook should not be configured at too quick of a frequency as this can delay other hooks from running. You can read more about status hooks [here](https://juju.is/docs/sdk/update-status-event).
 
 ### Replication
 #### Adding Replicas
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ juju run-action mongodb/<unit_number> get-primary --wait
 
 Similarly, the primary replica is displayed as a status message in `juju status`, however one should note that this hook gets called on regular time intervals and the primary may be outdated if the status hook has not been called recently. 
 
-Further we highly suggest configuring the status hook to run frequently. In addition to reporting the primary, secondaries, and other statuses, the status hook performs self healing in the case of a network cut. To change the frequence of the update status hook do:
+Further we highly suggest configuring the status hook to run frequently. In addition to reporting the primary, secondaries, and other statuses, the status hook performs self healing in the case of a network cut. To change the frequency of the update status hook do:
 ```shell
 juju model-config update-status-hook-interval=<time(s/m/h)>
 ```
--- a/src/charm.py
+++ b/src/charm.py
@@ -342,11 +342,7 @@ class MongodbOperatorCharm(ops.charm.CharmBase):
         if not "db_initialised" in self.app_peer_data:
             return
 
-        # first check the health of the unit before updating its replica set status
-        self._health_check()
-
         # update the units status based on it's replica set status.
-        # TODO will this break if one of the units is unreachable -- NO
         with MongoDBConnection(self.mongodb_config) as mongo:
             replset_status = mongo.get_replset_status()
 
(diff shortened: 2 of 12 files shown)


tests/integration/ha_tests/test_ha.py
```diff
@@ -19,24 +19,31 @@
     clear_db_writes,
     count_primaries,
     count_writes,
+    cut_network_from_unit,
```
marcoppenheimer, line 22: Lmao at this point I'd just import `helpers` and call `helpers.<method>`. 

Definitely don't need to do so in this PR though.

marcoppenheimer (approved):

tests/integration/ha_tests/test_ha.py
MiaAltieri: 🤣

tests/integration/relation_tests/new_relations/test_charm_relations.py
```diff
@@ -368,9 +369,9 @@ async def test_removed_relation_no_longer_has_access(ops_test: OpsTest):
     try:
         client.admin.command("replSetGetStatus")
     except OperationFailure as e:
-        # error code 13 for OperationFailure is an authentication error, meaning disabling of
-        # authentication was unsuccessful
-        if e.code == 13:
+        # error code 18 for OperationFailure is an authentication error, meaning disabling of
```
MiaAltieri, line 372: update to codes

src/charm.py
```diff
@@ -352,9 +353,19 @@ def _on_update_status(self, event):
...
+                self.unit.status = build_unit_status(replset_status, self._unit_ip(self.unit))
+        except ServerSelectionTimeoutError as e:
+            # ServerSelectionTimeoutError is commonly due to ReplicaSetNoPrimary
+            logger.debug("Got error: %s, while checking replica set status", str(e))
+            self.unit.status = WaitingStatus("Waiting for primary re-election..")
+        except AutoReconnect as e:
+            # AutoReconnect is raised when a connection to the database is lost and an attempt to
+            # auto-reconnect will be made by pymongo.
+            logger.debug("Got error: %s, while checking replica set status", str(e))
+            self.unit.status = WaitingStatus("Waiting to reconnect to unit..")
```
marcoppenheimer, line 368: I know it's another comment in the same place, but as you're not using `replset_status` anywhere else in this method, could you modify `build_unit_status` to take `mongodb_config` as an arg instead of `replset_status`, build the status there, and do the context-manager `with`, and the error handling there too?

tests/integration/ha_tests/helpers.py
```diff
@@ -91,26 +91,31 @@ def unit_uri(ip_address: str, password, app=APP_NAME) -> str:
...
 
-async def fetch_primary(replica_set_hosts: List[str], ops_test: OpsTest) -> str:
+async def fetch_primary(
+    replica_set_hosts: List[str], ops_test: OpsTest, down_unit=None, app=None
+) -> str:
     """Returns IP address of current replica set primary."""
     # connect to MongoDB client
-    app = await app_name(ops_test)
-    password = await get_password(ops_test, app)
+    app = app if app else await app_name(ops_test)
```
marcoppenheimer, line 116: This can be `app or await app_name(ops_test)`

tests/integration/ha_tests/helpers.py
```diff
@@ -164,25 +169,25 @@ async def count_primaries(ops_test: OpsTest) -> int:
...
+    down_unit=None,
+    app=None,
 ) -> str:
     """Returns the primary of the replica set.
 
     Retrying 5 times to give the replica set time to elect a new primary, also checks against the
     valid_ips to verify that the primary is not outdated.
     """
-    primary_ip = await fetch_primary(replica_set_hosts, ops_test)
+    app = app if app else await app_name(ops_test)
```
marcoppenheimer, line 182: Same here

tests/integration/ha_tests/helpers.py
```diff
@@ -164,25 +169,25 @@ async def count_primaries(ops_test: OpsTest) -> int:
     wait=wait_exponential(multiplier=1, min=2, max=30),
 )
 async def replica_set_primary(
-    replica_set_hosts: List[str], ops_test: OpsTest, return_name=False
+    replica_set_hosts: List[str],
+    ops_test: OpsTest,
+    down_unit=None,
+    app=None,
 ) -> str:
```
marcoppenheimer, line 176: You have a type-hint here for `str`, whereas on L190 you return type `Unit`, and depend on type `Unit` on L468 is that right?

marcoppenheimer: And you return `None` on L186, so you'll want to hint `Optional` too.

tests/integration/ha_tests/helpers.py
```diff
@@ -685,8 +697,56 @@ async def unit_hostname(ops_test: OpsTest, unit_name: str) -> str:
...
+def wait_network_restore(model_name: str, hostname: str, old_ip: str) -> None:
+    """Wait until network is restored.
+
+    Args:
+        model_name: The name of the model
+        hostname: The name of the instance
+        old_ip: old registered IP address
+    """
+    if instance_ip(model_name, hostname) == old_ip:
+        raise Exception
```
marcoppenheimer, line 735: Here would be a good place to raise an exception with a body message

marcoppenheimer (changes requested):

src/charm.py
MiaAltieri: good idea :)

README.md
```diff
@@ -28,6 +28,12 @@ juju run-action mongodb/<unit_number> get-primary --wait
 
 Similarly, the primary replica is displayed as a status message in `juju status`, however one should note that this hook gets called on regular time intervals and the primary may be outdated if the status hook has not been called recently. 
 
+Further we highly suggest configuring the status hook to run frequently. In addition to reporting the primary, secondaries, and other statuses, the status hook performs self healing in the case of a network cut. To change the frequency of the update status hook do:
+```shell
+juju model-config update-status-hook-interval=<time(s/m/h)>
```
paulomach, line 33: Is there any downsides for doing so in production env?

MiaAltieri: Great question. Obviously one of the drawbacks of this design (which I think you face as well) is that the healing is done only as often as the update status frequency. But AFAICT this is my (and your) only option for network self healing

Fortunately the mongod command that is run in update status is cheap and shouldn't be expensive in terms of run time so this shouldn't slow any hooks down.

However, one notable draw back is this command executing too frequently when the database is under high read requests (or update status running so quickly that it prevents other hooks from running.) I can add something to this effect to the README thanks for the idea @paulomach  :)

paulomach: Sure, this is an issue I guess everyone will face. My assumption is, with bigger deployments this kind of issue tend to dilute, in the sense that one unit being down for the standard interval won't affect the related application(s) as much. So yeah, adding some notice to the users in the README is the way to go.

marceloneppel (approved):

marcoppenheimer (approved):

tests/integration/ha_tests/test_ha.py
```diff
@@ -6,39 +6,12 @@
 import asyncio
 import time
 
+import helpers
```
paulomach, line 9: always good to get rid of some lines!

paulomach (approved):
exquisite!

PR #106, opened 2022-10-06
fixing TLS

MiaAltieri:
### Problem
Both rotating and setting the internal TLS private key resulted in the replicas occasionally losing communication with each other. When I looked at the files in each of the units it the replicas had different internal certificates. This resulted in them not being able to communicate with one another.

After reviewing the code in [ _on_certificate_available](https://github.com/canonical/mongodb-operator/blob/91e5a8611291619d10a702545e1d446d254b934d/lib/charms/mongodb_libs/v0/mongodb_tls.py#L151). The issue is when the leader unit is the _last_ to receive the ` _on_certificate_available` event, when it is last to receive this event it updates the [certs, ca, and chain](https://github.com/canonical/mongodb-operator/blob/91e5a8611291619d10a702545e1d446d254b934d/lib/charms/mongodb_libs/v0/mongodb_tls.py#L173-L177) last. This means all other units have restarted with _old_ cert, ca, and chain. 

### Solution
1. When rotating/reseting certs update this [unit field](https://github.com/canonical/mongodb-operator/blob/86f5f1262ef66ca5012f8804f6e7efd712686cce/lib/charms/mongodb_libs/v0/mongodb_tls.py#L69) and [app field](https://github.com/canonical/mongodb-operator/blob/86f5f1262ef66ca5012f8804f6e7efd712686cce/lib/charms/mongodb_libs/v0/mongodb_tls.py#L77) to None (note the units will still have access to their current app and unit certs, as these certs are stored on each machine at the specified filepath)
2. When receiving a certificate_available event, [wait for both certificates to be set in the app data](https://github.com/canonical/mongodb-operator/blob/86f5f1262ef66ca5012f8804f6e7efd712686cce/lib/charms/mongodb_libs/v0/mongodb_tls.py#L184), this is crucial since the app data is how we copy over the certs to the machine before restarting.

### Changes
1. Updated tls_certs lib to newest version
2. Small change made to `lib/charms/mongodb_libs/v0/mongodb_tls.py` in `_on_certificate_expiring` this change sets the csr before requesting a new one (like we do when we request new CSRs elsewhere in the code). this is important because if `request_certificate_renewal` triggers `cert_available` events on the other units before the secret is set, the other units will not have access to the CSR in the app data 
3. Changed `_on_set_tls_private_key` to set the certs for both internal and external to none. This forces the charm to wait for 

### Do not review 
do not review [lib/charms/tls_certificates_interface/v1/tls_certificates.py](https://github.com/canonical/mongodb-operator/compare/tls-fix?expand=1#diff-1d1ab8a9e9efcaec09c52a01e4fcad28ffa7d7648590db74f19a738c5f86f6b6), this file was updated and this change reflects this

### DIFF
--- a/lib/charms/mongodb_libs/v0/mongodb_tls.py
+++ b/lib/charms/mongodb_libs/v0/mongodb_tls.py
@@ -66,6 +66,7 @@ class MongoDBTLS(Object):
         logger.debug("Request to set TLS private key received.")
         try:
             self._request_certificate("unit", event.params.get("external-key", None))
+            self.charm.set_secret("unit", "cert", None)
             if not self.charm.unit.is_leader():
                 event.log(
                     "Only juju leader unit can set private key for the internal certificate. Skipping."
@@ -73,6 +74,7 @@ class MongoDBTLS(Object):
                 return
 
             self._request_certificate("app", event.params.get("internal-key", None))
+            self.charm.set_secret("app", "cert", None)
             logger.debug("Successfully set TLS private key.")
         except ValueError as e:
             event.fail(str(e))
@@ -178,8 +180,8 @@ class MongoDBTLS(Object):
 
         if renewal and self.substrate == "k8s":
             self.charm.unit.get_container("mongod").stop("mongod")
-            logger.debug("Successfully renewed certificates.")
-        elif not self.charm.get_secret("app", "cert") or not self.charm.get_secret("unit", "cert"):
+
+        if self._waiting_for_certs():
             logger.debug(
                 "Defer till both internal and external TLS certificates available to avoid second restart."
             )
@@ -198,6 +200,17 @@ class MongoDBTLS(Object):
             self.charm.unit.status = ActiveStatus()
         else:
             self.charm.on_mongod_pebble_ready(event)
+
+    def _waiting_for_certs(self):
+        """Returns a boolean indicating whether additional certs are needed."""
+        if not self.charm.get_secret("app", "cert"):
+            logger.debug("Waiting for application certificate.")
+            return True
+        if not self.charm.get_secret("unit", "cert"):
+            logger.debug("Waiting for application certificate.")
+            return True
+
+        return False
 
     def _on_certificate_expiring(self, event: CertificateExpiringEvent) -> None:
         """Request the new certificate when old certificate is expiring."""
@@ -223,11 +236,12 @@ class MongoDBTLS(Object):
             sans=self._get_sans(),
         )
         logger.debug("Requesting a certificate renewal.")
+        self.charm.set_secret(scope, "csr", new_csr.decode("utf-8"))
+
         self.certs.request_certificate_renewal(
             old_certificate_signing_request=old_csr,
             new_certificate_signing_request=new_csr,
         )
-        self.charm.set_secret(scope, "csr", new_csr.decode("utf-8"))
 
     def _get_sans(self) -> List[str]:
         """Create a list of DNS names for a MongoDB unit.
--- a/lib/charms/mongodb_libs/v0/mongodb_tls.py
+++ b/lib/charms/mongodb_libs/v0/mongodb_tls.py
@@ -66,7 +66,7 @@ class MongoDBTLS(Object):
         logger.debug("Request to set TLS private key received.")
         try:
             self._request_certificate("unit", event.params.get("external-key", None))
-            self.charm.set_secret("unit", "cert", None)
+
             if not self.charm.unit.is_leader():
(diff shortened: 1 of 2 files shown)


lib/charms/mongodb_libs/v0/mongodb_tls.py
```diff
@@ -66,13 +66,15 @@ def _on_set_tls_private_key(self, event: ActionEvent) -> None:
         logger.debug("Request to set TLS private key received.")
         try:
             self._request_certificate("unit", event.params.get("external-key", None))
+            self.charm.set_secret("unit", "cert", None)
```
delgod, line 69: Please move this line into the `_request_certificate` call.
the same for line 77

lib/charms/mongodb_libs/v0/mongodb_tls.py
```diff
@@ -223,11 +236,12 @@ def _on_certificate_expiring(self, event: CertificateExpiringEvent) -> None:
             sans=self._get_sans(),
         )
         logger.debug("Requesting a certificate renewal.")
+        self.charm.set_secret(scope, "csr", new_csr.decode("utf-8"))
+
         self.certs.request_certificate_renewal(
             old_certificate_signing_request=old_csr,
             new_certificate_signing_request=new_csr,
         )
-        self.charm.set_secret(scope, "csr", new_csr.decode("utf-8"))
```
delgod, line 230: any reason why we set a secret before requesting a new cert?

MiaAltieri: @delgod see PR description :) (Changes section 2)

MiaAltieri: @delgod yes! see the PR description, Changes section 2. It also keeps consistent with the other ways in which we request new certs and saving CSRs

lib/charms/mongodb_libs/v0/mongodb_tls.py
```diff
@@ -176,16 +178,16 @@ def _on_certificate_available(self, event: CertificateAvailableEvent) -> None:
...
+        if self._waiting_for_certs():
             logger.debug(
                 "Defer till both internal and external TLS certificates available to avoid second restart."
             )
             event.defer()
             return
 
+        if renewal and self.substrate == "k8s":
+            self.charm.unit.get_container("mongod").stop("mongod")
+
```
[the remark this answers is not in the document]
MiaAltieri: great question @marcoppenheimer :) due to the nature of K8s having `on_pebble_ready`, the K8s charm does it in it's [on_pebble_ready handler](https://github.com/delgod/mongodb-k8s-operator/blob/3d31905436df7c1b4a776e7d99c996857757afc3/src/charm.py#L86) 

(so in short we do push the cert to the workload on K8s you just don't see it here)

marcoppenheimer (approved):

lib/charms/mongodb_libs/v0/mongodb_tls.py
```diff
@@ -78,6 +79,7 @@ def _on_set_tls_private_key(self, event: ActionEvent) -> None:
             event.fail(str(e))
 
     def _request_certificate(self, scope: str, param: Optional[str]):
+        self.charm.set_secret(scope, "cert", None)
```
delgod, line 82: maybe we shouldn't assign None if the failure happened inside the generate_csr function.
in other words, we can move this line after `self.charm.set_secret(scope, "csr", csr.decode("utf-8"))` line

lib/charms/mongodb_libs/v0/mongodb_tls.py
delgod: > Small change made to lib/charms/mongodb_libs/v0/mongodb_tls.py in _on_certificate_expiring this change sets the csr before requesting a new one (like we do when we request new CSRs elsewhere in the code). this is important because if request_certificate_renewal triggers cert_available events on the other units before the secret is set, the other units will not have access to the CSR in the app data

`request_certificate_renewal` make a change to the TLS relation.
`set_secret` make a change to the peer relation.
Both changes become applied only when the event is fully finished.

MiaAltieri: Smart, gotcha, thanks Mykola :)

lib/charms/mongodb_libs/v0/mongodb_tls.py
MiaAltieri: good idea, thanks Mykola :)

delgod (approved):

PR #110, opened 2022-10-11
fix flakey relation test 

MiaAltieri:
### Problem
Relation test regarding metadata between application and mongoDB is flakey

### Solution
1. `replica_set_primary` is a helper for the HA tests which looks for the first MongoDB application deployed and finds the primary of that application. Since these relation tests have multiple MongoDB applications, it is not guaranteed that `replica_set_primary` will look for the primary of our desired applications. Thus we pass in an option that allows `replica_set_primary` to be run on a specific application
2. Further, some conditions need some time to be met. So we add a retry to wait for condition to be met. 

### Other
Code was also reformatted to reduce redundancy.

### DIFF
--- a/tests/integration/ha_tests/helpers.py
+++ b/tests/integration/ha_tests/helpers.py
@@ -108,10 +108,10 @@ async def get_password(ops_test: OpsTest, app) -> str:
     return action.results["admin-password"]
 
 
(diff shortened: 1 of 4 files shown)


averma-canonical (approved):

paulomach (approved):
Nice

PR #116, opened 2022-10-19
Fix hanging units - "Waiting to Initialise Replica Set"

MiaAltieri:
This PR is a fix for #87 

### Issue
On rare occasions in a deployment of the MongoDB charm one or more units will be failed to be added to the replica set and the leader unit will report "Waiting to Initialise Replica Set" or "Waiting to reconfigure replica set"

### Context
After reproducing this issue it was discovered that this is due to one of the replicas not having `mongod` running. When insepcting with `systemd` it seems that this is due to a failure to restart by retrying to restart too many times and then giving up.

### Solution
This check  does a fail-safe self healing in case the automated self healing from systemd fails

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -347,6 +347,22 @@ class MongodbOperatorCharm(ops.charm.CharmBase):
             self.unit.status = BlockedStatus("cannot have both legacy and new relations")
             return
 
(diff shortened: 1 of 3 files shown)


WRFitch (approved):

welpaolo (approved):
LGTM

PR #119, opened 2022-10-21
remove lib deps

MiaAltieri:
### Problem
1. MongoDB libs require `systemd` from the `operator libs linux` libs, this means that the k8s charm would have to import this library even though it doesn't use it
2. Relations handle `systemd` failures. After some thought it is better if this fails loudly so we raise an error in this case. If restarting fails we should blow up noisly and go into error. I think going into blocked here isn't the most useful, especially since that can easily get overwritten with a less informative message and then its harder to determine why the cluster/unit is failing.

### Solution
Remove the import of `systemd` and blow up noisly by raising an exception

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -746,18 +746,14 @@ class MongodbOperatorCharm(ops.charm.CharmBase):
         """Restarts the mongod service with its associated configuration."""
         if auth is None:
             auth = self.auth_enabled()
-        try:
-            stop_mongod_service()
-            update_mongod_service(
-                auth,
-                self._unit_ip(self.unit),
-                config=self.mongodb_config,
-            )
-            start_mongod_service()
-            return True
-        except systemd.SystemdError as e:
-            logger.error("Failed to restart systemd, error: %s", str(e))
-            raise
+
+        stop_mongod_service()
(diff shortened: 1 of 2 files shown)


src/charm.py
```diff
@@ -746,14 +746,18 @@ def restart_mongod_service(self, auth=None):
...
+        try:
+            stop_mongod_service()
+            update_mongod_service(
+                auth,
+                self._unit_ip(self.unit),
+                config=self.mongodb_config,
+            )
+            start_mongod_service()
+            return True
+        except systemd.SystemdError as e:
```
zmraul, line 758: This method is always called as a void. So I think the `return True` is unnecessary.
Also, the exception is not handled anywhere. It's explicit here, but if it wasn't, it would raise regardless and show on the `juju-debug` log.
Also also, I think you can remove `restart_mongod_service` from `src/machine_helpers.py`?

zmraul (changes requested):

src/charm.py
MiaAltieri: great feedback thank you!

WRFitch (approved):

PR #121, opened 2022-10-21
ci: update charming-actions

jnsgruk:
Talking to @taurus-forever about the recently merged change, I noticed that we didn't bump the actual upload action, and also that we weren't actually using the `charming-actions/channel` action.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -20,11 +20,8 @@ jobs:
         uses: actions/checkout@v2
         with:
           fetch-depth: 0
-      - name: Select charmhub channel
-        uses: canonical/charming-actions/upload-charm@2.1.1
-        id: channel
       - name: Upload charm to charmhub
-        uses: canonical/charming-actions/upload-charm@2.0.0-rc
+        uses: canonical/charming-actions/upload-charm@2.1.1
         with:
           credentials: "${{ secrets.CHARMHUB_TOKEN }}"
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -20,11 +20,8 @@ jobs:
         uses: actions/checkout@v2
         with:
           fetch-depth: 0
-      - name: Select charmhub channel
-        uses: canonical/charming-actions/upload-charm@2.1.1
-        id: channel
```
taurus-forever, line 25: Actually other repos uses "channel" here:

      - name: Select charmhub channel
        uses: canonical/charming-actions/channel@2.1.1
        id: channel

Is it necessary here too?

jnsgruk: No, because we're hard coding the channel in the next step anyway - so it negates the need for the `channel`action :)

MiaAltieri (approved):

.github/workflows/release.yaml
taurus-forever: TIL. TL;DR: not necessary here as we have hardcoded channel name in line "31 28":
>  channel: "dpe/edge"

Necessary here only in the case:
> channel: "${{ steps.channel.outputs.name }}"
See https://github.com/canonical/charming-actions/tree/main/channel for more details.

taurus-forever (approved):

PR #131, opened 2022-12-02
Tutorial pt2: deployment and db access

MiaAltieri:
Adding:

1. Deploy charm - verify deployment is running with Juju Status
2. Access the DB via the admin user with PyMongo (give a disclaimer to not use the Admin user in production environment

to the tutorial

### DIFF
--- a/documentation/tutorial.md
+++ b/documentation/tutorial.md
@@ -100,7 +100,8 @@ Machine  State    Address       Inst id        Series  AZ  Message
 ```
 
 ## Accessing MongoDB:
-*Disclaimer: this part of the tutorial accesses MongoDB via the `admin` user. It is not recommended that you directly interface with the admin user in a production environment. In a production environment [create a separate user](https://www.mongodb.com/docs/manual/tutorial/create-users/) and connect to MongoDB with that user instead.*
+*Disclaimer: this part of the tutorial accesses MongoDB via the `admin` user. It is not recommended that you directly interface with the admin user in a production environment. In a production environment [create a separate user](https://www.mongodb.com/docs/manual/tutorial/create-users/) and connect to MongoDB with that user instead. Later in the section covering Relations we will cover how to access MongoDB without the admin user.
+*
 
 For this part of the Tutorial we will access MongoDB - via the MongoDB shell `mongosh`. You can read more about the MongoDB shell [here](https://www.mongodb.com/docs/mongodb-shell/).
 
@@ -158,7 +159,7 @@ Machine  State    Address       Inst id        Series  AZ  Message
 Use the IP address listed underneath `Public address` for `mongodb/0` as your host.
 
 #### Retrieving the database name:
-In this case we are logging in via the `admin` user so we will be connecting to the `admin` database. Use `admin` as the database name.
+In this case we are logging in via the `admin` user so we will be connecting to the `admin` database. Use `admin` as the database name. Once we access the database via the MongoDB URI, we will create a `test-db` database to store data.
 
 #### Retrieving the replica set name:
 The replica set name is the name of the application. In this tutorial we didn’t use a custom name for the application, so the application name is `mongodb`. You can read more about deploying a charm with a custom name [here](https://juju.is/docs/olm/deploy-a-charm-from-charmhub#heading--override-the-name-of-a-deployed-application). Use `mongodb` as the replica set name.
@@ -209,4 +210,31 @@ config  120.00 KiB
 local   404.00 KiB
 ```
 
-Feel free to test out any other MongoDB commands, when you’re ready to leave the shell you can just type `exit`
+Now that we have access to MongoDB we can create a database named `test-db`. To create this database enter:
+```
+use test-db
+```
+Now lets create a user called `testUser` read/write access to the database `test-db` that we just created . Enter:
+```
+db.createUser({
+  user: "testUser",
+  pwd: "password",
+  roles: [
+    { role: "readWrite", db: "test-db" }
+  ]
+})
+```
+You can verify that you added the user correctly, by entering the command `show users` into the mongo shell, this will output:
+```
+[
+  {
+    _id: 'test-db.testUser',
+    userId: new UUID("6e841e28-b1bc-4719-bf42-ba4b164fc546"),
+    user: 'testUser',
+    db: 'test-db',
+    roles: [ { role: 'readWrite', db: 'test-db' } ],
+    mechanisms: [ 'SCRAM-SHA-1', 'SCRAM-SHA-256' ]
+  }
+]
+```
+Feel free to test out any other MongoDB commands, when you’re ready to leave the shell you can just type `exit`.
--- a/documentation/tutorial.md
+++ b/documentation/tutorial.md
@@ -68,3 +68,145 @@ tutorial overlord    localhost/localhost  2.9.37   unsupported  23:20:53Z
 Model "admin/tutorial" is empty.
 ```
 
+
+## Deploying Charmed MongoDB:
+To deploy Charmed MongoDB, fetch the charm from [Charmhub](https://charmhub.io/mongodb?channel=dpe/edge) and deploy it to your model:
+```
+juju deploy mongodb --channel dpe/edge
+```
+
+Juju will now fetch Charmed MongoDB and begin deploying it to the LXD cloud. This process can take several minutes depending on how provisioned your machine is. You can track the progress by running:
+```
+watch -c juju status --color
+```
+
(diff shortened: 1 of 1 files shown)


documentation/tutorial.md
```diff
@@ -68,3 +68,145 @@ tutorial overlord    localhost/localhost  2.9.37   unsupported  23:20:53Z
...
+mongodb/0*  active    idle   0        <host IP>    27017/tcp  Replica set primary
+
+Machine  State    Address       Inst id        Series  AZ  Message
+0        started  10.23.62.156  juju-d35d30-0  focal       Running
+
+```
+Use the IP address listed underneath `Public address` for `mongodb/0` as your host.
+
+#### Retrieving the database name:
+In this case we are logging in via the `admin` user so we will be connecting to the `admin` database. Use `admin` as the database name.
```
deusebio, line 161:
```suggestion
In this case we are logging in via the `admin` user so we will be connecting to the `admin` database. Use `admin` as the database name. We will in the following create a "test" database where to store some data. 
```

documentation/tutorial.md
```diff
@@ -68,3 +68,145 @@ tutorial overlord    localhost/localhost  2.9.37   unsupported  23:20:53Z
...
+mongodb [primary] admin>
+```
+
+You can now interact with MongoDB directly using any [MongoDB commands](https://www.mongodb.com/docs/manual/reference/command/). For example entering `show dbs` should output something like:
+```
+admin   172.00 KiB
+config  120.00 KiB
+local   404.00 KiB
+```
+
```
deusebio, line 211: I would just type 3 commands, like:

1. Creating a new "test" database
2. Add a user
3. Create a collection with RW operations for the user

documentation/tutorial.md
```diff
@@ -68,3 +68,145 @@ tutorial overlord    localhost/localhost  2.9.37   unsupported  23:20:53Z
...
+## Accessing MongoDB:
+*Disclaimer: this part of the tutorial accesses MongoDB via the `admin` user. It is not recommended that you directly interface with the admin user in a production environment. In a production environment [create a separate user](https://www.mongodb.com/docs/manual/tutorial/create-users/) and connect to MongoDB with that user instead.*
+
+For this part of the Tutorial we will access MongoDB - via the MongoDB shell `mongosh`. You can read more about the MongoDB shell [here](https://www.mongodb.com/docs/mongodb-shell/).
+
+
+### Install the MongoDB shell:
+While MongoDB is installed within the LXD containers that the application is hosted on, it is not installed on your machine. This means that you will have to install the MongoDB shell yourself. Install `mongosh` by following the official [instructions on how to install the MongoDB shell](https://www.mongodb.com/docs/mongodb-shell/install/#std-label-mdb-shell-install).
```
deusebio, line 109: what about launching an lxd container con mongoDB instead? This way we don't mess up with the user environment

deusebio (commented):
Cool! I really like the way you lead the reader through it, with just the right amount of depth,  proving extra links if one wants to deepen the concepts!

Some comments attached. I suppose the more relevant one is about running mongo shell within a lxd container. 

If we agree, when this bit is done, I will go ahead and tests the tutorial line by line

documentation/tutorial.md
MiaAltieri: hmm interesting point. If we didn't want to "mess up with the user environment" we could just `ssh mongodb/unit` and then access mongodb via mongosh there (since mongosh is already installed). But I think doing either of these abstracts things a little too much from the reader. 

But also I don't think installing mongosh really messes up the environment since the user is also installing other packages in this tutorial

deusebio: Yes. It is not something major that I feel very strong about. But let's see what the other reviewer (@zmraul ) thinks about this

documentation/tutorial.md
MiaAltieri: I don't think this is the correct place for this. What do you think about putting this further down on line 167

documentation/tutorial.md
WRFitch: 2c: In SQL-land there's actually dedicated charms for shells, but that's obviously a lot more work. If I was running this charm, I'd expect to need a local shell if there wasn't a dedicated shell charm for me to use.

documentation/tutorial.md
```diff
@@ -68,3 +68,145 @@ tutorial overlord    localhost/localhost  2.9.37   unsupported  23:20:53Z
...
+
+### MongoDB URI:
+Connecting to the database requires a [URI](https://www.mongodb.com/docs/manual/reference/connection-string/). We use a URI of the format:
```
WRFitch, line 112: Perhaps document that this is provided to related charms

WRFitch (approved):

documentation/tutorial.md
deusebio: yes. But, I mean, but when I "tell a story", sometimes I believe it is good to create anticipations and connect different part of the story. This works on many things I have been working on: creating a pitch, a presentation or (maybe) a tutorial. It should not however distract the reader/audience but rather get him interested. I believe this was a good point to do that, but it is not a strong point, if you and @WRFitch feel otherwise. Beside, I really don't know how Daniele or Nick would feel about this, did not double check this with them to be fair.

documentation/tutorial.md
deusebio: Uhm, we don't have that and of course we should definitely not implement that within the context of this task. Should we then rather add this in the requirement bit?

documentation/tutorial.md
```diff
@@ -68,3 +68,145 @@ tutorial overlord    localhost/localhost  2.9.37   unsupported  23:20:53Z
...
+
+Unit        Workload  Agent  Machine  Public address  Ports      Message
+mongodb/0*  active    idle   0        10.23.62.156    27017/tcp
+
+Machine  State    Address       Inst id        Series  AZ  Message
+0        started  10.23.62.156  juju-d35d30-0  focal       Running
+```
+
+## Accessing MongoDB:
+*Disclaimer: this part of the tutorial accesses MongoDB via the `admin` user. It is not recommended that you directly interface with the admin user in a production environment. In a production environment [create a separate user](https://www.mongodb.com/docs/manual/tutorial/create-users/) and connect to MongoDB with that user instead.*
```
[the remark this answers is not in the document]
zmraul: Hm, sections are not numbered so can section 4 be found at a glance?

Also, the purpose of a tutorial, as I understood from Daniele, is to remove as much ambiguity and choices as possible. Adding the note where we explain that this is not a production setup is good, but I don't see the use on having a link to a different section. I think the goal is to minimize branching attention to things, and just focus on completing the tutorial, step by step.

documentation/tutorial.md
zmraul: I think installing `mongosh` on the host is ok here. My other choice would be to `ssh` into the unit to see it working as Mia said.

zmraul (approved):
This is a very thorough tutorial :) nicely done!

documentation/tutorial.md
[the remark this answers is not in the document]
MiaAltieri: ooh, great point! What if instead of branching we state:
```
Later in [the section covering Relations](#relations) we will cover how to access MongoDB without the admin user.
```

(note section \#relations  will be added in a following PR). Thoughts @zmraul and @deusebio

documentation/tutorial.md
MiaAltieri: Can we get a second opinion on this @WRFitch and @zmraul 🙏

documentation/tutorial.md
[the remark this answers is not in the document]
MiaAltieri: I could go either way, @deusebio what do you think?

documentation/tutorial.md
MiaAltieri: okay cool lets keep the install of mongosh, thank you everyone

documentation/tutorial.md
zmraul: I don't have an opinion on this. I think it's just personal preference.

documentation/tutorial.md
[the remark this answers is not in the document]
deusebio: I would drop the link, while retaining the reference that this will be covered later. It would be best to require reader to read through, without skipping parts (as a link would nudge)

documentation/tutorial.md
MiaAltieri: Minor tweak: MongoDB wise and tutorial wise I think it will make more sense to add a user which has RW operations over a database not a collection. So I'll modify this a smidge

deusebio (approved):
Good to go!

PR #132, opened 2022-12-02
Tutorial pt3: Scaling + Passwords

MiaAltieri:
Adding:

1. Scale charm up - verify deployment is running with Juju Status
2. Scale charm down - verify deployment is running with Juju Status
3. Access Password 
4. Rotate Password - verify change by accessing password
5. Change Password to a specific password - verify change by accessing password

to tutorial

### DIFF
--- a/documentation/tutorial.md
+++ b/documentation/tutorial.md
@@ -141,7 +141,7 @@ unit-mongodb-0:
 ```
 Use the password under the result: `admin-password`.
 
-**Retrieving the hosts:** The hosts are the units hosting the MongoDB application. In this case we only have one host; you can find the host’s IP address with `juju status`. This will output information about the Charmed MongoDB application along with it’s IP address:
+**Retrieving the hosts:** The hosts are the units hosting the MongoDB application. The host’s IP address can be found with `juju status`:
 ```
 Model     Controller  Cloud/Region         Version  SLA          Timestamp
 tutorial  overlord    localhost/localhost  2.9.37   unsupported  11:31:16Z
@@ -158,9 +158,9 @@ Machine  State    Address       Inst id        Series  AZ  Message
 ```
 Use the IP address listed underneath `Public address` for `mongodb/0` as your host.
 
-**Retrieving the database name:** In this case we are logging in via the `admin` user so we will be connecting to the `admin` database. Use `admin` as the database name. Once we access the database via the MongoDB URI, we will create a `test-db` database to store data.
+**Retrieving the database name:** In this case we are connecting to the `admin` database. Use `admin` as the database name. Once we access the database via the MongoDB URI, we will create a `test-db` database to store data.
 
-**Retrieving the replica set name:** The replica set name is the name of the application. In this tutorial we didn’t use a custom name for the application, so the application name is `mongodb`. You can read more about deploying a charm with a custom name [here](https://juju.is/docs/olm/deploy-a-charm-from-charmhub#heading--override-the-name-of-a-deployed-application). Use `mongodb` as the replica set name.
+**Retrieving the replica set name:** The replica set name is the name of the application on Juju hosting MongoDB. The application name in this tutorial is `mongodb`. Use `mongodb` as the replica set name. 
 
 ### Connect via MongoDB URI:
 Now that we have the necessary fields to connect to the URI, we can connect to MongoDB via the URI. Enter the following into the command line, replace the values for  `username`, `password`, `hosts`, `database name`, and the `replica set name` with what you’ve retrieved above:
--- a/documentation/tutorial.md
+++ b/documentation/tutorial.md
@@ -1,6 +1,7 @@
 # Charmed MongoDB tutorial:
 Ready to try out Charmed MongoDB? This tutorial will guide you through the steps to get Charmed MongoDB up and running. 
 
+
 ## Minimum requirements:
 Before we start be sure you have the following requirements:
 - Ubuntu 20.04(focal) or later.
@@ -8,6 +9,7 @@ Before we start be sure you have the following requirements:
 - 2 CPU threads.
 - At least 20GB of available storage.
 - Access to the internet for downloading the required snaps and charms.
+
 
 ## Install and prepare LXD:
 The fastest, simplest way to get started with Charmed MongoDB is to set up a local LXD cloud. LXD is a system container and virtual machine manager; Charmed MongoDB will be run in one of these containers and managed by Juju. While this tutorial covers the basics of LXD, you can explore more LXD [here](https://linuxcontainers.org/lxd/getting-started-cli/). The first step on our journey is to install LXD. LXD is installed from a snap package:
@@ -33,6 +35,7 @@ You can list all LXD containers by entering the command `lxc list` in to the com
 | NAME | STATE | IPV4 | IPV6 | TYPE | SNAPSHOTS |
 +------+-------+------+------+------+-----------+
 ```
+
 
 ## Install and prepare Juju:
 [Juju](https://juju.is/) is an operation Lifecycle manager(OLM) for clouds, bare metal, LXD or Kubernetes. We will be using it to deploy and manage Charmed MongoDB. As with LXD, Juju is installed from a snap package:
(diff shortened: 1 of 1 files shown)


documentation/tutorial.md
```diff
@@ -68,3 +71,416 @@ tutorial overlord    localhost/localhost  2.9.37   unsupported  23:20:53Z
...
+admin   172.00 KiB
+config  120.00 KiB
+local   404.00 KiB
+```
+
+Feel free to test out any other MongoDB commands, when you’re ready to leave the shell you can just type `exit`
+
+
+## Scaling Charmed MongoDB: 
+Replication is a popular feature of MongoDB; replicas copy data making a database highly reliable.
```
deusebio, line 227:
```suggestion
Replication is a popular feature of MongoDB; replicas copy data making a database highly available. 
```

documentation/tutorial.md
```diff
@@ -68,3 +71,416 @@ tutorial overlord    localhost/localhost  2.9.37   unsupported  23:20:53Z
...
+
+Machine  State    Address       Inst id        Series  AZ  Message
+0        started  10.23.62.156  juju-d35d30-0  focal       Running
+1        started  10.23.62.55   juju-d35d30-1  focal       Running
+2        started  10.23.62.243  juju-d35d30-2  focal       Running
+```
+
+You can trust that Charmed MongoDB added these replicas correctly. But if you wanted to verify the replicas got added correctly you could connect to MongoDB via `mongosh`. Since your replica set has 2 additional hosts you will need to update the hosts in your URI. Using the IP addresses from `juju status` add all of your hosts to the URI and separate them with commas. Your URI should look like this:
+```
+mongodb://<username>:<password>@<host1,<host2>,<host3>>/<database name>?replicaSet=<replica set name>
```
deusebio, line 259: Uhm, just wondering here... Do we have a way to provide this information to the user out of the box? Like an action that would return this info. It is not something to be worked on in this task, but I thought it was just worth mentioning that this is something nice to provide to the user, if not already there (in this case I would put how to do this in the tutorial...)

documentation/tutorial.md
```diff
@@ -68,3 +71,416 @@ tutorial overlord    localhost/localhost  2.9.37   unsupported  23:20:53Z
...
+mongodb/0*  active    idle   0        10.23.62.156    27017/tcp  Replica set primary
+mongodb/1   active    idle   1        10.23.62.55     27017/tcp  Replica set secondary
+
+Machine  State    Address       Inst id        Series  AZ  Message
+0        started  10.23.62.156  juju-d35d30-0  focal       Running
+1        started  10.23.62.55   juju-d35d30-1  focal       Running
+
+```
+
+As previously mentioned you can trust that Charmed MongoDB removed this replica correctly.
```
deusebio, line 416: Is there any check one could do here to make sure this has been done correctly? If we had the action to retrieve the URI, one could retrieve the new URI and make a check call. But if we don't have that info, I'm not sure. But, as a reader, when I read "you can trust" I'm thinking, "is there a way I can make sure or should I really take your word for"? :D

documentation/tutorial.md
```diff
@@ -68,3 +71,416 @@ tutorial overlord    localhost/localhost  2.9.37   unsupported  23:20:53Z
...
+  timing:
+    completed: 2022-12-02 14:53:30 +0000 UTC
+    enqueued: 2022-12-02 14:53:25 +0000 UTC
+    started: 2022-12-02 14:53:28 +0000 UTC
+```
+The admin password under the result: `admin-password` should match whatever you passed in when you entered the command.
+
+*Note when you change the admin password you will also need to update the admin password the in MongoDB URI; as the old password will no longer be valid*
+
+
```
deusebio, line 485: Maybe I would also show off the juju commands to list all actions that have now been executed on our deployment. Was it somehting like "juju operations" ? (I think we actions v2 interface, but I suppose there is the same in the v1 interface, isn't it?)

deusebio (approved):
All good! Just some minor comments attached!

documentation/tutorial.md
MiaAltieri: GREAT question. IMO from a **strict** MongoDB point of view those instructions are the **only way** to verify. (other than writing data to the primary replica and checking it copied on the secondary) 

But if we are being less strict, the user could verify this by typing `juju status` and seeing:
```
Model     Controller  Cloud/Region         Version  SLA          Timestamp
tutorial  overlord    localhost/localhost  2.9.37   unsupported  10:30:16Z

App                  Version  Status  Scale  Charm                Channel   Rev  Exposed  Message
mongodb                       active      2  mongodb              dpe/edge   96  no       Replica set secondary

Unit                    Workload  Agent  Machine  Public address  Ports      Message
mongodb/0*              active    idle   0        10.23.62.156    27017/tcp  Replica set secondary
mongodb/1               active    idle   1        10.23.62.55     27017/tcp  Replica set primary

Machine  State    Address       Inst id        Series  AZ  Message
0        started  10.23.62.156  juju-d35d30-0  focal       Running
1        started  10.23.62.55   juju-d35d30-1  focal       Running
```
(however messages such as `replica set primary` and `replica set secondary` only appear once the update status hook is triggered which can take up to 5 minutes. (unless the hook is triggered to fire quicker -- which is easy to do)

documentation/tutorial.md
MiaAltieri: good question we could repeat the steps from above and show the replica set configuration. (illustrating that the replica is removed in the config) what do you think of that?

 as for making an action to retrieve the URI - 
- creating a new action requires: creating specifications and approve them via the VP and PMs
- we could however update the `get-password` action such that it outputs the URI

thoughts?

documentation/tutorial.md
MiaAltieri: 🤦‍♀️ oops thank you

documentation/tutorial.md
MiaAltieri: Not sure:
```
juju operations
ERROR juju: "operations" is not a juju command. See "juju --help".
```

But also I don't think this is necessary for learning how to use Charmed MongoDB. We could also just do `history | grep juju`

documentation/tutorial.md
deusebio: so, there is no action that returns the uri? I suppose one could do the "juju show-unit" to see the databag though. It is a bit hacky, but figuring this out all the time bit by bit (username, passwords, hosts, replica_set name) can be a bit tedious. It is good for the tutorial, but it would also be good to have either an action or using the show-unit, as a tip

documentation/tutorial.md
deusebio: No, that's fine. By the way, the task at hand is about writing the tutorial. I would go down implementing actions for this. I'm just mentioning that I can see a use-case for this to be userful for the user. I would not repeat all the steps one more. Possibly we can say:

```suggestion
As previously mentioned you can trust that Charmed MongoDB removed this replica correctly. This can be checked however by verifying that the new URI (where the removed host has been excluded) works properly.
```

documentation/tutorial.md
deusebio: I believe you should first change the API interface using 

```
export JUJU_FEATURES=actions-v2
```

I believe I saw this in Hello Kubecon workshop by Jon 

Anyhow, there should be a way to see this in the actions-v1 interface too I suppose. Maybe juju list-*?

MiaAltieri: Hmm; regardless of this I don't think seeing the previously used Juju commands adds much to the tutorial

documentation/tutorial.md
```diff
@@ -68,3 +71,416 @@ tutorial overlord    localhost/localhost  2.9.37   unsupported  23:20:53Z
...
+### MongoDB URI:
+Connecting to the database requires a [URI](https://www.mongodb.com/docs/manual/reference/connection-string/). We use a URI of the format:
+```
+mongodb://<username>:<password>@<host>/<database name>?replicaSet=<replica set name>
+```
+
+Connecting via the URI requires that you know the values for `username`, `password`, `host`, `database name`, and the `replica set name`. We will now show you how to retrieve the necessary fields. 
+
+
+#### Retrieving the username:
```
zmraul, line 125: This is something that I'm noticing now but was present on previous PRs. I feel like this area has too many titles with little to explain under them. Instead of having a ton of `####` titles, maybe just streamline it with just a quick explanation of default values. I wouldn't say that explaining why those values are like that should be part of the tutorial.

zmraul (commented):
Looking good :)

documentation/tutorial.md
MiaAltieri: Agree about the titles. Do you think the explainations should be changed?

zmraul (approved):
:+1:

PR #133, opened 2022-12-05
change to set-password and get-password

MiaAltieri:
[JIRA](https://warthogs.atlassian.net/jira/software/c/projects/DPE/boards/874?view=detail&selectedIssue=DPE-1002&quickFilter=2360)

### DIFF
--- a/actions.yaml
+++ b/actions.yaml
@@ -5,10 +5,12 @@ get-primary:
   description: Report primary replica
 
 get-password:
-  description: Get the admin user password used by charm.
+  description: Change the admin user's password, which is used by charm.
+    It is for internal charm users and SHOULD NOT be used by applications.
 
 set-password:
-  description: Change the admin user password used by charm.
+  description: Change the admin user's password, which is used by charm.
+    It is for internal charm users and SHOULD NOT be used by applications.
   params:
     password:
       type: string
--- a/actions.yaml
+++ b/actions.yaml
@@ -4,10 +4,10 @@
 get-primary:
   description: Report primary replica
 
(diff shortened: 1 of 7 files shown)


WRFitch (approved):
LGTM once HA tests pass :)

actions.yaml
```diff
@@ -4,10 +4,10 @@
...
 
-get-admin-password:
+get-password:
   description: Get the admin user password used by charm.
 
-set-admin-password:
+set-password:
   description: Change the admin user password used by charm.
   params:
     password:
```
deusebio, line 13: In Kafka , the signature for set-password also has a `username` parameter

https://github.com/canonical/kafka-operator/blob/main/actions.yaml#L4

Should we try to get a unified UX experience for parameters too?

Probably it would be best to get the same description across

deusebio (approved):
The code modifications look all right!

I'm just pointing out that, although this ticket is about unifying UX, the UX with respect to Kafka is still somewhat different. 

I don't know whether there is a common view on what the "get-password" and "set-password" signature should be. If it has not been decided yet, we can merge this PR, but I can see some value on promoting unified actions UX.

actions.yaml
MiaAltieri: Agreed about the description - but I don't think it makes sense if we only have the admin user in MongoDB to have a user parameter. What do you think @deusebio

deusebio: Oh, I see. Probably it is really not possible to get a fully unified UX. Let's see whether this is really something we want to get at a later stage.

Issue #138, opened 2022-12-12
`KeyError: 'password'` after relating to application


carlcsaposs-canonical:
# To Reproduce
```sh
juju add-model tutorial
juju deploy mongodb --channel dpe/edge
git clone https://github.com/MiaAltieri/application-charm.git
cd application-charm/
charmcraft pack
juju deploy ./*charm
juju relate mongodb application
```
# Environment
<details>
<summary>cloud-init</summary>

```yaml
#cloud-config
package_upgrade: true
packages:
  - gnome-keyring
  - retry
  - tox
snap:
  commands:
    - snap refresh
    - snap install juju --classic --channel=2.9/stable
    - snap install charmcraft --classic
    - snap install lxd
    - snap install microk8s --classic
    - snap alias microk8s.kubectl kubectl
    - snap install jhack
    - snap connect jhack:dot-local-share-juju snapd
runcmd:
  - adduser ubuntu lxd
  - newgrp lxd
  - lxd init --auto
  - lxc network set lxdbr0 ipv6.address none
  # https://linuxcontainers.org/lxd/docs/master/howto/network_bridge_firewalld/#ufw-add-rules-for-the-bridge
  - ufw allow in on lxdbr0
  - ufw route allow in on lxdbr0
  - ufw route allow out on lxdbr0
  - adduser ubuntu microk8s
  - newgrp microk8s
  - microk8s status --wait-ready
  - retry --times 3 --delay 5 microk8s enable dns
  - microk8s status --wait-ready
  - kubectl rollout status --namespace kube-system --watch --timeout=5m deployments/coredns
  - microk8s enable hostpath-storage
  - kubectl rollout status --namespace kube-system --watch --timeout=5m deployments/hostpath-provisioner
  - runuser ubuntu --login --command "juju bootstrap microk8s micro --agent-version=2.9.29"
  - runuser ubuntu --login --command "juju bootstrap localhost lxd --agent-version=2.9.29"
  - runuser ubuntu --login --command "juju model-defaults logging-config='<root>=INFO; unit=DEBUG'"
  - runuser ubuntu --login --command "microk8s config > ~/.kube/config"
  - runuser ubuntu --login --command "ssh-keygen -t ed25519 -f ~/.ssh/id_ed25519 -N ''"
  - runuser ubuntu --login --command "cat ~/.ssh/id_ed25519.pub"

```
</details>

Google Cloud compute engine VM on Ubuntu 22.04
<details>
<summary>full details</summary>

[...]

PR #144, opened 2022-12-14
BUG FIX check for user before updating data

MiaAltieri:
### Bug
When relating MongoDB with another charm via the `database` interface. It is possible that the charm attempts to update the relation hosts before a user has been created for that relation. Causing #138 

### Solution
Add a check that verifies that the user has been created before updating information for that user


delgod (approved):

dratushnyy (approved):

dmitry-ratushnyy (approved):

MiaAltieri:
Out of Date

PR #145, opened 2022-12-16
charm creates the PBM user which is needed for backups

MiaAltieri:
## Summary
For the reader, creating backups requires a bit of necessary infrastructure. The first step is to create the PBM user which is needed for backups. 

## Context
Backups for Charmed MongoDB make use of [PBM](https://docs.percona.com/percona-backup-mongodb/index.html). For this package to work correctly and to make backups a user needs to exist on the MongoDB database

## Future PRs
As mentioned this is the first PR of several to get the backup action up and running. The rest are:
- Configuration options for PBM configurations in S3 storage
- Implement the backup action
- Add unit tests
- Integration tests

### DIFF
--- a/lib/charms/mongodb/v0/mongodb_backups.py
+++ b/lib/charms/mongodb/v0/mongodb_backups.py
@@ -1,0 +1,99 @@
+# Copyright 2022 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+"""In this class, we manage client database relations.
+
+This class creates a user and database for each application relation
+and expose needed information for client connection via fields in
+external relation.
+"""
+import logging
+
+from charms.mongodb.v0.helpers import generate_password
+from charms.mongodb.v0.mongodb import MongoDBConfiguration, MongoDBConnection
+from charms.operator_libs_linux.v1 import snap
+from ops.framework import Object
+
+# The unique Charmhub library identifier, never change it
+LIBID = "asdfl29130831jkh20apj09q1dkos0"
+
+# Increment this major API version when introducing breaking changes
+LIBAPI = 0
+
+# Increment this PATCH version before using `charmcraft publish-lib` or reset
+# to 0 if you are raising the major API version
+LIBPATCH = 0
+
+logger = logging.getLogger(__name__)
+
+
+PBM_S3_CONFIGS = [
+    ("storage.s3.region", "s3-storage-region"),
+    ("storage.s3.bucket", "s3-storage-bucket"),
+    ("storage.s3.prefix", "s3-storage-prefix"),
+    ("storage.s3.credentials.access-key-id", "s3-access-key-id"),
+    ("storage.s3.credentials.secret-access-key", "s3-secret-access-key"),
+    ("storage.s3.serverSideEncryption.kmsKeyID", "s3-kms-key-id"),
+]
+
+PBM_PRIVILEGES = {"resource": {"anyResource": True}, "actions": ["anyAction"]}
+
+
+class MongoDBBackups(Object):
+    """In this class, we manage client database relations."""
+
+    def __init__(self, charm, substrate="k8s"):
+        """Manager of MongoDB client relations."""
+        super().__init__(charm, "client-relations")
+        self.charm = charm
+        self.substrate = substrate
+        self.framework.observe(self.charm.on.config_changed, self._on_pbm_config_changed)
+
+    def _on_pbm_config_changed(self, event) -> None:
+        """Handles PBM configurations."""
+        # handling PBM configurations requires that the pbm snap is installed.
+        if "db_initialised" not in self.charm.app_peer_data:
+            logger.debug("Cannot set PBM configurations, MongoDB has not yet started.")
+            event.defer()
+            return
+
+        snap_cache = snap.SnapCache()
+        pbm_snap = snap_cache["percona-backup-mongodb"]
+
+        if not pbm_snap.present:
+            logger.debug("Cannot set PBM configurations, PBM snap is not yet installed.")
+            event.defer()
+            return
+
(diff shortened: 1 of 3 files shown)


src/charm.py
```diff
@@ -295,13 +296,6 @@ def _on_install(self, event) -> None:
             auth=auth, machine_ip=self._unit_ip(self.unit), config=self.mongodb_config
         )
```
delgod, line 298: I think we should add pbm user before calling `self.client_relations.oversee_users` in `_initialise_replica_set` to make pbm user created during the init.

lib/charms/mongodb/v0/mongodb_backups.py
```diff
@@ -0,0 +1,109 @@
+# Copyright 2022 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+"""In this class, we manage client database relations.
+
+This class creates a user and database for each application relation
+and expose needed information for client connection via fields in
+external relation.
+"""
```
marceloneppel, line 9: It seems that this description is the same as the one from https://github.com/canonical/mongodb-operator/blob/main/lib/charms/mongodb/v0/mongodb_provider.py#L4, and not related to backups.

lib/charms/mongodb/v0/mongodb_backups.py
```diff
@@ -0,0 +1,109 @@
...
+]
+
+PBM_PRIVILEGES = {"resource": {"anyResource": True}, "actions": ["anyAction"]}
+
+
+class MongoDBBackups(Object):
+    """In this class, we manage client database relations."""
+
+    def __init__(self, charm, substrate="k8s"):
+        """Manager of MongoDB client relations."""
```
marceloneppel, line 48: Same as the previous comment about the description.

lib/charms/mongodb/v0/mongodb_backups.py
```diff
@@ -0,0 +1,109 @@
...
+            logger.debug("Cannot set PBM configurations, PBM snap is not yet installed.")
+            event.defer()
+            return
+
+        try:
+            self.create_pbm_user()
+        except PyMongoError as e:
+            self.charm.unit.status = BlockedStatus("config change failed")
+            logger.error("Deferring config_changed since: error=%r", e)
+            event.defer()
```
marceloneppel, line 75: If it defers here, can the charm create the user in the next call to this event?

If so, probably we should change the `BlockedStatus` to some other status as blocked status needs [human/manual intervention](https://juju.is/docs/olm/status-values#heading--workload--charm-status).

I thought on `WaitingStatus` (which I admit is little weird for an error that is being). A better approach could be to set the blocked status only after some retries and then stop deferring or to not defer in this case. Or even separate the errors that could arise in the ones that should block the charm and the ones that can be automatically solved after a defer.

marceloneppel (commented):
Nice work Mia!

lib/charms/mongodb/v0/mongodb_backups.py
```diff
@@ -0,0 +1,109 @@
...
+
+        return MongoDBConfiguration(
+            replset=self.charm.app.name,
+            database="admin",
+            username="pbmuser",
+            password=self.charm.get_secret("app", "pbm_password"),
+            hosts=self.charm.mongodb_config.hosts,
+            roles=["pbm"],
+            tls_external=self.charm.tls.get_tls_files("unit") is not None,
+            tls_internal=self.charm.tls.get_tls_files("unit") is not None,
```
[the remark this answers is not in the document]
MiaAltieri: yes! as always excellent eyes Marcelo

lib/charms/mongodb/v0/mongodb_backups.py
MiaAltieri: I am embarrassed about this 🤦 I will correct this asap

marceloneppel (approved):
Looks great!

lib/charms/mongodb/v0/mongodb_backups.py
```diff
@@ -0,0 +1,87 @@
...
+    """In this class, we manage mongodb backups."""
+
+    def __init__(self, charm, substrate="k8s"):
+        """Manager of MongoDB client relations."""
+        super().__init__(charm, "client-relations")
+        self.charm = charm
+        self.substrate = substrate
+        self.framework.observe(self.charm.on.config_changed, self._on_pbm_config_changed)
+
+    def _on_pbm_config_changed(self, event) -> None:
```
delgod, line 51: Why do we install snap on config_changed? I think we have a special event for such actions "on install"

delgod (changes requested):

lib/charms/mongodb/v0/mongodb_backups.py
```diff
@@ -0,0 +1,87 @@
...
+        # TODO handle PBM configurations set by the user
+        self.charm.unit.status = ActiveStatus()
+
+    @property
+    def _pbm_config(self) -> MongoDBConfiguration:
+        """Construct the config object for pbm user and creates user if necessary."""
+        if not self.charm.get_secret("app", "pbm_password"):
+            self.charm.set_secret("app", "pbm_password", generate_password())
+
+        logger.error("password is %s", self.charm.get_secret("app", "pbm_password"))
```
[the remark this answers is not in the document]
MiaAltieri: oops this was from testing I am embarrased

lib/charms/mongodb/v0/mongodb_backups.py
MiaAltieri: We don't this only checks if it has been installed. (we cannot configure pbm snap if it hasn't been installed)

delgod (approved):

PR #157, opened 2023-02-10
Update charm libraries. Automate charm libraries publishing on release.

taurus-forever:
The local libraries were updated without proper LIBPATCH bump and were not published on charmhub.
The PR updated the local libraries, bump LIBPATCH for all missing libraries and automate charm libraries publishing on release.


MiaAltieri (approved):
looks good. my bad with not updating the mongodb libs, thanks for helping me with this 🙏

delgod (approved):

PR #163, opened 2023-02-17
Restore int basic

MiaAltieri:
Basic Integration tests for restore:
1. Does restore work
2. Does restoring a new cluster to and old cluster work? (GCP & AWS) 

Other additions:
1. Flakey backup tests revealed that the `create-backup` action **should only be run on the leader**, this is because only one backup can be run at a time. Code in `lib` was updated accordingly 
2. Also added a small fix to `src/charm.py` for when a failed restart is retried that the correct ports are opened 
3. Added a fix to `src/machine_helpers.py` for failures associated to mongodb starting too quickly


tests/integration/backup_tests/test_backups.py
```diff
@@ -231,6 +241,143 @@ async def test_multi_backup(ops_test: OpsTest, add_writes_to_db) -> None:
...
+
+    # save old password, since after restoring we will need this password to authenticate.
+    old_password = await ha_helpers.get_password(ops_test, db_app_name)
+
+    # remove the old cluster and make a new cluster with the same name. Backups with PBM are
+    # incompatible if their cluster names differ.
+    await helpers.destory_cluster(ops_test, db_app_name)
+
+    db_charm = await ops_test.build_charm(".")
+    await ops_test.model.deploy(db_charm, num_units=3, application_name=db_app_name)
```
delgod, line 340: Can we please change the `application_name` for the newly created cluster?
It is important to have the possibility to create new clusters from backups.

tests/integration/backup_tests/test_backups.py
```diff
@@ -231,6 +241,143 @@ async def test_multi_backup(ops_test: OpsTest, add_writes_to_db) -> None:
...
+    await asyncio.gather(
+        ops_test.model.wait_for_idle(apps=[db_app_name], status="active"),
+    )
+
+    db_unit = await helpers.get_leader_unit(ops_test)
+    action = await db_unit.run_action("set-password", **{"password": old_password})
+    action = await action.wait()
+    assert action.status == "completed"
```
delgod, line 348: I don't think it is a necessary step for MongoDB. Can we avoid it?

delgod (changes requested):

tests/integration/backup_tests/test_backups.py
MiaAltieri: When restoring a backup from a different named cluster `pbm` gives an error about the replica set name being different and the restore doesn't complete

tests/integration/backup_tests/test_backups.py
```diff
@@ -231,6 +241,143 @@ async def test_multi_backup(ops_test: OpsTest, add_writes_to_db) -> None:
...
+        for attempt in Retrying(stop=stop_after_delay(4), wait=wait_fixed(5)):
+            with attempt:
+                backups = await helpers.count_logical_backups(db_unit)
+                assert backups == prev_backups + 1, "Backup not created."
+    except RetryError:
+        assert backups == prev_backups + 1, "Backup not created."
+
+    # clear writes
+    assert number_writes > 0, "no writes backed up"
+    await ha_helpers.clear_db_writes(ops_test)
```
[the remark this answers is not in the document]
MiaAltieri: makes sense, on it. Thanks for the quality suggestion

tests/integration/backup_tests/test_backups.py
MiaAltieri: I would need to think of how do avoid this. Perhaps there is a way to make a back up but exclude specific things in the admin DB like `password`. 

Right now the restore will restore things like password and replica set name, which means we are required to use the same replica set name and password

tests/integration/backup_tests/test_backups.py
MiaAltieri: Supporting restore to a different cluster name along with correction in `backup-list` will require some additional code and tests. I would prefer to do this in a separate PR

tests/integration/backup_tests/test_backups.py
MiaAltieri: As discussed via MM, we decided this was "okay-ish"

delgod (approved):

tests/integration/backup_tests/test_backups.py
```diff
@@ -231,6 +241,143 @@ async def test_multi_backup(ops_test: OpsTest, add_writes_to_db) -> None:
...
+        )
+
+    # verify all writes are present
+    try:
+        for attempt in Retrying(stop=stop_after_delay(4), wait=wait_fixed(20)):
+            with attempt:
+                number_writes_restored = await ha_helpers.count_writes(ops_test)
+                assert number_writes == number_writes_restored, "writes not correctly restored"
+    except RetryError:
+        assert number_writes == number_writes_restored, "writes not correctly restored"
```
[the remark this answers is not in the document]
shayancanonical: discussed offline: both `continuous_writes` and `write_unwanted_data` write to the same database and collection. therefore this check suffices to ensure that the unwanted writes are not present in the database after the restore is complete

shayancanonical (approved):

PR #166, opened 2023-02-17
correct removal retry

MiaAltieri:
## Problem
1. The PR #159 was misguided. When the retry period was added for `remove_replset_member ` that meant other calling functions would also have a timeout of 10 minutes. Meaning that when a unit scales from 1 unit to 0, instead of waiting 10 minutes, it would wait 20 minutes since `remove_replset_member` gets called **twice** in the removal sequence. 
2. Removing an entire replica set hangs. This is because you cannot step down a primary in a single node replica set and you cannot configure a replica set to have 0 nodes

## Solution
1. Move the long timeout period to `storage_detached`
2. Do not attempt to step down primary or reconfigure to a 0 node replica set if the entire replica set is getting removed.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -146,6 +146,14 @@ class MongodbOperatorCharm(ops.charm.CharmBase):
         If the removing unit is primary also allow it to step down and elect another unit as
         primary while it still has access to its storage.
         """
+        # if we are removing the last replica it will not be able to step down as primary and we
+        # cannot reconfigure the replica set to have 0 members. To prevent retrying for 10 minutes
+        # set this flag to True. please note that planned_units will always be >=1. When planned
+        # units is 1 that means there are no other peers expected.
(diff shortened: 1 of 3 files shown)


delgod (approved):

src/charm.py
```diff
@@ -146,12 +146,27 @@ def _on_mongodb_storage_detaching(self, event: ops.charm.StorageDetachingEvent)
...
-                mongo.remove_replset_member(self._unit_ip(self.unit))
+            # retries over a period of 10 minutes in an attempt to resolve race conditions it is
+            # not possible to defer in storage detached.
+            logger.debug("Removing %s from replica set", self._unit_ip(self.unit))
+            for attempt in Retrying(
+                stop=stop_after_attempt(10),
+                wait=wait_fixed(1),
+                reraise=True,
+            ):
+                with attempt:
```
Mehdi-Bendriss, line 166: Nice

Mehdi-Bendriss (approved):
LGTM!

PR #168, opened 2023-02-17
Cache charmcraft pack container, skip unstable tests except on schedule

carlcsaposs-canonical:
Ported from https://github.com/canonical/mysql-k8s-operator/pull/146

### DIFF
--- a/tests/integration/backup_tests/helpers.py
+++ b/tests/integration/backup_tests/helpers.py
@@ -7,7 +7,7 @@ from pymongo import MongoClient
 from pytest_operator.plugin import OpsTest
 from tenacity import RetryError, Retrying, stop_after_attempt, wait_fixed
 
-import tests.integration.ha_tests.helpers as ha_helpers
+from ..ha_tests import helpers as ha_helpers
 
 S3_APP_NAME = "s3-integrator"
 TIMEOUT = 10 * 60


tests/unit/test_charm.py
```diff
@@ -344,7 +345,7 @@ def test_mongodb_relation_joined_all_replicas_not_ready(self, connection):
     @patch_network_get(private_address="1.1.1.1")
     @patch("ops.framework.EventBase.defer")
     @patch("charm.MongoDBConnection")
-    @patch("lib.charms.mongodb.v0.mongodb.MongoClient")
+    @patch("charms.mongodb.v0.mongodb.MongoClient")
```
MiaAltieri, line 348: Sorry for the noob question, but why this change?

MiaAltieri (commented):
Carl this looks awesome! Thank you so much for your valuable contribution here! There are some changes I am confused about that I am not sure how they are related to the `cache`ing of `charmcraft pack`. Is it possible you have a different formatter and re-formatted some files accidentally? 

This is something I am very new to, so if they are related you can totally ignore this comment. But I would love a small explanation so I can expand my understanding :)

MiaAltieri (approved):
thanks for the explaination. LGTM

delgod (approved):

PR #172, opened 2023-02-23
Add monitor user + roles

MiaAltieri:
## Problem: 
Conversation with Guillermo and Mykola illustrated that a the MongoDB exporter charm couldn't run commands due to authorisation issues 

## Solution:
1. Create a monitor user
2. Create monitor roles

## Testing
Integration tests verify we can run the necessary commands

### DIFF
--- a/tests/integration/helpers.py
+++ b/tests/integration/helpers.py
@@ -26,7 +26,7 @@ def unit_uri(ip_address: str, password, app=APP_NAME) -> str:
     return f"mongodb://operator:" f"{password}@" f"{ip_address}:{PORT}/admin?replicaSet={app}"
 
 
-async def get_password(ops_test: OpsTest, app=APP_NAME) -> str:
+async def get_password(ops_test: OpsTest, app=APP_NAME, username="operator") -> str:
     """Use the charm action to retrieve the password from provided unit.
 
     Returns:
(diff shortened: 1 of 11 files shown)


tests/integration/helpers.py
```diff
@@ -36,9 +36,11 @@ async def get_password(ops_test: OpsTest, app=APP_NAME) -> str:
     unit_name = ops_test.model.applications[app].units[0].name
     unit_id = unit_name.split("/")[1]
 
-    action = await ops_test.model.units.get(f"{app}/{unit_id}").run_action("get-password")
+    action = await ops_test.model.units.get(f"{app}/{unit_id}").run_action(
+        "get-password", **{"username": username}
+    )
     action = await action.wait()
-    return action.results["operator-password"]
+    return action.results[f"{username}-password"]
```
delgod, line 43: lets simplify parsing of results to 3rd-party charms.
```suggestion
    return action.results["password"]
```
can we check what keys other charms have?

delgod (changes requested):

tests/integration/helpers.py
MiaAltieri: Yes will do. 

[This is the info](https://docs.google.com/spreadsheets/d/1mb_cN92iEP0UEJzh7RHZXwJlYd0Yi_y7N6Frw4z3rUs/edit#gid=0) I've collected on how the other charms manage passwords

marceloneppel (approved):

tests/integration/helpers.py
marceloneppel: @delgod, should we align with the other database charm developers to change the other charms as well?

delgod: @marceloneppel yes please

delgod (approved):

PR #180, opened 2023-03-03
Backup list detects new cluster

MiaAltieri:
### Problem

1. `list-backups` will list backups from a previous cluster as `failed` when they should be listed as `finished`
2. failed backups show no error message 

### Context

If a remapping error + some other error occurs then we should **NOT** list the backup as finished we should display the error meessage

### Solution

1. Backups that are failed because of a remapping issue are displayed as finished
2. Include error message in failed backups

### Testing

1. Unit tests verify parsing of error messages
2. Integration tests for cluster migration verify that backups show


marceloneppel (approved):
LGTM!

PR #184, opened 2023-03-10
Drop generate args

MiaAltieri:
### Problem

- Both K8s and VM charm use their own specialised function for generating arguments for mongod. 
- We want K8s to use the same dir structure as the VM charm (excluding the snap dir prefex)

### Solution

Merge the two functions and then include/ignore snap dir based on whether the snap was used

### Testing

Unit tests updated

### TODO
Update K8s charm to use this function correctly


delgod (approved):

Mehdi-Bendriss (approved):
LGTM

delgod (approved):

PR #185, opened 2023-03-10
Upgrade jammy

MiaAltieri:
### Problem

charm needs to be available on jammy

### Solution

Add charm to jammy along with necessary reqs for crypto

### DIFF
--- a/lib/charms/mongodb/v0/helpers.py
+++ b/lib/charms/mongodb/v0/helpers.py
@@ -19,7 +19,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 4
+LIBPATCH = 5
 
 
 # path to store mongodb ketFile
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,15 @@
-tenacity==8.0.1
-pymongo==4.2.0
-ops >= 1.2.0
-jsonschema==4.15.0
-cryptography==38.0.1
-pyOpenSSL==22.0.0
+attrs==22.2.0
+cffi==1.15.1
+importlib-resources==5.10.2
+tenacity==8.1.0
+pymongo==4.3.3
+ops==2.0.0
+jsonschema==4.17.3
+cryptography==38.0.4
+pure-sasl==0.6.2
+pycparser==2.2
+pkgutil-resolve-name==1.3.10
+pyrsistent==0.19.3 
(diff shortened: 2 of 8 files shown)


requirements.txt
```diff
@@ -1,6 +1,15 @@
...
+ops==2.0.0
+jsonschema==4.17.3
+cryptography==38.0.4
+pure-sasl==0.6.2
+pycparser==2.2
+pkgutil-resolve-name==1.3.10
+pyrsistent==0.19.3 
+pyyaml==6.0
+zipp==3.11.0
+pyOpenSSL==22.1.0
```
marcoppenheimer, line 15: **thought/non-blocking:** In Kafka, we're dynamically building the `requirements.txt` through `poetry`, where we run the following on every local run of `tox -e format`:

```
    poetry install
    poetry export -f requirements.txt -o requirements.txt
```

`poetry install` grabs the packages from `pyproject.toml:tool.poetry.dependencies`, where then `poetry` resolves potential version conflicts, and then exports the exact version hashes to `requirements.txt` for 'actual charm deployment' reliability. 

But different to your code here, the packages in `tool.poetry.dependencies` are inclusive ordered comparison clauses (i.e `>=`). One downside of this is that sometimes it'll struggle to resolve conflicts. But one plus-side is that we're always using the most-recent possible combination of packages, and may make it easier to bump versions where necessary (e.g going from `ops==2.0.0` to `ops==2.1.0`) without breaking some other dependency. 

Worth considering, it's been really helpful for us! Definitely not important in this ticket though.

lib/charms/mongodb/v0/helpers.py
```diff
@@ -19,7 +19,7 @@
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 4
+LIBPATCH = 5
```
marcoppenheimer, line 22: **question:** What bumped the versions with no diff?

marcoppenheimer (approved):
Looks good! Left some questions/thoughts but nothing blocking.

requirements.txt
MiaAltieri: Yes I definitely want to add poetry to this project, will need to wait for later though.

lib/charms/mongodb/v0/helpers.py
MiaAltieri: CI: was complaining - must have forgotten to bump in a previous PR 😅

delgod (approved):

PR #187, opened 2023-03-14
set lib patch to correct vs

MiaAltieri:
### Problem

lib patch versions were incorrectly updated (i.e. lib patch was too high) [release CI reported](https://github.com/canonical/mongodb-operator/actions/runs/4414747786/jobs/7738041500):
```
Library charms.mongodb.v0.mongodb_vm_legacy_provider has a wrong LIBPATCH number, it's too high and needs to be consecutive, Charmhub highest version is 0.2.
Library charms.mongodb.v0.mongodb_backups has a wrong LIBPATCH number, it's too high and needs to be consecutive, Charmhub highest version is 0.1.
Library charms.mongodb.v0.helpers has a wrong LIBPATCH number, it's too high and needs to be consecutive, Charmhub highest version is 0.3.
```

### Solution

Put the lib patch to a correct number

### DIFF
--- a/lib/charms/mongodb/v0/helpers.py
+++ b/lib/charms/mongodb/v0/helpers.py
@@ -19,7 +19,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 5
+LIBPATCH = 4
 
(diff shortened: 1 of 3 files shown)


WRFitch (approved):

paulomach (approved):
amazing

PR #188, opened 2023-03-15
Update pbm uri

MiaAltieri:
### Problem

1. Uri in snap is now set with `pbm-uri` not as `uri`
2. When hosts are updated the URI that `pbm` should be updated
3. When password is rotated the URI that pbm uses should be updated

### Solution
1. Replace setting the `uri` with setting the `pbm-uri`
2. Update the hosts in the pbm uri whenever they get changed
3. When password is rotated the URI that pbm is updated

### Future PR
Will do something similar for TLS


marceloneppel (approved):
Looks good!

delgod (approved):

MiaAltieri:
Closing this PR as it needs more consideration before merging and it is out of scope for this pulse

PR #189, opened 2023-03-21
update backup & HA tests

MiaAltieri:
### Problem

1. keys for AWS changed so tests no longer passed
2. HA test check for all db processes down fails

### Solution

1. update backup configs in backup tests for the new AWS keys/configs 
2. update HA test check for all db processes down to ignore the mongodb_exporter process

PR #192, opened 2023-03-22
Add metrics endpoint

MiaAltieri:
### Problem

We need to expose metric endpoint for COS

### Solution

- Provide the relation
- add necessary libs
- use node exporter and mongodb exporter in the charm

### Testing

Verified the endpoint was up with MongoDB related metrics:
```
juju deploy ./*charm
curl http://<unit-ip>:9216/metrics | grep mongo
curl http://<unit-ip>:9100/metrics 
```

### Future PRs

Future PRs will include:
- unit tests
- int tests
- updates for cos sidecar log pattern
- grafana
- update uri for monitoring uri when passwords change or when there is a network cut and the ip changes

### Files to Ignore
`lib/charms/observability_libs/v0/juju_topology.py`
`lib/charms/prometheus_k8s/v0/prometheus_scrape.py`

### Failing tests
- backup tests will keep failing until #189 is merged
- failures in HA tests will keep failing until #189 is merged


MiaAltieri:
Closing until next course of action is decided. This is regarding the fact that cos has now changed to the sidecar pattern

zmraul (changes requested):
This has the dashboards, but is missing the Grafana integration no?
Needs `charmcraft fetch-lib charms.grafana_k8s.v0.grafana_dashboard` and on `charm.py`:
```python
. . .
self.metrics_endpoints = ....
self.grafana_dashboards = GrafanaDashboardProvider(self)
. . .
```

MiaAltieri:
@zmraul oops you're right I was too hasty in adding the graphana dashboards - I will remove and save for the graphana metric

delgod (approved):

PR #195, opened 2023-03-24
specify licenses

MiaAltieri:
### Problem

License should be stated and shared

### Solution

Add Licenses

### Other
libraries updated in #194


taurus-forever (approved):
LGTM from my side, but @delgod doesn't want to see licenses files hardcoded in our Git repo, they need to be copied from artifacts we are installing (apt packages).


P.S. build process is not happy here too.

MiaAltieri:
closing PR until decision from legal is made

PR #196, opened 2023-03-27
Metrics int tests

MiaAltieri:
### Problem

1. No integration tests for metrics integration 
2. The call to update monitor connection in relation_changed events was being skipped due to the if statement before it

### Solution

1. fix for code to correctly update connection uri when the password is changed on non-leader units
2. sanity check that endpoints are running test
3. endpoints still function correctly after the monitor user password changes test
4. endpoints still function correctly after a network cut

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -81,6 +81,7 @@ jobs:
           - legacy-integration
           - tls-integration
           - backup-integration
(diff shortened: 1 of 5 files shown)


zmraul (approved):
Nice. Very good idea testing the endpoints after a password rotation as well!

PR #206, opened 2023-04-17
backup tests need longer wait for sync

MiaAltieri:
## Issue
backup tests fail due to the timeout for resync

## Context
resync is successful - but it takes longer than the time we provide in the tests

## Solution
Add a longer timeout period after resync


MiaAltieri:
this was a testing PR to enter CI - closing as no longer in use.

Note that the tests eventually pass - but the CI fails due to upterm

PR #217, opened 2023-05-16
enable logging for mongod for k8s

MiaAltieri:
## Issue
By default the MongoDB K8s charm has its logs printed as the output of running the container and are not saved on the container. This means that the logs cannot be ported over to the cos bundle.

## Solution
To enable logging we must change the start up arguments, these are generated within the shared library code and therefore the change library code change must be made in the VM charm first

### DIFF
--- a/lib/charms/mongodb/v0/helpers.py
+++ b/lib/charms/mongodb/v0/helpers.py
@@ -36,6 +36,7 @@ MONGODB_SNAP_DATA_DIR = "/var/snap/charmed-mongodb/current"
 
 DATA_DIR = "/var/lib/mongodb"
 CONF_DIR = "/etc/mongod"
+MONGODB_LOG_FILENAME = "mongodb.log"
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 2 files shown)


delgod (approved):

lib/charms/mongodb/v0/helpers.py
```diff
@@ -82,16 +82,19 @@ def get_mongod_args(
     Returns:
         A string representing the command used to start MongoDB.
     """
-    #
     full_data_dir = f"{MONGODB_COMMON_DIR}{DATA_DIR}" if snap_install else DATA_DIR
     full_conf_dir = f"{MONGODB_SNAP_DATA_DIR}{CONF_DIR}" if snap_install else CONF_DIR
+    # in k8s the default logging options that are used for the vm charm are ignored and logs are
+    # the output of the container. To enable logging to a file it must be set explicitly
+    logging_options = "" if snap_install else f"--logpath={full_data_dir}/mongodb.log"
```
dmitry-ratushnyy, line 89: just at small note: If we are using constants to build `full_data_dir` MONGODB_COMMON_DIR and CONF_DIR I think it make sense to use a constant for "mongodb.log", something like "MONGODB_LOG_FILE_NAME" or similar

dmitry-ratushnyy (approved):

lib/charms/mongodb/v0/helpers.py
MiaAltieri: Good suggestion thank you @dmitry-ratushnyy

PR #224, opened 2023-06-11
Fix restart of pbm-agent service (backup)

dmitry-ratushnyy:
## Issue

Sometimes pbm (backup) service is failing to restart with the following error
`systemd[1]: snap.charmed-mongodb.pbm-agent.service: Start request repeated too quickly`

## Solution

1. Replace a call to `snap.restart(service_name)`  with `snap.stop(service_name) \ snap.start(service_name)`
2. Added additional logging


src/charm.py
```diff
@@ -5,6 +5,7 @@
 import json
 import logging
 import subprocess
+import time
 from subprocess import check_call
```
delgod, line 9: Could we use one style with subprocess: or `subprocess.check_call` or `from subprocess import check_call, check_output`

delgod (changes requested):

src/charm.py
dmitry-ratushnyy: sure, I will change it to `from subprocess import check_call, check_output`

delgod (approved):

PR #237, opened 2023-08-18
Bump PyYaml and ops libraries verion

dmitry-ratushnyy:
## Issue
Building of charm is failed with juju 3.1.5 which is installed by default with snap

PyYaml pinned to version 6.0.0 causes issues with building charm due to incompatibility with cython 3.

## Solution
Bump PyYaml and ops library versions

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -12,7 +12,7 @@ pycparser==2.2
 pkgutil-resolve-name==1.3.10
 pydantic==1.10.7
 pyrsistent==0.19.3 
-pyyaml==6.0
+pyyaml>6.0
 zipp==3.11.0
 pyOpenSSL==22.1.0
 typing-extensions==4.5.0 
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,7 +4,7 @@ cosl==0.0.5
 importlib-resources==5.10.2
 tenacity==8.1.0
 pymongo==4.3.3
-ops==2.0.0
+ops==2.4.1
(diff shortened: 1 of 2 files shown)


MiaAltieri:
I think there might be some issues still with compatibility:
```
The conflict is caused by:
    The user requested pyyaml>6.0
    juju 2.9.38.1 depends on pyyaml<=6.0 and >=5.1.2

To fix this you could try to:
1. loosen the range of package versions you've specified
2. remove package versions to allow pip attempt to solve the dependency conflict
```

MiaAltieri (commented):
The conflict is caused by:
    The user requested pyyaml>6.0
    juju 2.9.38.1 depends on pyyaml<=6.0 and >=5.1.2

To fix this you could try to:
1. loosen the range of package versions you've specified
2. remove package versions to allow pip attempt to solve the dependency conflict

delgod:
should be solved via poetry

MiaAltieri (changes requested):
I believe we have the juju version hardcoded in [the tests to 2.9.38.1](https://github.com/canonical/mongodb-operator/blob/82d0f37a0f247ddd1272d984e3c5223aa6698972/tox.ini#L103) so I think these changes will introduce conflicts with the version of juju we are testing on, it might be necessary to bump that version

dmitry-ratushnyy:
Closing this PR in favour of https://github.com/canonical/mongodb-operator/pull/240

PR #242, opened 2023-09-01
[DPE-2442] Add juju secrets support

dmitry-ratushnyy:
## Issue
Add support of juju secrets to the charm

## Related issues
[DPE-2441](https://warthogs.atlassian.net/browse/DPE-2441)
[DPE-2442](https://warthogs.atlassian.net/browse/DPE-2442)
[DPE-2443](https://warthogs.atlassian.net/browse/DPE-2443)

[DPE-2441]: https://warthogs.atlassian.net/browse/DPE-2441?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DPE-2442]: https://warthogs.atlassian.net/browse/DPE-2442?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DPE-2443]: https://warthogs.atlassian.net/browse/DPE-2443?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


requirements.txt
```diff
@@ -4,15 +4,15 @@ cosl==0.0.5
 importlib-resources==5.10.2
 tenacity==8.1.0
 pymongo==4.3.3
-ops==2.0.0
+ops==2.4.0
```
MiaAltieri, line 7:
```suggestion
ops==2.4.1
```

requirements.txt
```diff
@@ -4,15 +4,15 @@ cosl==0.0.5
...
 jsonschema==4.17.3
 cryptography==38.0.4
 pure-sasl==0.6.2
 pycparser==2.2
 pkgutil-resolve-name==1.3.10
 pydantic==1.10.7
 pyrsistent==0.19.3 
-pyyaml==6.0
+pyyaml > 6.0
```
MiaAltieri, line 15:
```suggestion
pyyaml==6.0.1
```

tox.ini
```diff
@@ -72,8 +72,9 @@ pass_env =
     CI_PACKED_CHARMS
```
MiaAltieri, line 74: fun fact, if you want to work with juju `2.9` with the pyyaml conflicts, you can use `2.9.44.0`

src/charm.py
```diff
@@ -930,21 +1000,29 @@ def _unit_ip(self, unit: Unit) -> str:
...
+        elif scope == APP_SCOPE:
             return self.app_peer_data.get(key, None)
         else:
             raise RuntimeError("Unknown secret scope.")
 
-    def set_secret(self, scope: str, key: str, value: Optional[str]) -> None:
+    def set_secret(self, scope: str, key: str, value: Optional[str]) -> Optional[str]:
         """Set secret in the secret storage."""
-        if scope == "unit":
+        if self._juju_has_secrets:
```
MiaAltieri, line 1015: thank you!

src/charm.py
```diff
@@ -930,21 +1000,29 @@ def _unit_ip(self, unit: Unit) -> str:
...
             return self.unit_peer_data.get(key, None)
-        elif scope == "app":
+        elif scope == APP_SCOPE:
             return self.app_peer_data.get(key, None)
         else:
             raise RuntimeError("Unknown secret scope.")
 
-    def set_secret(self, scope: str, key: str, value: Optional[str]) -> None:
+    def set_secret(self, scope: str, key: str, value: Optional[str]) -> Optional[str]:
         """Set secret in the secret storage."""
```
MiaAltieri, line 1014:
```suggestion
        """Set secret in the secret storage.
        
        Juju versions > 3.0 use `juju secrets`, this function first checks which secret store is being used before 
        setting the secret. 
        """
```

src/charm.py
```diff
@@ -979,6 +1057,118 @@ def auth_enabled(self) -> bool:
...
+            return self.unit
+
+    def _peer_data(self, scope: Scopes):
+        if not self._peers:
+            return {}.setdefault(scope, {})
+        scope_obj = self._scope_obj(scope)
+        return self._peers.data[scope_obj]
+
+    def _juju_secret_set(self, scope: Scopes, key: str, value: str) -> str:
+        """Helper function setting Juju secret."""
```
MiaAltieri, line 1073:
```suggestion
        """Helper function setting Juju secret in Juju versions >3.0."""
```

src/charm.py
```diff
@@ -979,6 +1057,118 @@ def auth_enabled(self) -> bool:
...
+        peer_data = self._peer_data(scope)
+        self._juju_secrets_get(scope)
+
+        secret = self.secrets[scope].get(Config.Secrets.SECRET_LABEL)
+
+        # It's not the first secret for the scope, we can re-use the existing one
+        # that was fetched in the previous call
```
MiaAltieri, line 1080: good comment, can you add a few words on why this is important, Maybe something like:
```
        # It's not the first secret for the scope, we can re-use the existing one
        # that was fetched in the previous call, as fetching secrets from juju is
        # slow
``` 
(not sure what the reason is, but that is a guess)

src/charm.py
```diff
@@ -979,6 +1057,118 @@ def auth_enabled(self) -> bool:
...
+
+            self.secrets[scope][Config.Secrets.SECRET_LABEL] = secret
+            self.secrets[scope][Config.Secrets.SECRET_CACHE_LABEL] = {key: value}
+            logging.debug(f"Secret {scope}:{key} published (as first). ID: {secret.id}")
+            peer_data.update({Config.Secrets.SECRET_INTERNAL_LABEL: secret.id})
+
+        return self.secrets[scope][Config.Secrets.SECRET_LABEL].id
+
+    def _juju_secrets_get(self, scope: Scopes) -> Optional[bool]:
+        """Helper function to get Juju secret."""
```
MiaAltieri, line 1113:
```suggestion
        """Helper function to retrieve all Juju secrets.
        
        Function returns a boolean based on X, but infact retrieves Juju secrets and stores them in the charm. (include expert knowledge)
        """
```

src/charm.py
```diff
@@ -979,6 +1057,118 @@ def auth_enabled(self) -> bool:
...
+        peer_data = self._peer_data(scope)
+
+        if not peer_data.get(Config.Secrets.SECRET_INTERNAL_LABEL):
```
MiaAltieri, line 1116: noob question, why is it needed to check for internal secrets in peer data? is it because it means that the charm hasn't been set up, i.e. mongod user etc? (maybe adding a comment here will be useful for maintainability)

src/charm.py
MiaAltieri: Since this function has a lot of expert knowledge in it, i.e. the function makes has a specific design that relies on your expert knowledge, it would be nice to add a sentence or two to the docstring so that it is clear why the function works in this way

src/charm.py
```diff
@@ -979,6 +1057,118 @@ def auth_enabled(self) -> bool:
...
+                raise SecretNotAddedError(f"Couldn't set secret {scope}:{key}")
+
+            self.secrets[scope][Config.Secrets.SECRET_LABEL] = secret
+            self.secrets[scope][Config.Secrets.SECRET_CACHE_LABEL] = {key: value}
+            logging.debug(f"Secret {scope}:{key} published (as first). ID: {secret.id}")
+            peer_data.update({Config.Secrets.SECRET_INTERNAL_LABEL: secret.id})
+
+        return self.secrets[scope][Config.Secrets.SECRET_LABEL].id
+
+    def _juju_secrets_get(self, scope: Scopes) -> Optional[bool]:
```
MiaAltieri, line 1112: seems like some functions rely on the result of `_juju_secrets_get `, should it really return an optional bool?

src/charm.py
```diff
@@ -979,6 +1057,118 @@ def auth_enabled(self) -> bool:
...
+
+            # We retrieve and cache actual secret data for the lifetime of the event scope
+            self.secrets[scope][Config.Secrets.SECRET_CACHE_LABEL] = secret.get_content()
+
+        if self.secrets[scope].get(Config.Secrets.SECRET_CACHE_LABEL):
+            return True
+        return False
+
+    def _juju_secret_get(self, scope: Scopes, key: str) -> Optional[str]:
+        if not key:
```
MiaAltieri, line 1142: nit docstring

tests/integration/helpers.py
```diff
@@ -77,3 +77,30 @@ async def find_unit(ops_test: OpsTest, leader: bool, app=APP_NAME) -> ops.model.
...
+async def get_leader_id(ops_test: OpsTest) -> int:
+    """Returns the unit number of the juju leader unit."""
+    leader_unit_id = 0
+    for unit in ops_test.model.applications[APP_NAME].units:
+        if await unit.is_leader_from_status():
+            return leader_unit_id
+
+        leader_unit_id += 1
+
+    return leader_unit_id
```
MiaAltieri, line 91: this function will work as you expect. IIRC ` ops_test.model.applications[APP_NAME].units` does not always return the units in order, ie I think it can return them in `unit/2`, `unit/0`, `unit/1`. (also in VM if you have 3 units, remove unit `unit/0` and then do `juju add-unit` you will then have `unit/1`, `unit/2`, `unit/3` 

instead I would parse the name with something like
```
        if await unit.is_leader_from_status():
            # parse unit id from unit name
            return = int(unit.name.split("/")[1])

File: tests/integration/test_charm.py [hunk 13]
```diff
@@ -176,11 +181,98 @@ async def test_monitor_user(ops_test: OpsTest) -> None:
...
+        password2 = await get_password(ops_test, username="monitor")
+        assert password2 == password
+
+
+@pytest.mark.usefixtures("only_with_juju_secrets")
+async def test_reset_and_get_password_secret_same_as_cli(ops_test: OpsTest) -> None:
+    """Test verifies that we can set and retrieve the correct password using Juju 3.x secrets."""
+    new_password = str(uuid4())
+
+    # Re=setting existing password
```
MiaAltieri, line 217:
```suggestion
    # Resetting existing password
```

tests/integration/test_charm.py
```diff
@@ -176,11 +181,98 @@ async def test_monitor_user(ops_test: OpsTest) -> None:
...
+
+
+@pytest.mark.usefixtures("only_with_juju_secrets")
+async def test_no_password_change_on_invalid_password(ops_test: OpsTest) -> None:
+    """Test that in general, there is no change when password validation fails."""
+    leader_id = await get_leader_id(ops_test)
+    password1 = await get_password(ops_test, username="monitor")
+
+    # The password has to be minimum 3 characters
+    await set_password(ops_test, unit_id=leader_id, username="monitor", password="ca" * 1000000)
```
MiaAltieri, line 271: `* 1000000)`?

MiaAltieri (commented):
mostly nits :)

tests/integration/test_charm.py
dmitry-ratushnyy: This test was taken from K8S charm as is :) 

And actually helped with setting max password length (otherwise there will be a crash in setting params for snap in connect_to_exporter)

lib/charms/mongodb/v0/mongodb_tls.py
```diff
@@ -36,49 +43,54 @@
```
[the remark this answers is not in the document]
dmitry-ratushnyy: This was copied from K8s charm, but let me bump lib version, thank you

lib/charms/mongodb/v0/mongodb_tls.py
```diff
@@ -182,46 +190,47 @@ def _on_certificate_available(self, event: CertificateAvailableEvent) -> None:
             event.defer()
             return
 
-        if renewal and self.substrate == "k8s":
-            self.charm.unit.get_container("mongod").stop("mongod")
```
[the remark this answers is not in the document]
dmitry-ratushnyy: This was copied from k8s charm as-is (with the lates changes)

src/charm.py
dmitry-ratushnyy: no, it should not, and function naming is a bit misleading in my opinion, but for this PR let's keep it as-is because it was copied from k8s charm in that way and will require changes in both charms

src/charm.py
dmitry-ratushnyy: Done, thank you

tests/integration/helpers.py
dmitry-ratushnyy: Done, thank you

tests/integration/ha_tests/helpers.py
```diff
@@ -487,8 +488,11 @@ async def kill_unit_process(ops_test: OpsTest, unit_name: str, kill_code: str):
     if len(ops_test.model.applications[app].units) < 2:
         await ops_test.model.applications[app].add_unit(count=1)
         await ops_test.model.wait_for_idle(apps=[app], status="active", timeout=1000)
+    if version("juju") < "3":
+        kill_cmd = f"run --unit {unit_name} -- pkill --signal {kill_code} -f {DB_PROCESS}"
+    else:
+        kill_cmd = f"exec --unit {unit_name} -- pkill --signal {kill_code} -f {DB_PROCESS}"
```
delgod, line 494: Why not use `exec` always?

src/charm.py
dmitry-ratushnyy: This code is actually from k8s charm with implementation of secrets, but I will update the comment, thank you 😄

src/charm.py
dmitry-ratushnyy: To be fair - it is @juditnovak and @delgod knowledge, not mine 😄 

And it is copied from k8s charm as-is. I would prefer to keep them same for now. 
Return value does not make much sense and the function can be refactored, but not in this PR

src/charm.py
dmitry-ratushnyy: @juditnovak ^^ 😄

tests/unit/test_charm.py
```diff
@@ -654,11 +667,10 @@ def test_set_password_provided(self, pbm_status, connection):
         action_event = mock.Mock()
         action_event.params = {"password": "canonical123"}
         self.harness.charm._on_set_password(action_event)
-        new_password = self.harness.charm.app_peer_data["operator-password"]
+        new_password = self.harness.charm.get_secret("app", "operator-password")
 
         # verify app data is updated and results are reported to user
         self.assertEqual("canonical123", new_password)
-        action_event.set_results.assert_called_with({"password": "canonical123"})
```
delgod, line 661: just Q: why we remove this line?

src/charm.py
delgod: Peer data bag stores the Secret ID inside in `SECRET_INTERNAL_LABEL` field.
The current code cannot retrieve the Secret object if we don't have the Secret ID (line 1126 below).
I think this part is going to be revoked in two weeks.

src/charm.py
delgod: the function name can be better. in fact, it is just caching secret content into the internal dictionary.

delgod (changes requested):

src/charm.py
juditnovak: How about something like this?

```
This function is responsible for direct communication with the Juju Secret strore to retrieve the Mono Charm's single, unique Secret object's metadata, and --on success-- its contents.

In parallel with retrieving secret information, it's immediately locally cached, making sure that we have the snapshot of the secret for the lifetime of the event (that's being processed) without additional fetch requests to the Juju Secret Store.

(Note: metadata, i.e. the Secret object itself is cached as it may be necessary for later operations, like updating contents.)

The function is returning a boolean that reflects success or failure of the above.
```

This logic should be described somewhere else in a clear, conscize way. But as for now, as a firefighting it could provide "quick help" :-)

src/charm.py
juditnovak: We store the ID of the the secret object in the peer data.
If the ID is not there, we can't fetch the secret (yet).

So technically this means that no secret was set up for Mongo yet.
If I'm correct it's the username/password that comes first, basically initiating the secret.

However not that this will change in a week or two -- we'll be using Secret labels instead of IDs in Mongo.

tests/integration/tls_tests/helpers.py
```diff
@@ -70,7 +74,11 @@ async def check_tls(ops_test: OpsTest, unit: ops.model.Unit, enabled: bool) -> b
 
 async def time_file_created(ops_test: OpsTest, unit_name: str, path: str) -> int:
     """Returns the unix timestamp of when a file was created on a specified unit."""
-    time_cmd = f"run --unit {unit_name} --  ls -l --time-style=full-iso {path} "
+    if version("juju") < "3":
```
juditnovak, line 77: For consistency, I would rather suggest to use `JujuVersion.from_environ().has_secrets`, that providing an overall mock for the test execution. (i.e. in control of Juju version decision on a global scope.)

tox.ini
```diff
@@ -57,6 +57,8 @@ description = Run unit tests
 deps =
     pytest
     requests
+    pytest-mock
+    juju==3.2.0.1
```
juditnovak, line 61: Shouldn't it be a separate PR to move the version, and then to add secrets?

It allows to see all potential "surprise" introduced with the version change -- independent from code changes.

juditnovak (changes requested):
The tests should be consistent relying on the `juju_has_secrets` autoreused fixture.

tests/integration/backup_tests/test_backups.py
```diff
@@ -150,6 +150,7 @@ async def test_create_and_list_backups(ops_test: OpsTest) -> None:
         assert backups == 1, "Backup not created."
 
 
+@pytest.mark.skip("skip")
```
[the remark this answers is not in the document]
dmitry-ratushnyy: fixed, thank you

tests/unit/test_charm.py
dmitry-ratushnyy: restored, thank you

tox.ini
dmitry-ratushnyy: I prefer to have it in the same pr

tests/integration/ha_tests/helpers.py
dmitry-ratushnyy: changed, thank you

tox.ini
MiaAltieri: I agree with Judit

MiaAltieri (approved):
LGTM!

delgod (approved):

src/charm.py
```diff
@@ -223,6 +244,10 @@ def db_initialised(self, value):
                 f"'db_initialised' must be a boolean value. Proivded: {value} is of type {type(value)}"
             )
 
+    @property
+    def _juju_has_secrets(self) -> bool:
+        return JujuVersion.from_environ().has_secrets
```
juditnovak, line 249: I think this is another "expensive" juju call each time. 
I used the internal `_jujuversion` instance attribute to make sure not to issue the command every single time we refer to the version of Juju. I'd suggest to put that back.

src/charm.py
```diff
@@ -520,6 +547,11 @@ def _on_set_password(self, event: ActionEvent) -> None:
             return
 
         new_password = event.params.get(Config.Actions.PASSWORD_PARAM_NAME, generate_password())
+        if len(new_password) > Config.Secrets.MAX_PASSWORD_LENGTH:
+            event.fail(
+                f"Password cannot be longer than {Config.Secrets.MAX_PASSWORD_LENGTH} characters."
+            )
+            return
```
juditnovak, line 554: Is there a lower limit, too? Hmm, on that note: if we pass an empty string value for a Juju Secret, the operation will fail.

src/charm.py
```diff
@@ -532,15 +564,43 @@ def _on_set_password(self, event: ActionEvent) -> None:
...
 
         if username == BackupUser.get_username():
             self._connect_pbm_agent()
 
         if username == MonitorUser.get_username():
             self._connect_mongodb_exporter()
 
-        event.set_results({Config.Actions.PASSWORD_PARAM_NAME: new_password})
+        event.set_results(
+            {Config.Actions.PASSWORD_PARAM_NAME: new_password, "secret-id": secret_id}
```
juditnovak, line 578: Semi-nitpick: we are actually using the secret URI. I'm wondering if we may wanna change it for the event contract, for clarity? [It's not a very important, rather almost a hypotetical question :-) ]

src/charm.py
```diff
@@ -1006,9 +1091,140 @@ def restart_backup_service(self) -> None:
...
+            return False
+
+        regex = re.compile(".*[^/][/:]")
+
+        pure_id1 = regex.sub("", secret_id1)
+        pure_id2 = regex.sub("", secret_id2)
+
+        if pure_id1 and pure_id2:
+            return pure_id1 == pure_id2
+        return False
```
juditnovak, line 1124: Instead of this function now we should be able to use `secret1.unique_id == secret2.unique_id`

delgod (dismissed):

src/charm.py
dmitry-ratushnyy: I prefer for now to keep the code same in k8s and vm charms and do this change later

src/charm.py
dmitry-ratushnyy: Lower limit will be introduced later for both charms. The upper limit set to avoid crash. 😄

src/charm.py
dmitry-ratushnyy: Not sure how this operation is "expensive" to be honest 😄

juditnovak: OK, you're right, [apparently](https://ops.readthedocs.io/en/latest/#ops.JujuVersion.from_environ) it's just checking an env var. 

I thought it was a more elaborate call (like the ones behind secret-related operations).

src/charm.py
```diff
@@ -532,15 +564,43 @@ def _on_set_password(self, event: ActionEvent) -> None:
...
-        event.set_results({Config.Actions.PASSWORD_PARAM_NAME: new_password})
+        event.set_results(
+            {Config.Actions.PASSWORD_PARAM_NAME: new_password, "secret-id": secret_id}
+        )
+
+    def _on_secret_remove(self, event: SecretRemoveEvent):
+        # We are keeping this function empty on purpose until the issue with secrets
+        # is not fixed. The issue is: https://bugs.launchpad.net/juju/+bug/2023364
+        logging.error(
+            f"_on_secret_remove: Secret {event._id} seems to have no observers, could be removed"
```
MiaAltieri, line 585:
```suggestion
            f"_on_secret_remove: Secret {event._id} seems to have no observers, could not be removed"
```

src/charm.py
```diff
@@ -532,15 +564,43 @@ def _on_set_password(self, event: ActionEvent) -> None:
...
+        elif self._compare_secret_ids(
+            event.secret.id, self.unit_peer_data.get(Config.Secrets.SECRET_INTERNAL_LABEL)
+        ):
+            scope = UNIT_SCOPE
+        else:
+            logging.debug("Secret %s changed, but it's unknown", event.secret.id)
+            return
+        logging.debug("Secret %s for scope %s changed, refreshing", event.secret.id, scope)
+        self._update_juju_secrets_cache(scope)
+        self._connect_mongodb_exporter()
```
MiaAltieri, line 602:
```suggestion

        # changed secrets means that the URIs used for PBM and mongodb_exporter are now out of date
        self._connect_mongodb_exporter()
```

src/charm.py
MiaAltieri: I like Judits docstring here, I think it explains the complexities that are going on. Unless you feel strongly about removing it.

tests/integration/test_charm.py
```diff
@@ -176,13 +180,96 @@ async def test_monitor_user(ops_test: OpsTest) -> None:
...
+    _, stdout, _ = await ops_test.juju(*complete_command.split())
+    data = json.loads(stdout)
+
+    assert password == new_password
+    assert data[secret_id]["content"]["Data"]["monitor-password"] == password
+
+
+@pytest.mark.usefixtures("only_without_juju_secrets")
+async def test_reset_and_get_password_no_secret(ops_test: OpsTest, mocker) -> None:
+    """Test verifies that we can set and retrieve the correct password using Juju 2.x."""
```
MiaAltieri, line 234: in tox this test is set up with `3.2.0.1`, I don't think this test runs on CI with `2.x`. i.e. this test never runs on CI, this is okay with me for now. But maybe in the future we can run it on CI with juju `2.x`?

MiaAltieri (approved):
Awesome work Dmitry :)

tests/integration/test_charm.py
dmitry-ratushnyy: We not supporting Juju 2.x anymore

src/charm.py
juditnovak: @MiaAltieri no, the message is correct as it is :-) 

The point is this. When the `secret-remove` event is raised, it's a "message" to the secret owner, that it could safely get rid of the secret completely. Currently we're ignoring this event (just logging that it happened). But the meaning of the event is that the secret is ready to be removed by the owner (whom at this point may decide what it wants to do).

NOTE that the behavior was broken not that long ago -- so safer not to take any action on this event so far I guess :-)

src/charm.py
```diff
@@ -912,21 +985,33 @@ def _unit_ip(self, unit: Unit) -> str:
...
+        if scope == UNIT_SCOPE:
             if not value:
                 del self.unit_peer_data[key]
                 return
             self.unit_peer_data.update({key: str(value)})
-        elif scope == "app":
+        elif scope == APP_SCOPE:
             if not value:
                 del self.app_peer_data[key]
                 return
```
juditnovak, line 1017: This is just cosmetics, but why not:

```
        if not value:
            del self._peer_data[key]
            return
```

juditnovak: (i.e. no `if` is needed)

src/charm.py
```diff
@@ -1006,9 +1091,140 @@ def restart_backup_service(self) -> None:
...
+        if not secret:
+            logging.error(f"Secret {scope}:{key} wasn't deleted: no secrets are available")
+            return
+
+        secret_cache = self.secrets[scope].get(Config.Secrets.SECRET_CACHE_LABEL)
+        if not secret_cache or key not in secret_cache:
+            logging.error(f"No secret {scope}:{key}")
+            return
+
+        secret_cache[key] = Config.Secrets.SECRET_DELETED_LABEL
```
juditnovak, line 1223: I thought we agreed that deletion labels aren't needed anymore...

Do we keep them still?

juditnovak: I mean I would really like to believe that https://bugs.launchpad.net/juju/+bug/2028094 is already in `stable`...

src/config.py
```diff
@@ -40,9 +43,38 @@ class Monitoring:
...
+        Scopes = Literal[APP_SCOPE, UNIT_SCOPE]
+
+    class Secrets:
+        """Secrets related constants."""
+
+        SECRET_LABEL = "secret"
+        SECRET_CACHE_LABEL = "cache"
+        SECRET_KEYFILE_NAME = "keyfile"
+        SECRET_INTERNAL_LABEL = "internal-secret"
+        SECRET_DELETED_LABEL = "None"
```
juditnovak, line 79: We should be able to drop this, see also comments above

tests/conftest.py
```diff
@@ -0,0 +1,44 @@
...
+    if version("juju") < "3":
+        mocker.patch.object(
+            JujuVersion, "has_secrets", new_callable=PropertyMock
+        ).return_value = False
+        return False
+    else:
+        mocker.patch.object(
+            JujuVersion, "has_secrets", new_callable=PropertyMock
+        ).return_value = True
+        return True
```
juditnovak, line 27: FYI, the above function [got enhanced in Opensearch](https://github.com/canonical/opensearch-operator/blob/main/tests/conftest.py#L13) to look at the `LIBJUJU_VERSION_SPECIFIER`

Thus no libjuju installation is needed for unittests to use the fixture

tests/integration/test_charm.py
juditnovak: ...but as long as we did...

I was running these tests on both Juju2 and Juju3 pipelines. 
The ones with the `only_with_juju_secrets` fixture were ONLY executed on Juju3 and skipped on Juju2 while for the `only_without_juju_secrets` it was the opposite.

Thus this particular test was only run on Juju2.

NOW, if there's literally no more Juju3 in pipelines, then all marked tests (`only_without_juju_secrets`) can be removed, together with both fixtures themselves.

tests/integration/test_charm.py
juditnovak: Yes, this was my purposefully invalid password, passing the limit :-)  * 1000000 :-)

juditnovak (changes requested):

tests/conftest.py
dmitry-ratushnyy: Because we are not supporting Juju 2.9 this will be removed in the separate PR

src/charm.py
dmitry-ratushnyy: For now I prefer to keep the code for both charms similar and do the required updates later in a separate PR

src/charm.py
dmitry-ratushnyy: In which version of ops this is available?

src/charm.py
dmitry-ratushnyy: not sure what you mean by `if` is needed

dmitry-ratushnyy: Also I prefer not change code related to old secrets at this PR

juditnovak (approved):

juditnovak (approved):
Approving under the condition to pls address outstanding issues in the following tickets:

https://warthogs.atlassian.net/browse/DPE-2609
https://warthogs.atlassian.net/browse/DPE-2610
https://warthogs.atlassian.net/browse/DPE-2611
https://warthogs.atlassian.net/browse/DPE-2612

MiaAltieri (approved):

juditnovak (approved):

delgod (approved):

PR #245, opened 2023-09-12
Update GH runners: use juju 3.1.5 for CI

dmitry-ratushnyy:
## Issue
We need to test Juju secrets integration on CI and to test the charm with latest Juju version which is installed with snap by default

## Solution
Configure Github runners to use Juju 3.1.5


MiaAltieri (commented):
Noob question, is it necessary to change the tox files to also use Juju 3.1.5? Because right now the file specifies `juju==2.9.44.0` for each test

delgod (approved):

PR #250, opened 2023-09-18
Fix HA tests

dmitry-ratushnyy:
## Issue
Some HA tests are broken, but because they marked unstable the issue does not happen on CI

## Solution
FIx HA tests

### DIFF
--- a/tests/integration/ha_tests/test_ha.py
+++ b/tests/integration/ha_tests/test_ha.py
@@ -657,7 +657,7 @@ async def test_network_cut(ops_test, continuous_writes):
 
 @pytest.mark.abort_on_fail
 @pytest.mark.unstable
-async def test_scale_up_down(ops_test: OpsTest):
+async def test_scale_up_down(ops_test: OpsTest, continuous_writes):
     """Scale up and down the application and verify the replica set is healthy."""
     scales = [3, -3, 4, -4, 5, -5, 6, -6, 7, -7]
     for count in scales:
@@ -667,7 +667,7 @@ async def test_scale_up_down(ops_test: OpsTest):
 
 @pytest.mark.abort_on_fail
 @pytest.mark.unstable
-async def test_scale_up_down_removing_leader(ops_test: OpsTest):
+async def test_scale_up_down_removing_leader(ops_test: OpsTest, continuous_writes):
     """Scale up and down the application and verify the replica set is healthy."""
     scales = [3, -3, 4, -4, 5, -5, 6, -6, 7, -7]
     for count in scales:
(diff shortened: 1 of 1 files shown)


tests/integration/ha_tests/test_ha.py
```diff
@@ -657,7 +657,7 @@ async def test_network_cut(ops_test, continuous_writes):
 
 @pytest.mark.abort_on_fail
 @pytest.mark.unstable
-async def test_scale_up_down(ops_test: OpsTest, continuous_writes):
+async def test_scale_up_down(ops_test: OpsTest):
```
MiaAltieri, line 660: why remove `continous_writes`? I am fairly certain this must be here

some context: `continous_writes` is a fixture which applies a continous stream of writes to the DB, these writes and then used to verify if replication is operating correctly, or if the database properly recieves writes in edge case scenarios (ie killing a primary, scaling, up and scaling down). 

AFAICT removing this means that this test is no longer testing is the database can replicate writes during scaling.

tests/integration/ha_tests/test_ha.py
```diff
@@ -711,7 +711,7 @@ async def scale_and_verify(ops_test: OpsTest, count: int, remove_leader: bool =
     assert primary is not None, "Replica set has no primary"
 
 
-async def _verify_writes(ops_test: OpsTest, continuous_writes):
+async def _verify_writes(ops_test: OpsTest):
```
MiaAltieri, line 714: removing `continous_writes` here is correct. I am not sure why this was here in the first place. The `continous_writes` fixture should only be used on the test functions

MiaAltieri (changes requested):

tests/integration/ha_tests/test_ha.py
MiaAltieri: At a closer look at when `continous_writes` [was added](https://github.com/canonical/mongodb-operator/commit/63fc5e4d792fbc0efcc1775ec84326b153debdb4) it seems like the fixture was incorrectly applied to this helper function when it should have been applied to the tests function `test_scale_up_down_removing_leader ` - can you add the fixture to that test?

tests/integration/ha_tests/test_ha.py
delgod: We must check continuous writes in HA tests.

dmitry-ratushnyy: restored

delgod (approved):

MiaAltieri (approved):

PR #251, opened 2023-09-18
hold snap revision to prevent refresh

MiaAltieri:
## Issue
Snaps will auto refresh, meaning that snaps installed inside a charm and pinned to a specific revision won't necessarily be kept at that revision. This means a refresh will update the snap, leading to potential errors within the charm 

## Solution
hold the snap revision

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -810,13 +810,11 @@ class MongodbOperatorCharm(CharmBase):
             try:
                 snap_cache = snap.SnapCache()
                 snap_package = snap_cache[snap_name]
-
-                if not snap_package.present:
-                    snap_package.ensure(
-                        snap.SnapState.Latest, channel=snap_channel, revision=snap_revision
-                    )
-                    # snaps will auto refresh so it is necessary to hold the current revision
-                    snap_package.hold()
+                snap_package.ensure(
+                    snap.SnapState.Latest, channel=snap_channel, revision=snap_revision
+                )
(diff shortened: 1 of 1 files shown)


welpaolo (approved):

delgod (approved):

src/charm.py
```diff
@@ -815,6 +815,8 @@ def _install_snap_packages(self, packages: List[str]) -> None:
                     snap_package.ensure(
```
zmraul, line 815: Good opportunity to remove the `if not snap_package.present` check, and just leave `snap_package.ensure() & hold`

src/charm.py
```diff
@@ -815,6 +815,8 @@ def _install_snap_packages(self, packages: List[str]) -> None:
...
                         snap.SnapState.Latest, channel=snap_channel, revision=snap_revision
                     )
+                    # snaps will auto refresh so it is necessary to hold the current revision
+                    snap_package.hold()
```
marcoppenheimer, line 819: **suggestion:** Deindent this one level. You probably want to hold regardless, not just if it's not present.

src/charm.py
MiaAltieri: 🙇

marcoppenheimer (approved):

zmraul (approved):

dmitry-ratushnyy (approved):

PR #252, opened 2023-09-18
enable services to start on system start

MiaAltieri:
## issue
as of now the charm does not start snap services with them `enabled`, enabled flag allows the service to start when the system starts up). Enabling start up can help with high-availability in the case that the system stops and starts again.

## solution
start all snap services as enabled


dmitry-ratushnyy:
A bit more description to PR would be nice 😄

delgod (approved):

welpaolo (approved):

dmitry-ratushnyy (approved):

PR #256, opened 2023-09-21
[DPE-2562] POC Copy secrets from config server to shard

MiaAltieri:
## Issue
Sharded components should have the same keyfile as the config server, and we want to use the same operator username/password for shards as the config server

## Solution
Pass secrets to shards when they join a relation with the config server and have shards listen for this event and act appropriately

## Limitations
This does not consider what happens when the user changes the password on the config server, that should be handled in a separate PR

## Testing
As a POC there are not tests included in this PR. Testing was performed by hand  by hand:
```
juju deploy ./*charm --config role="config-server" config-server-one -n3
juju deploy ./*charm --config role="shard" shard-one -n3

# verify can only relate a shard to a single config server
juju integrate config-server-two:config-server shard-one:sharding
ERROR cannot add relation "config-server-two:config-server shard:sharding": establishing a new relation for shard:sharding would exceed its maximum relation limit of 1 (quota limit exceeded)

# verify same password
juju run config-server-one/0 get-password
juju run shard-one/0 get-password

# verify equal key files 
juju ssh config-server-one/0
sudo cat /var/snap/charmed-mongodb/current/etc/mongod/keyFile
juju ssh shard-one/0
sudo cat /var/snap/charmed-mongodb/current/etc/mongod/keyFile
```


src/machine_helpers.py
```diff
@@ -50,26 +47,3 @@ def add_args_to_env(var: str, args: str):
 
     with open(Config.ENV_VAR_PATH, "w") as service_file:
         service_file.writelines(env_vars)
-
```
MiaAltieri, line 53: move this to `src/charm.py` so the code is easily accessed by shared lib code

juditnovak (dismissed):
Sry, I know I wasn't invited, but just to confirm: looks good from Secrets perspective :-)

On the other hand: no tests? (If I'm missing context, pls ignore :-) )

lib/charms/mongodb/v0/shards_interface.py
```diff
@@ -0,0 +1,205 @@
...
+        if not self.charm.is_role(Config.Role.CONFIG_SERVER):
+            logger.info(
+                "skipping relation joined event ShardingRequirer is only be executed by config-server"
+            )
+            return
+
+        if not self.charm.unit.is_leader():
+            return
+
+        if "db_initialised" not in self.charm.app_peer_data:
```
dmitry-ratushnyy: is it possible to use a function from charm?

MiaAltieri:
Hi @juditnovak I forgot to update the PR description, this is a Proof of Concept, and so tests are out of scope for now. But look out for them in 24.04. Thank you so much for looking over the PR 🥰 , excited to get your hard work merged in here

lib/charms/mongodb/v0/shards_interface.py
```diff
@@ -0,0 +1,205 @@
...
+            event.defer()
+
+        # TODO Future PR, sync tls secrets and PBM password
+        self._update_relation_data(
+            event.relation.id,
+            {
+                "operator-password": self.charm.get_secret(
```
dmitry-ratushnyy: maybe we can move this to a constant?

lib/charms/mongodb/v0/shards_interface.py
```diff
@@ -0,0 +1,205 @@
...
+                    Config.Relations.APP_SCOPE,
+                    MongoDBUser.get_password_key_name_for_user("operator"),
+                ),
+                "key-file": self.charm.get_secret(
```
dmitry-ratushnyy: maybe we can move this to a constant?

lib/charms/mongodb/v0/shards_interface.py
```diff
@@ -0,0 +1,205 @@
...
+            logger.error("sharding interface not supported with config role=replication")
+            return
+
+        if not self.charm.is_role(Config.Role.SHARD):
+            logger.info(
+                "skipping relation changed event ShardingProvider is only be executed by shards"
+            )
+            return
+
+        if "db_initialised" not in self.charm.app_peer_data:
```
dmitry-ratushnyy: is it possible to use a function from charm?

lib/charms/mongodb/v0/shards_interface.py
MiaAltieri: in most cases yes, with: `self.charm.<function name>`

dmitry-ratushnyy: Sorry for my poor explanation. There is a property in charm, called `db_initialised`, could you please use it here?

lib/charms/mongodb/v0/shards_interface.py
```diff
@@ -0,0 +1,205 @@
...
+
+        Args:
+            relation_id: the identifier for a particular relation.
+            data: dict containing the key-value pairs
+                that should be updated in the relation.
+        """
+        if self.charm.unit.is_leader():
+            relation = self.charm.model.get_relation(self.relation_name, relation_id)
+            if relation:
+                relation.data[self.charm.model.app].update(data)
```
delgod, line 104: Both keys are secrets used for auth, we need to use juju secrets.
it can be separate PR

src/charm.py
```diff
@@ -542,6 +544,10 @@ def _on_get_password(self, event: ActionEvent) -> None:
 
     def _on_set_password(self, event: ActionEvent) -> None:
         """Set the password for the admin user."""
+        if self.is_role(Config.Role.SHARD):
```
delgod, line 578: sorry, offtopic, can we prevent role change

```
@@ -238,7 +238,13 @@ class MongodbOperatorCharm(CharmBase):
     @property
     def role(self) -> str:
         """Returns role of MongoDB deployment."""
-        return self.model.config["role"]
+        if not "role" in self.app_peer_data:
+            self.app_peer_data["role"] = self.model.config["role"]
+        return self.app_peer_data.get("role")
+
+    def is_role_changed(self) -> bool:
+        """Checks if application is running in provided role."""
+        return self.role != self.model.config["role"]

     def is_role(self, role_name: str) -> bool:
         """Checks if application is running in provided role."""
```

lib/charms/mongodb/v0/shards_interface.py
```diff
@@ -0,0 +1,205 @@
...
+        # TODO Future PR, enable shard drainage by listening for relation departed events
+
+    def _on_relation_joined(self, event):
+        """Handles providing shards with secrets and adding shards to the config server."""
+        if self.charm.is_role(Config.Role.REPLICATION):
+            self.unit.status = BlockedStatus("role replication does not support sharding")
+            logger.error("sharding interface not supported with config role=replication")
+            return
+
+        if not self.charm.is_role(Config.Role.CONFIG_SERVER):
```
delgod, line 60: why config server Requirer. tupically one provider and many requirers

delgod (changes requested):

lib/charms/mongodb/v0/shards_interface.py
MiaAltieri: TIL 
good point - let me switch this

src/charm.py
MiaAltieri: sure thing

lib/charms/mongodb/v0/shards_interface.py
MiaAltieri: Yes of course. We need to first migrate the charm to use external juju secrets. Right now all external secrets in the charm are sent via relation data bag.

lib/charms/mongodb/v0/shards_interface.py
MiaAltieri: Ah gotcha, sure thing !

dmitry-ratushnyy (dismissed):

src/charm.py
```diff
@@ -236,7 +242,17 @@ def db_initialised(self) -> bool:
     @property
     def role(self) -> str:
         """Returns role of MongoDB deployment."""
-        return self.model.config["role"]
+        if "role" not in self.app_peer_data and self.unit.is_leader():
+            self.app_peer_data["role"] = self.model.config["role"]
+        else:
+            # if leader hasn't set the role yet, use the one set by model
+            self.model.config["role"]
```
delgod:
```suggestion
            return self.model.config["role"]
```

src/charm.py
```diff
@@ -295,6 +311,22 @@ def _on_install(self, event: InstallEvent) -> None:
...
+        To prevent a user from migrating a cluster, and causing the component to become
+        unresponsive therefore causing a cluster failure, error the component. This prevents it
+        from executing other hooks with a new role.
+        """
+        # TODO in the future (24.04) support migration of components
+        if self.is_role_changed():
+            logger.error(
+                f"cluster migration currently not supported, cannot change from { self.model.config['role']} to {self.role}"
+            )
+            raise ShardingMigrationError(
```
delgod, line 327: just Q: hm, can we decline config change with exception?

delgod (dismissed):

src/charm.py
MiaAltieri: seems so, we get:
```
ubuntu@ip-172-31-37-213:~/mongodb-operator$ juju status
Model    Controller  Cloud/Region         Version  SLA          Timestamp
default  overlord    localhost/localhost  3.1.5    unsupported  09:29:42Z

App                Version  Status       Scale  Charm    Channel  Rev  Exposed  Message
config-server-one           error            1  mongodb             0  no       hook failed: "config-changed"
```

delgod: Let me rephrase: can we return an error when the user runs "juju config ..." instead of success? like with actions

lib/charms/mongodb/v0/shards_interface.py
```diff
@@ -0,0 +1,211 @@
...
+        if not self.charm.db_initialised:
+            event.defer()
+
+        # shards rely on the config server for secrets
+        relation_data = event.relation.data[event.app]
+        self.update_keyfile(key_file_contents=relation_data.get(KEYFILE_KEY))
+        try:
+            self.update_operator_password(new_password=relation_data.get(OPERATOR_PASSWORD_KEY))
+        except RetryError:
+            self.charm.unit.status = BlockedStatus("Shard not added to config-server")
```
delgod, line 163:
```suggestion
            self.charm.unit.status = BlockedStatus("Shard not added to config-server: cannot set operator password")
```

lib/charms/mongodb/v0/shards_interface.py
```diff
@@ -0,0 +1,211 @@
...
+
+    def update_operator_password(self, new_password: str) -> None:
+        """Updates the password for the operator user.
+
+        Raises:
+            RetryError
+        """
+        current_password = (
+            self.charm.get_secret(
+                Config.Relations.APP_SCOPE, MongoDBUser.get_password_key_name_for_user("operator")
```
delgod: getting secret can be slow (up to 2 seconds), can we add the following block before getting secret
```
if not new_password or not self.charm.unit.is_leader():
    return
```

lib/charms/mongodb/v0/shards_interface.py
```diff
@@ -0,0 +1,211 @@
...
+                "skipping relation changed event ShardingProvider is only be executed by shards"
+            )
+            return
+
+        if not self.charm.db_initialised:
+            event.defer()
+
+        # shards rely on the config server for secrets
+        relation_data = event.relation.data[event.app]
+        self.update_keyfile(key_file_contents=relation_data.get(KEYFILE_KEY))
```
delgod, line 143: restart on high loaded databases can be very slow (e.g. up to 10-20 minutes).
maybe we need some long-startup check here, before run update password

lib/charms/mongodb/v0/shards_interface.py
```diff
@@ -0,0 +1,211 @@
...
+        if not key_file_contents:
+            return
+
+        # put keyfile on the machine with appropriate permissions
+        self.charm.push_file_to_unit(
+            parent_dir=Config.MONGOD_CONF_DIR, file_name=KEY_FILE, file_contents=key_file_contents
+        )
+
+        # when the contents of the keyfile change, we must restart the service
+        self.charm.restart_mongod_service()
```
delgod, line 228: can we compare old and new keyfile to do restart only if it is needed.
restarted can be slow

delgod (changes requested):

lib/charms/mongodb/v0/shards_interface.py
MiaAltieri: My bad, I thought secrets were cached, they are not. Thank you Mykola!

lib/charms/mongodb/v0/shards_interface.py
MiaAltieri: hmmm, I am of the opinion that messages in Juju status shouldn't provide so much context. (That is what Jon taught me when I first joined). I think it is better for users to see the error and then check `juju debug-log` to get the context of why the charm is in blocked status.

My point of view is that the status should only report a status relevant to the **primary goal of the hook execution** - i.e. the primary goal here is to add the shard to the config server. (Setting the operator password is merely part of achieving that goal) 

If I have understood wrong, please let me know. We can accept your change above. But if we do this, and offer context like what you suggested, we should update status messages elsewhere in the code.

src/charm.py
MiaAltieri: From Mykola: 
> Let me rephrase: can we return an error when the user runs "juju config ..." instead of success? like with actions

MiaAltieri: Unfortunately, we cannot do something like `event.fail`, ConfigChanged events do not have the `fail` function (those are special to acctions)

```
AttributeError: 'ConfigChangedEvent' object has no attribute 'fail'
```

lib/charms/mongodb/v0/shards_interface.py
MiaAltieri: yes!

lib/charms/mongodb/v0/shards_interface.py
dmitry-ratushnyy: Maybe we can use "OperatorUser" from users.py

OperatorUser.get_username()

lib/charms/mongodb/v0/shards_interface.py
```diff
@@ -0,0 +1,228 @@
...
+LIBID = "55fee8fa73364fb0a2dc16a954b2fd4a"
+
+# Increment this major API version when introducing breaking changes
+LIBAPI = 0
+
+# Increment this PATCH version before using `charmcraft publish-lib` or reset
+# to 0 if you are raising the major API version
+LIBPATCH = 1
+KEYFILE_KEY = "key-file"
+OPERATOR_PASSWORD_KEY = "operator-password"
```
dmitry-ratushnyy: is it possible to use this `MongoDBUser.get_password_key_name_for_user(username)` instead of  OPERATOR_PASSWORD_KEY  or maybe define constant like
OPERATOR_PASSWORD_KEY = MongoDBUser.get_password_key_name_for_user(username)

lib/charms/mongodb/v0/shards_interface.py
```diff
@@ -0,0 +1,228 @@
...
+        self.framework.observe(
+            charm.on[self.relation_name].relation_joined, self._on_relation_joined
+        )
+        # TODO Future PR, enable shard drainage by listening for relation departed events
+
+    def _on_relation_joined(self, event):
+        """Handles providing shards with secrets and adding shards to the config server."""
+        if self.charm.is_role(Config.Role.REPLICATION):
+            self.charm.unit.status = BlockedStatus("role replication does not support sharding")
+            logger.error("sharding interface not supported with config role=replication")
```
dmitry-ratushnyy, line 57: nits: maybe we can use `Config.Role.REPLICATION` in the error message?

lib/charms/mongodb/v0/shards_interface.py
```diff
@@ -0,0 +1,228 @@
...
+            charm.on[self.relation_name].relation_changed, self._on_relation_changed
+        )
+
+        # TODO Future PR, enable shard drainage by observing relation departed events
+
+    def _on_relation_changed(self, event):
+        """Retrieves secrets from config-server and updates them within the shard."""
+        if self.charm.is_role(Config.Role.REPLICATION):
+            self.charm.unit.status = BlockedStatus("role replication does not support sharding")
+            logger.error("sharding interface not supported with config role=replication")
```
dmitry-ratushnyy, line 128: nit: maybe we can use `Config.Role.REPLICATION` in the error message?

src/charm.py
```diff
@@ -898,6 +898,24 @@ def _instatiate_keyfile(self, event: StartEvent) -> None:
             file_contents=self.get_secret(APP_SCOPE, Config.Secrets.SECRET_KEYFILE_NAME),
         )
 
+    def get_keyfile_contents(self) -> str:
+        """Retrieves the contents of the keyfile on host machine."""
+        # wait for keyFile to be created by leader unit
+        if not self.get_secret(APP_SCOPE, Config.Secrets.SECRET_KEYFILE_NAME):
+            logger.debug("waiting for leader unit to generate keyfile contents")
+            return
```
delgod, line 906: Does non-juju-leader really need to wait? it has new keyfile from relation - can restart without this check

lib/charms/mongodb/v0/shards_interface.py
delgod: correct, it is ok to have short Status message if there are good explanation in debug log exists

lib/charms/mongodb/v0/shards_interface.py
delgod: we cache secrest in moment of the first retrieval till the end of event

delgod (changes requested):

lib/charms/mongodb/v0/shards_interface.py
MiaAltieri: Agreed, and I think you will find we [have good log messages for this case](https://github.com/canonical/mongodb-operator/blob/6213835db63cf5afaed5541ed34810e51daa1a54/lib/charms/mongodb/v0/shards_interface.py#L192-L199)

src/charm.py
MiaAltieri: this function is independent of restarting and relations. This function `"""Retrieves the contents of the keyfile on host machine."""`, if the leader has not yet generated the keyfile contents, then the file doesn't exist and this function cannot operate

dmitry-ratushnyy (dismissed):

lib/charms/mongodb/v0/shards_interface.py
MiaAltieri: good suggestion!

dmitry-ratushnyy (dismissed):

delgod (approved):

dmitry-ratushnyy (approved):

PR #259, opened 2023-09-27
[DPE-2633] Update data platform libs to use juju secrets for charm external secrets

dmitry-ratushnyy:
## Issue
The charm must support external secrets to keep sensitive relation data

## Solution

To support external secrets we need to import a new version of data platform libs and fix any issues it can cause


tests/integration/relation_tests/new_relations/application-charm/lib/charms/data_platform_libs/v0/database_requires.py
dmitry-ratushnyy, line 1: This file is fetched from data-platform libs as-is

tests/integration/relation_tests/new_relations/application-charm/lib/charms/data_platform_libs/v0/data_interfaces.py
dmitry-ratushnyy, line 1: This file is fetched from data-platform libs as-is

tests/integration/relation_tests/new_relations/application-charm/lib/charms/data_platform_libs/v0/database_requires.py
dmitry-ratushnyy, line 1: removed deprecated lib

.github/workflows/ci.yaml
```diff
@@ -104,6 +104,17 @@ jobs:
         uses: actions/download-artifact@v3
         with:
           name: ${{ needs.build.outputs.artifact-name }}
+      - name: Free disk space
```
MiaAltieri, line 109: interesting! This looks valuable, but I am a noob why is this need?

lib/charms/mongodb/v0/mongodb_provider.py
```diff
@@ -248,7 +248,7 @@ def update_app_relation_data(self) -> None:
```
MiaAltieri: bump lib patch. Perhaps we should merge #266 first since we want `6/edge` to have its own LIBAPI

MiaAltieri (commented):
LGTM - a question though, do we need to also future PRs for TLS and S3 relations? I think those pass some secrets as well

.github/workflows/ci.yaml
dmitry-ratushnyy: There was issue on CI: some tests were not able to start because of "No space left on device" error. You can [check here ](https://github.com/canonical/mongodb-operator/actions/runs/6498579157) for example. This change frees some space

Screenshot 2023-10-13 at 11 05 02

lib/charms/mongodb/v0/mongodb_provider.py
dmitry-ratushnyy: Done, thank you 😄

delgod (dismissed):

dmitry-ratushnyy:
> LGTM - a question though, do we need to also future PRs for TLS and S3 relations? I think those pass some secrets as well

No, I don't think so. The only changes I made in the mongodb_provides were because how we were getting passwords. The rest should work "out of box"

MiaAltieri (dismissed):

juditnovak (commented):
Looks good to me in the 1st place -- however I'd approve once there's a green pipeline for tests that may use any relations.

(I came across too many tests "hiding" using databag, I only believe they're all gone once the pielines come up green :-) )

lib/charms/mongodb/v1/mongodb_provider.py
```diff
@@ -260,11 +260,27 @@ def update_app_relation_data(self) -> None:
...
+            relation (Relation): The relation for each the password is cached.
+
+        Returns:
+            str: The password.
+        """
+        password = self.database_provides.fetch_my_relation_field(relation.id, "password")
+        if password:
+            return password
+        password = generate_password()
+        self.database_provides.update_relation_data(relation.id, {"password": password})
```
MiaAltieri, line 276: noob question, we use external juju secrets, why update `relation_data`?

MiaAltieri (commented):
I believe a followup PR will be needed to use external secrets within `shards_interface.py` :) but this work looks good, just left a question/comment

lib/charms/mongodb/v1/mongodb_provider.py
dmitry-ratushnyy: we need to  keep the  password in secrets, and the data_interfaces lib will do the rest

juditnovak: This function is seamlessly selecting between databag or secrets to store the relation's data.
It's an abstraction layer, that hides the actual destination where the data will be saved. If the relation is using secrets, the call will create/update the corresponding secret. If we're on Juju2 (on either side), the function will update the databag.

In this case the password will be saved in a secret.

delgod (approved):

juditnovak (approved):
LGTM

MiaAltieri (approved):
looks good to me, but please make an extra PR for sharding since sharding passes secrets via data bag

PR #260, opened 2023-09-29
[DPE-2645] POC Add shards to cluster

MiaAltieri:
## Issue
Shards are not currently added to cluster

## Solution
Add shards to cluster.

## Limitations
This does limit the number of shards added to a cluster at a given time. This should be done by checking the Primary shard, and basing the number of shards added on the number of secondaries in the Primary shard. There is no [MongoDB Command](https://www.mongodb.com/docs/manual/reference/command/nav-sharding/) to retrieve the primary shard, but there is a [mongosh one](https://www.mongodb.com/docs/manual/reference/method/sh.status/#sh.status--). This can be implemented, but might be tricky. This should be saved for a future epic.

## Testing
As a POC there are not tests included in this PR. Testing was performed by hand  by hand:
```
[direct: mongos] admin> sh.status()
shardingVersion
{ _id: 1, clusterId: ObjectId("65152d236139b01f40f20c7d") }
---
shards
[
  {
    _id: 'shard-one',
    host: 'shard-one/10.61.64.125:27017',
    state: 1,
    topologyTime: Timestamp({ t: 1695886769, i: 4 })
  }
]
---
active mongoses
[ { '6.0.6-5': 1 } ]
---
autosplit
{ 'Currently enabled': 'yes' }
---
balancer
{
  'Currently running': 'no',
  'Currently enabled': 'yes',
  'Failed balancer rounds in last 5 attempts': 0,
  'Migration Results for the last 24 hours': 'No recent migrations'
}
---
databases
[
  {
    database: { _id: 'config', primary: 'config', partitioned: true },
    collections: {
      'config.system.sessions': {
        shardKey: { _id: 1 },
        unique: false,
        balancing: true,
        chunkMetadata: [ { shard: 'shard-one', nChunks: 1024 } ],
        chunks: [
          'too many chunks to print, use verbose if you want to force print'
        ],
        tags: []
      }
    }
  }
]

[direct: mongos] admin> use test_db
[direct: mongos] admin> db.createCollection("cool_animals")
[direct: mongos] admin> db.cool_animals.insertOne({"horses":"unicorn"})
[direct: mongos] admin> sh.enableSharding("test_db")
{
  ok: 1,
  '$clusterTime': {
    clusterTime: Timestamp({ t: 1695887676, i: 1 }),
    signature: {
      hash: Binary.createFromBase64("/dNkOmzElTKr0AiBFTYgUWrxu24=", 0),
      keyId: Long("7283777600688750617")
    }
  },
  operationTime: Timestamp({ t: 1695887676, i: 1 })
}


# then on shard
shard-one [primary] admin> show dbs
admin    208.00 KiB
config   620.00 KiB
local    564.00 KiB
test_db   40.00 KiB
```


lib/charms/mongodb/v0/mongos.py
```diff
@@ -0,0 +1,168 @@
...
+    tls_internal: bool
+
+    @property
+    def uri(self):
+        """Return URI concatenated from fields."""
+        hosts = [f"{host}:{self.port}" for host in self.hosts]
+        hosts = ",".join(hosts)
+        # Auth DB should be specified while user connects to application DB.
+        auth_source = ""
+        if self.database != "admin":
```
dmitry-ratushnyy, line 58: Maybe we can use https://github.com/canonical/mongodb-operator/blob/6/edge/lib/charms/mongodb/v0/users.py#L74 or move this to constants? :)

src/charm.py
```diff
@@ -754,6 +760,23 @@ def _is_user_created(self, user: MongoDBUser) -> bool:
...
+    ) -> MongosConfiguration:
+        external_ca, _ = self.tls.get_tls_files(UNIT_SCOPE)
+        internal_ca, _ = self.tls.get_tls_files(APP_SCOPE)
+
+        return MongosConfiguration(
+            database=user.get_database_name(),
+            username=user.get_username(),
+            password=self.get_secret(APP_SCOPE, user.get_password_key_name()),
+            hosts=hosts,
+            port=27018,
```
dmitry-ratushnyy: Could we use port from config.py here ?

src/charm.py
```diff
@@ -754,6 +760,23 @@ def _is_user_created(self, user: MongoDBUser) -> bool:
     def _set_user_created(self, user: MongoDBUser) -> None:
```
dmitry-ratushnyy, line 757: is it possible to use "_is_user_created" from charm here? https://github.com/canonical/mongodb-operator/blob/0e75a6f4e8612116e7e22ef637f042826b563317/src/charm.py#L751

lib/charms/mongodb/v0/mongos.py
```diff
@@ -0,0 +1,168 @@
...
+        Raises:
+            ConfigurationError, ConfigurationError, OperationFailure
+        """
+        shard_list = self.client.admin.command("listShards")
+        curr_members = [
+            self._hostname_from_hostport(member["host"]) for member in shard_list["shards"]
+        ]
+        return set(curr_members)
+
+    def add_shard(self, shard_name, shard_hosts, shard_port=27017):
```
dmitry-ratushnyy: Could we use a constant from config here? :)

lib/charms/mongodb/v0/mongos.py
MiaAltieri: No, we cant/should not. `admin`, here is not a user, but the database that is used to authenticate the connection to MongoDB

src/charm.py
MiaAltieri: No, and this code isn't relevant to this PR, plus `is_user_created`, relies `set_user_created`, we cannot replace it.

See [src/charm.py here](https://github.com/canonical/mongodb-operator/blob/0e75a6f4e8612116e7e22ef637f042826b563317/src/charm.py#L750C1-L755C75)

lib/charms/mongodb/v0/mongos.py
MiaAltieri: good idea, thank you

src/charm.py
MiaAltieri: good idea, thank you

lib/charms/mongodb/v0/mongos.py
```diff
@@ -0,0 +1,168 @@
...
+        self.client = None
+
+    def get_shard_members(self) -> Set[str]:
+        """Gets shard members.
+
+        Returns:
+            A set of the shard members as reported by mongos.
+
+        Raises:
+            ConfigurationError, ConfigurationError, OperationFailure
```
MiaAltieri: duplicate ConfigurationError

lib/charms/mongodb/v0/mongos.py
```diff
@@ -0,0 +1,168 @@
...
+        curr_members = [
+            self._hostname_from_hostport(member["host"]) for member in shard_list["shards"]
+        ]
+        return set(curr_members)
+
+    def add_shard(self, shard_name, shard_hosts, shard_port=27017):
+        """Adds shard to the cluster.
+
+        Raises:
+            ConfigurationError, ConfigurationError, OperationFailure
```
MiaAltieri: duplicate

lib/charms/mongodb/v0/shards_interface.py
```diff
@@ -84,8 +92,67 @@ def _on_relation_joined(self, event):
...
+            cluster_shards = mongo.get_shard_members()
+            relation_shards = self._get_shards_from_relations(departed_shard_id)
+
+            # TODO Future PR, limit number of shards add at a time, based on the number of
+            # replicas in the primary shard
+            for shard in relation_shards - cluster_shards:
+                try:
+                    shard_hosts = self._get_shard_hosts(shard)
+                    if not len(shard_hosts):
+                        logger.info("host info for shard %s not yet added, skipping", shard)
```
delgod, line 141: should we defer instead of skipping?

lib/charms/mongodb/v0/shards_interface.py
```diff
@@ -103,6 +170,28 @@ def _update_relation_data(self, relation_id: int, data: dict) -> None:
...
+                if relation.id != departed_shard_id
+            ]
+        )
+
+    def _get_shard_hosts(self, shard_name) -> str:
+        """Retrieves the hosts for a specified shard."""
+        relations = self.model.relations[self.relation_name]
+        for relation in relations:
+            if self._get_shard_name_from_relation(relation) == shard_name:
+                return json.loads(relation.data[relation.app].get(HOSTS_KEY, "[]"))
```
delgod: Can we get a list of hosts from the relation itself (e.g. `relation.units`)? without passing data via relation

delgod (changes requested):

lib/charms/mongodb/v0/shards_interface.py
MiaAltieri: nope we should not, when the juju updates the relation data, the config server will receive **another** `relation_changed` event.

We can have a defer but it isn't necessary, and IMO a bad practice. The reason is, is that once the units are added to `relation.units`, the `config-server` receives a `relation-changed` event, because the relation data has changed. The relation data has changed from `relation.units` being `set()` to `set(Unit_1)` so the config-server receives the relation changed event.

Further, it is not guaranteed that the deferred one will have the new/correct relation data.That depends on if juju has updated the relation data. (but if juju has changed the relation data, we will have already received a relation_changed_event, so the deferred one is pointless here) 

[From Juju docs:](https://juju.is/docs/sdk/the-lifecycle-of-charm-integrations)
> A relation-changed hook can always exit without error when the current remote unit is missing data, because **the hook is guaranteed to be run again when that data changes:** assuming the remote unit is running a charm that agrees on how to implement the interface, **the data will change and the hook will be run again.**

lib/charms/mongodb/v0/shards_interface.py
MiaAltieri: I would hope so, but `relation.units` doesn't seem to give us the information about the units. I asked [the ops team](https://chat.charmhub.io/charmhub/pl/m7jmq3wasfg6dj8m945mqseosa)

MiaAltieri: From what I can tell from looking at [`ops.model.Unit`](https://ops.readthedocs.io/en/latest/_modules/ops/model.html#Unit) there is no straightforward way to get the IP address. I assume that what we have here is the accepted way of doing things. (After-all this keeps with the pattern we use for client relations in Data Platform Interface)

dmitry-ratushnyy (dismissed):

lib/charms/mongodb/v0/shards_interface.py
MiaAltieri: yes

lib/charms/mongodb/v0/shards_interface.py
MiaAltieri: I have also tested this by hand with `juju debug-hooks` on the `config-server` side where these hooks get executed one-by-one:
- `relation-joined` <- `config-server` adds relation data for shared passwords/keyfile. This is triggered from adding the relation
- `relation-changed` <- in this first one juju has not added the relation units and the function `add_shards` returns without adding the shards. This is trigged from relation data changed in `relation_joined` by `config-server`.
- `relation-changed` <- in this second one, juju has added the relation units and the function `add_shards` successfully adds the shards. This is trigged from juju adding the relation units.

Mehdi-Bendriss: We had a discussion and it seems that a subsequent deferred run of the hook may not have much effect here as the state of the shards is not expected to change until a new `_changed` event is received with updated data, i think it's safe to omit in this specific case

delgod (approved):

dmitry-ratushnyy (approved):

PR #261, opened 2023-10-04
[DPE-2646] - POC Rotate password

MiaAltieri:
## Issue
changes in password on config server aren't propagated to shard

## Solution
update password passed on shard when config server password changes

## Testing
```

# verify same password
juju run config-server-one/0 get-password
juju run shard-one/0 get-password

# verify cannot set shard password
juju run shard-one/0 set-password
juju run config-server-one/0 get-password
juju run shard-one/0 get-password

# verify setting config server password updates shard
juju run config-server-one/0 set-password
juju run config-server-one/0 get-password
juju run shard-one/0 get-password
```


delgod (approved):

src/charm.py
```diff
@@ -782,6 +779,25 @@ def _get_user_or_fail_event(self, event: ActionEvent, default_username: str) ->
...
+            return
+
+        # changing the backup password while a backup/restore is in progress can be disastrous
+        pbm_status = self.backups._get_pbm_status()
+        if isinstance(pbm_status, MaintenanceStatus):
+            event.fail("Cannot change password while a backup/restore is in progress.")
+            return
+
+        # only leader can write the new password into peer relation.
+        if not self.unit.is_leader():
```
dmitry-ratushnyy, line 795: Is it possible that a unit will be a leader but also a SHARD? Meaning `if self.is_role(Config.Role.SHARD)` will return `True`

MiaAltieri: Yes it is possible, that the unit is both `leader` and `SHARD`. But it doesn't matter if the unit is `leader` if it is a `SHARD`, since as long as the unit is a `SHARD` we should not allow password rotation.

dmitry-ratushnyy (approved):

PR #262, opened 2023-10-06
[DPE-2648] POC Remove shards

MiaAltieri:
## Issue
Shard removal not implemented

## Solution
Remove shards

## Context
When tested by hand, it appears that it takes several calls of `removeShard` for the shard to be properly removed.

## Limitations
This PR is limited by the behaviour that remote (and local) applications exhibit when scaling down. Specifically if two applications `A` and `B` are related and one of these applications scale down both `A` and `B` will receive `relation-broken` and `relation-departed` events. This can lead to some unwanted behaviour. (Which is shown in the consistently failing `relation` tests). 

The workaround for this will require more research and potentially a new library for the code needed to fix this situation and will be done in the next PR.

[See bug filed by DPE team regarding this](https://bugs.launchpad.net/juju/+bug/1979811)
[See ongoing conversation to find a clean solution](https://chat.charmhub.io/charmhub/pl/ze3qfn3yxirq7c4p4dyuo9k4bw)

## Testing
As this PR is a POC testing was done by hand:
```

# relate shards

# remove a relation to shard
juju remove-relation config-server-one:config-server   shard-two:sharding

# manually verify shard config 
juju ssh config-server
charmed.mongodb.mongosh <URI>
sh.status()

# attempt to remove last replica and verify failure
juju remove-relation config-server-one:config-server   shard-one:sharding
```


lib/charms/mongodb/v0/mongos.py
```diff
@@ -4,11 +4,11 @@
...
 from dataclasses import dataclass
-from typing import Optional, Set
+from typing import Dict, Optional, Set
 from urllib.parse import quote_plus
 
+from charms.mongodb.v0.mongodb import NotReadyError
 from pymongo import MongoClient
-from pymongo.errors import PyMongoError
 
 from config import Config
```
MiaAltieri, line 13: bump libpatch

lib/charms/mongodb/v0/shards_interface.py
```diff
@@ -33,16 +44,25 @@
 # to 0 if you are raising the major API version
 LIBPATCH = 2
```
MiaAltieri: bump libpatch

lib/charms/mongodb/v0/shards_interface.py
```diff
@@ -264,6 +337,80 @@ def _on_relation_changed(self, event):
...
+            return
+
+        self.charm.unit.status = MaintenanceStatus("Draining shard from cluster")
+        # mongos hosts must be retrieved via relation data, as relation.units are not available in
+        # broken
+        mongos_hosts = json.loads(event.relation.data[event.relation.app].get(HOSTS_KEY))
+        drained = False
+        while not drained:
+            try:
+                time.sleep(10)
```
MiaAltieri, line 374:
```suggestion
                # no need to continuously check and abuse resources while shard is draining
                time.sleep(10)
```

lib/charms/mongodb/v0/shards_interface.py
```diff
@@ -264,6 +337,80 @@ def _on_relation_changed(self, event):
...
+                drained = self.drained(mongos_hosts, self.charm.app.name)
+                draining_status = (
+                    "Shard is still draining" if not drained else "Shard is fully drained."
+                )
+                logger.debug(draining_status)
+                # no need to continuously check and abuse resources while shard is draining
```
MiaAltieri: remove comment

lib/charms/mongodb/v0/mongos.py
```diff
@@ -158,6 +166,75 @@ def add_shard(self, shard_name, shard_hosts, shard_port=Config.MONGODB_PORT):
...
+            RemovePrimaryShardError
+        """
+        sc_status = self.client.admin.command("listShards")
+        # It is necessary to call removeShard multiple times on a shard to guarantee removal.
+        # Allow re-removal of shards that are currently draining.
+        if self._is_any_draining(sc_status, ignore_shard=shard_name):
+            cannot_remove_shard = (
+                f"cannot remove shard {shard_name} from cluster, another shard is draining"
+            )
+            logger.info(cannot_remove_shard)
```
dmitry-ratushnyy: maybe using error instead of info is better?

lib/charms/mongodb/v0/mongos.py
```diff
@@ -158,6 +166,75 @@ def add_shard(self, shard_name, shard_hosts, shard_port=Config.MONGODB_PORT):
...
+            raise NotReadyError(cannot_remove_shard)
+
+        # TODO Follow up PR, there is no MongoDB command to retrieve primary shard, this is
+        # possible with mongosh.
+        primary_shard = False
+        if primary_shard:
```
dmitry-ratushnyy, line 189: Not sure why we have this check? 😄

lib/charms/mongodb/v0/mongos.py
```diff
@@ -158,6 +166,75 @@ def add_shard(self, shard_name, shard_hosts, shard_port=Config.MONGODB_PORT):
...
+    def _is_shard_draining(self, shard_name: str) -> bool:
+        """Reports if a given shard is currently in the draining state.
+
+        Raises:
+            ConfigurationError, OperationFailure, ShardNotInClusterError,
+            ShardNotPlannedForRemovalError
+        """
+        sc_status = self.client.admin.command("listShards")
+        for shard in sc_status["shards"]:
+            if shard["_id"] == shard_name:
```
dmitry-ratushnyy, line 227: is it possible to combine. `if  shard["_id"] == shard_name:` and `if "draining" not in shard:`?

lib/charms/mongodb/v0/shards_interface.py
```diff
@@ -116,9 +141,27 @@ def _on_relation_event(self, event):
...
+            self.remove_shards(departed_relation_id)
+        except NotDrainedError:
+            # it is necessary to removeShard multiple times for the shard to be removed.
+            logger.info(
+                "Shard is still present in the cluster after removal, will defer and remove again."
+            )
+            event.defer()
+            return
+        except OperationFailure as e:
+            if e.code == 20:
```
dmitry-ratushnyy: Could we use constant instead of a magic number? 😄

lib/charms/mongodb/v0/shards_interface.py
```diff
@@ -264,6 +337,80 @@ def _on_relation_changed(self, event):
...
+
+        self.charm.unit.status = MaintenanceStatus("Draining shard from cluster")
+        # mongos hosts must be retrieved via relation data, as relation.units are not available in
+        # broken
+        mongos_hosts = json.loads(event.relation.data[event.relation.app].get(HOSTS_KEY))
+        drained = False
+        while not drained:
+            try:
+                # no need to continuously check and abuse resources while shard is draining
+                time.sleep(10)
```
dmitry-ratushnyy, line 374: just curious why do we sleep first?

lib/charms/mongodb/v0/shards_interface.py
```diff
@@ -53,27 +73,16 @@ def __init__(
...
-                "skipping relation joined event ShardingRequirer is only be executed by config-server"
-            )
-            return
-
-        if not self.charm.unit.is_leader():
-            return
-
-        if not self.charm.db_initialised:
-            event.defer()
+        if not self.pass_hook_checks(event):
```
dmitry-ratushnyy, line 86: Maybe some logging here could be useful for debug purposes?

lib/charms/mongodb/v0/mongos.py
MiaAltieri: We can remove the check as ` primary_shard = False` is not implemented yet (meant for future PR). Would you prefer that? I was hoping to get the content below the check reviewed so that the follow up PR would only have to implement the line `primary_shard = False`

lib/charms/mongodb/v0/shards_interface.py
MiaAltieri: Consider having sleep last:
```
       try:
               drained = self.drained(mongos_hosts, self.charm.app.name)
               draining_status = (
                    "Shard is still draining" if not drained else "Shard is fully drained."
                )
                logger.debug(draining_status)
                time.sleep(10)
      except ... : 
                log and retry
```
If the draining commands fail, the `sleep(10)` is skipped. Thus the check repetitively happens and fails without waiting in between. Ideally if there is a failure we should sleep in between failures so that we don't abuse system resources.

lib/charms/mongodb/v0/shards_interface.py
MiaAltieri: the command `pass_hook_checks` has [verbose logs for each scenario](https://github.com/canonical/mongodb-operator/blob/e8d7167762a262f6bcaf3a8c3c5460a46b720c86/lib/charms/mongodb/v0/shards_interface.py#L103-L127), we could include a log here but it wouldn't be as informative as the ones in `pass_hook_checks`.

I am happy to include a log message here. But not sure what I could say that would informative / not a restatement of what was logged in `pass_hook_checks`.

What would you like logged here? I can add it

lib/charms/mongodb/v0/mongos.py
MiaAltieri: ~Good eye! Thank you :)~

Update, sadly no it is not possible, since there are specific cases that we need to handle :/

lib/charms/mongodb/v0/shards_interface.py
MiaAltieri: Of course, thank you!

lib/charms/mongodb/v0/mongos.py
MiaAltieri: hmm, I am not sure. I don't think it is an error, since it results in a retry, but I will change it :)

lib/charms/mongodb/v0/shards_interface.py
dmitry-ratushnyy: Yeah, nice logging there. I though of having something like "Skipping relation joined event: hook checks re not passed" or similar. But it is not a blocker I think 😄

lib/charms/mongodb/v0/mongos.py
dmitry-ratushnyy: If you do plan to have a function that will do this check, I think it is a good idea to create a stub function for that , that will return `False` for now

lib/charms/mongodb/v0/mongos.py
dmitry-ratushnyy: I suggested to make it an error just because we are throwing exception here, but, not a blocker :)

lib/charms/mongodb/v0/shards_interface.py
dmitry-ratushnyy: Got it, thank you

lib/charms/mongodb/v0/shards_interface.py
MiaAltieri: Perf, I'll add exactly that :) Thank you

lib/charms/mongodb/v0/mongos.py
MiaAltieri: Great idea! I will do exactly that, thank you!

lib/charms/mongodb/v0/shards_interface.py
```diff
@@ -87,27 +96,43 @@ def _on_relation_joined(self, event):
...
 
         if not self.charm.unit.is_leader():
-            return
+            return False
 
         if not self.charm.db_initialised:
+            logger.info("Deferring %s. db is not initialised.", type(event))
             event.defer()
+            return False
+
```
delgod, line 128: Draining and adding should be deferred if backup/restore is in progress.

lib/charms/mongodb/v0/shards_interface.py
```diff
@@ -152,11 +195,40 @@ def add_shards(self, departed_shard_id):
...
+        """
+        with MongosConnection(self.charm.mongos_config) as mongo:
+            cluster_shards = mongo.get_shard_members()
+            relation_shards = self._get_shards_from_relations(departed_shard_id)
+
+            for shard in cluster_shards - relation_shards:
+                self.charm.unit.status = MaintenanceStatus(f"Draining shard {shard}")
+                logger.info("Attempting to removing shard: %s", shard)
+                mongo.remove_shard(shard)
+                logger.info("Attempting to removing shard: %s", shard)
```
delgod: same message twice
```suggestion
```

lib/charms/mongodb/v0/mongos.py
```diff
@@ -158,6 +166,75 @@ def add_shard(self, shard_name, shard_hosts, shard_port=Config.MONGODB_PORT):
...
+            # on it. All sharded collections should perform `MovePrimary`
+            cannot_remove_primary_shard = (
+                f"Shard {shard_name} is the primary shard, cannot remove."
+            )
+            logger.error(cannot_remove_primary_shard)
+            raise RemovePrimaryShardError(cannot_remove_primary_shard)
+
+        logger.info("Attempting to remove shard %s", shard_name)
+        self.client.admin.command("removeShard", shard_name)
+        logger.info("Shard %s, now draining", shard_name)
```
delgod: Please extend the log with `remaining.chunks` and `dbsToMove` info from `removeShard` output.
e.g.
```suggestion
        logger.info("Shard %s is %s now. Remaining chunks: %i. DBs to move: %s.", shard_name, status["state"], status["remaining"]["chunks"], ",".join(status["dbsToMove"])
```

lib/charms/mongodb/v0/shards_interface.py
```diff
@@ -53,27 +73,16 @@ def __init__(
...
-            )
-            return
-
-        if not self.charm.unit.is_leader():
-            return
-
-        if not self.charm.db_initialised:
-            event.defer()
+        if not self.pass_hook_checks(event):
             return
```
delgod, line 88: Should we defer `if not self.charm.db_initialised`?

delgod (changes requested):

lib/charms/mongodb/v0/shards_interface.py
MiaAltieri: Yes we should defer, (and [we do defer in this case, in the function `pass_hook_checks `](https://github.com/canonical/mongodb-operator/blob/e8d7167762a262f6bcaf3a8c3c5460a46b720c86/lib/charms/mongodb/v0/shards_interface.py#L122-L126)) 

Reasoning to defer is that we are not guaranteed to receive another `relation_joined` event, so we should keep it on the event queue until we can process the event

delgod (approved):

dmitry-ratushnyy (approved):

Issue #247, opened 2023-09-15
Flakey Relation Integration tests.


MiaAltieri:
As of recently CI reports [flakey relation integration tests](https://github.com/canonical/mongodb-operator/actions/runs/6179708184/job/16775133065) - seems to have started occuring after updating CI to 3.1.5.

MiaAltieri:
Issue seems to relation error:
```
unit-mongodb-3: 14:39:37 ERROR juju.worker.uniter resolver loop error: cannot enter scope for unit "mongodb/3" in relation "application:multiple-database-clusters mongodb:database": inconsistent state in EnterScope
```
which is currently an error in relations on [Juju 3.1.5 - update to 3.1.6 when available](https://chat.charmhub.io/charmhub/pl/gc8169mqr3fodqymn8fowxw77y)

MiaAltieri:
fixed in #272 and #263

PR #263, opened 2023-10-06
[DPE-2613] fix rel tests (6/edge)

MiaAltieri:
## Issue
After updating to Juju 3.1 the charm failed on its relation tests #247

This was due to when apps scale down in 3.1, they receive relation broken events. Thereby attempts to remove users were made when relations were actually not removed, but instead a unit was being removed.  

## Solution
Add a check in `relation-departed`. In this function we can tell if we are receiving the `relation-broken` event due to scale down or relation removal. 

## more
Once this PR is accepted, we will make another one for branch `5/edge`

### DIFF
--- a/lib/charms/mongodb/v0/mongodb_provider.py
+++ b/lib/charms/mongodb/v0/mongodb_provider.py
@@ -97,14 +97,14 @@ class MongoDBProvider(Object):
             return
 
         # assume relation departed is due to manual removal of relation.
-        remove_users = True
+        relation_departed = True
 
         # check if relation departed is due to current unit being removed. (i.e. scaling down the
         # application.)
         if event.departing_unit == self.charm.unit:
-            remove_users = False
+            relation_departed = False
 
-        self.charm.app_peer_data["remove_users"] = json.dumps(remove_users)
+        self.charm.app_peer_data[f"relation_{event.relation.id}_departed"] = json.dumps(relation_departed)
 
     def _on_relation_event(self, event):
         """Handle relation joined events.
@@ -141,14 +141,15 @@ class MongoDBProvider(Object):
             # we receives relation broken events when: the relation has been removed, units are
             # scaling down, or the application has been removed. Only proceed to process user
             # removal if the relation has been removed.
-            if "remove_users" not in self.charm.app_peer_data:
+            relation_departed_key = f"relation_{event.relation.id}_departed"
+            if relation_departed_key not in self.charm.app_peer_data:
                 logger.info(
                     "Deferring, must wait for relation departed hook to decide if relation should be removed."
                 )
                 event.defer()
                 return
 
-            if not json.loads(self.charm.app_peer_data["remove_users"]):
+            if not json.loads(self.charm.app_peer_data[relation_departed_key]):
                 logger.info(
                     "Relation broken event occurring due to scale down, do not proceed to remove users."
                 )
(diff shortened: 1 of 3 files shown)


lib/charms/mongodb/v0/mongodb_provider.py
```diff
@@ -79,6 +82,30 @@ def __init__(self, charm: CharmBase, substrate="k8s", relation_name: str = "data
...
+
+        # assume relation departed is due to manual removal of relation.
+        remove_users = True
+
+        # check if relation departed is due to current unit being removed. (i.e. scaling down the
+        # application.)
+        if event.departing_unit == self.charm.unit:
+            remove_users = False
+
+        self.charm.app_peer_data["remove_users"] = json.dumps(remove_users)
```
delgod:
```suggestion
        relation_departed = True

        # check if relation departed is due to current unit being removed. (i.e. scaling down the
        # application.)
        if event.departing_unit == self.charm.unit:
            relation_departed = False

        self.charm.app_peer_data[f"relation_{event.relation.id}_departed"] = json.dumps(relation_departed)
```

lib/charms/mongodb/v0/mongodb_provider.py
```diff
@@ -111,6 +138,21 @@ def _on_relation_event(self, event):
...
+                    "Deferring, must wait for relation departed hook to decide if relation should be removed."
+                )
+                event.defer()
+                return
+
+            if not json.loads(self.charm.app_peer_data["remove_users"]):
+                logger.info(
+                    "Relation broken event occurring due to scale down, do not proceed to remove users."
+                )
+                return
```
delgod:
```suggestion
            relation_departed_key = f"relation_{event.relation.id}_departed"
            if relation_departed_key not in self.charm.app_peer_data:
                logger.info(
                    "Deferring, must wait for relation departed hook to decide if relation should be removed."
                )
                event.defer()
                return

            if not json.loads(self.charm.app_peer_data[relation_departed_key]):
                logger.info(
                    "Relation broken event occurring due to scale down, do not proceed to remove users."
                )
                return
```

delgod (changes requested):

lib/charms/mongodb/v0/mongodb_provider.py
```diff
@@ -111,6 +140,22 @@ def _on_relation_event(self, event):
...
+            # removal if the relation has been removed.
+            relation_departed_key = f"relation_{event.relation.id}_departed"
+            if relation_departed_key not in self.charm.app_peer_data:
+                logger.info(
+                    "Deferring, must wait for relation departed hook to decide if relation should be removed."
+                )
+                event.defer()
+                return
+
+            if not json.loads(self.charm.app_peer_data[relation_departed_key]):
```
dmitry-ratushnyy: nits:
Maybe we can have a function "is_scalind_down" that will do this check? so the if will look like

if is_scaling_down():
   logger.info(...)
   return

MiaAltieri: Great idea, especially since multiple relation interfaces will implement this. I will do this right now

delgod (dismissed):

lib/charms/mongodb/v0/mongodb_provider.py
```diff
@@ -79,6 +82,16 @@ def __init__(self, charm: CharmBase, substrate="k8s", relation_name: str = "data
             self.database_provides.on.database_requested, self._on_relation_event
         )
 
+    def _on_relation_departed(self, event):
```
dmitry-ratushnyy: Maybe we can put this into charm.py 😄 

Reasoning:
There is no specific action taken in the provider related code. All actions are done in charm.py

MiaAltieri: great feedback! thank you

dmitry-ratushnyy (approved):

delgod (approved):

PR #265, opened 2023-10-11
[DPE-2701] POC Finish Removal / Scale down of sharded components

MiaAltieri:
## Issues
1. Juju 3.1.5 has [known bug](https://bugs.launchpad.net/juju/+bug/2029285) where secrets are removed before the storage detaches
2. relation broken executes when shards/config-server are scaling down
3. users can remove shards / config server application and break cluster

## Solution
1. Bump to Juju 3.1.6
2. Differentiate between scaling down / relation removal in relation-departed
3.1 when user removes config-server check if currently connected to shards, if so do not detach storage
3.2 when user removes shard check if currently connected to config-server, if so wait for storage to drain

## Testing
```

# relate shards

# scale down config-server and verify units don’t get (incorrectly) drained
juju add-unit config-server-one
juju remove-unit  config-server-one/0

# verify updated config
juju ssh config-server-one/1
charmed-mongodb.mongosh <URI>

# scale down  shard and verify units don’t get drained
juju add-unit shard-one
juju remove-unit  shard-one/0

# verify updated config
juju ssh config-server-one/1
charmed-mongodb.mongosh <URI>

# verify removing shard application waits to drain
juju remove-application shard-one/0

# verify removing config server goes into error
juju remove-application config-server
```


lib/charms/mongodb/v0/shards_interface.py
```diff
@@ -126,6 +130,12 @@ def pass_hook_checks(self, event: EventBase) -> bool:
             event.defer()
             return False
 
+        # adding/removing shards while a backup/restore is in progress can be disastrous
+        pbm_status = self.charm.backups._get_pbm_status()
```
dmitry-ratushnyy: Is it possible to avoid using of private methods of backups?  😄 

I see that here we are checking status of charm, right?

src/charm.py
```diff
@@ -419,7 +419,7 @@ def _on_relation_joined(self, event: RelationJoinedEvent) -> None:
         # app relations should be made aware of the new set of hosts
         try:
             self.client_relations.update_app_relation_data()
-            self.shard_relations._update_mongos_hosts()
+            self.config_server._update_mongos_hosts()
```
dmitry-ratushnyy: If we are using `_update_mongos_hosts()` outside `ShardingProvider` I think it is better to rename it to `update_mongos_hosts` because "_" prefix is in general for private methods\fields

MiaAltieri: Ah, gotcha! thank you!

lib/charms/mongodb/v0/shards_interface.py
MiaAltieri: > Is it possible to avoid using of private methods of backups? 😄

This makes me think. We actually use `_get_pbm_status` within `src/charm.py`, perhaps I should make this function `public`. What do you think?

> I see that here we are checking status of charm, right?

No we are checking the status of backups

dmitry-ratushnyy: > What do you think?

Yeah, I think we can mark it as public if it is used outside backups

MiaAltieri: Thanks for the guidiance, I am on it!

lib/charms/mongodb/v0/shards_interface.py
```diff
@@ -126,6 +130,12 @@ def pass_hook_checks(self, event: EventBase) -> bool:
             event.defer()
             return False
 
+        # adding/removing shards while a backup/restore is in progress can be disastrous
+        pbm_status = self.charm.backups.get_pbm_status()
+        if isinstance(pbm_status, MaintenanceStatus):
+            event.defer("Cannot add/remove shards while a backup/restore is in progress.")
```
dmitry-ratushnyy: I think here should be 
`logger.debug(message)` in addition to `event.defer()` because `event.defer` does not accept any params 😄

lib/charms/mongodb/v0/shards_interface.py
```diff
@@ -126,6 +130,12 @@ def pass_hook_checks(self, event: EventBase) -> bool:
...
+            return
```
dmitry-ratushnyy: maybe be is should be `return False` instead of return (since we are claiming that we always return boolean)?

lib/charms/mongodb/v0/shards_interface.py
MiaAltieri: 🤦‍♀️ good catch, thank you Dmitry! I am going to leave it as an `info` though until we have a longer conversation about loggging

lib/charms/mongodb/v0/shards_interface.py
MiaAltieri: 🤦‍♀️ good catch, thank you Dmitry!

dmitry-ratushnyy:
@MiaAltieri What do you think about updating log level, for logs that are coming before calls to `event.defer` from `logger.info` to `logger.debug` (keeping `logger.errors` as-is)

Reasoning:
In general, this information is not important to our users and make only sense to check it  when something is going wrong

MiaAltieri:
@dmitry-ratushnyy
> What do you think about updating log level, for logs that are coming before calls to event.defer from logger.info to logger.debug (keeping logger.errors as-is)

I am somewhat against this. While I think that we should get a system to differentiate between `logger.info` and `logger.debug`, I think the system of changing them before `event.defers` doesn't perfectly address their differences. But I would be curious what other people think. (TBH I would need to collect my thoughts more carefully here as well before forumulating more of an opinion)

But I do agree with you that we need to create a standardisation for this. Perhaps we can talk to the experts of charming when/where each one is appropriate. But I think the correct place to have this conversation will be in Mattermost and that the changes should be made in a separate PR as we should be consistent in our approach across all code.

dmitry-ratushnyy:
@MiaAltieri yep, lets have a discussion. I will create a ticket for tracking it

dmitry-ratushnyy (dismissed):

dmitry-ratushnyy (dismissed):

lib/charms/mongodb/v1/shards_interface.py
```diff
@@ -353,28 +385,59 @@ def _on_relation_changed(self, event):
...
             logger.error("sharding interface not supported with config role=replication")
-            return
+            return False
 
         if not self.charm.is_role(Config.Role.SHARD):
+            logger.info("skipping %s is only be executed by shards", type(event))
+            return False
+
+        if not self.charm.db_initialised:
+            event.defer()
```
delgod, line 400: log message before defer?

lib/charms/mongodb/v1/shards_interface.py
```diff
@@ -138,15 +171,11 @@ def _on_relation_event(self, event):
...
         departed_relation_id = None
-        if type(event) is RelationBrokenEvent:
-            departed_relation_id = event.relation.id
+        if isinstance(event, RelationBrokenEvent):
+            departed_relation_id = self._proceed_on_broken_event(event)
+            if not departed_relation_id:
+                return
 
         try:
             logger.info("Adding shards not present in cluster.")
```
delgod, line 181:
```suggestion
            logger.info("Adding/removing shards not present in cluster.")
```

lib/charms/mongodb/v1/shards_interface.py
```diff
@@ -353,28 +385,59 @@ def _on_relation_changed(self, event):
...
             )
+            event.defer()
             return
 
-        if not self.charm.db_initialised:
-            event.defer()
+        # check if were scaling down and add a log message
+        if self.charm.is_scaling_down(event.relation.id):
+            logger.info(
+                "Relation broken event occurring due to scale down, do not proceed to remove users."
```
delgod, line 423:
```suggestion
                "Relation broken event occurring due to scale down, do not proceed to add/remove shards."
```

delgod (approved):

lib/charms/mongodb/v1/shards_interface.py
MiaAltieri: I can add this, but I don't think it should go here. Even though this function handles add/remove, relation broken events are only for removal.

dmitry-ratushnyy (approved):

PR #267, opened 2023-10-12
[DPE-2695] + [partial DPE-2700] POC Identify primary shard + Prevent removal of Primary shards

MiaAltieri:
## Issue
Primary shards are not identified, this can pose problems if removing a primary shard

## Solution
Identify primary shards and prevent their removal

## Follow up PR
move databases that use a primary shard to a different shard if removing primary (I don't want to add this to this PR as to reduce the complexity in the review)

## Testing
As this is a POC there are no integration tests added, instead tests were performed by hand 
```

# relate shards

# write data to shard-one
juju ssh config-server-one/0
charmed-mongodb.mongosh <URI>
use test_db
db.createCollection("cool_animals")
db.cool_animals.insertOne({"horses":"unicorn"})
sh.enableSharding("test_db")

# write data to shard-two
use test_db_2
db.createCollection("cool_animals")
db.cool_animals.insertOne({"horses":"unicorn"})
sh.enableSharding("test_db")
db.adminCommand( { movePrimary : "test_db_2", to : "shard-one" } )

# show both shards
use config 
show collections 
db.databases.find()

# exit
exit
exit

# remove shard and verify error in config server and forever wait in shard 
juju remove-relation config-server-one:config-server shard-two:sharding
juju status --watch 1s
```

### DIFF
--- a/lib/charms/mongodb/v0/mongos.py
+++ b/lib/charms/mongodb/v0/mongos.py
@@ -184,9 +184,9 @@ class MongosConnection:
             logger.error(cannot_remove_shard)
             raise NotReadyError(cannot_remove_shard)
 
-        if shard_name in self.get_primary_shards():
-            database_names = self.get_databases_for_primary_shard(shard_name)
-            cannot_remove_primary_shard = f"These databases: {', '.join(database_names)}, use Shard {shard_name} is a primary shard, cannot remove shard."
+        databases_using_shard_as_primary = self.get_databases_for_shard(shard_name)
+        if databases_using_shard_as_primary:
+            cannot_remove_primary_shard = f"These databases: {', '.join(databases_using_shard_as_primary)}, use Shard {shard_name} is a primary shard, cannot remove shard."
             logger.error(cannot_remove_primary_shard)
             raise RemovePrimaryShardError(cannot_remove_primary_shard)
 
@@ -216,8 +216,8 @@ class MongosConnection:
             f"Shard {shard_name} not in cluster, could not retrieve draining status"
         )
 
-    def get_primary_shards(self) -> Optional[List[str]]:
-        """Returns the primary shards of the cluster.
+    def get_databases_for_shard(self, primary_shard) -> Optional[List[str]]:
+        """Returns a list of databases using the given shard as a primary shard.
 
         In Sharded MongoDB clusters, mongos selects the primary shard when creating a new database
         by picking the shard in the cluster that has the least amount of data. This means that:
@@ -226,21 +226,15 @@ class MongosConnection:
         """
         databases_collection = self._get_databases_collection()
         if databases_collection is None:
-            logger.info("No data written to sharded cluster yet, no primary shards.")
-            return None
-
-        return databases_collection.distinct("primary")
-
-    def get_databases_for_primary_shard(self, primary_shard) -> Optional[List[str]]:
-        """Returns a list of databases using the given shard as a primary shard."""
-        databases_collection = self._get_databases_collection()
-        if databases_collection is None:
             return
 
         return databases_collection.distinct("_id", {"primary": primary_shard})
 
     def _get_databases_collection(self) -> collection.Collection:
-        """Returns the database collection if present."""
+        """Returns the databases collection if present.
+
+        The collection `databases` only gets created once data is written to the sharded cluster.
+        """
         config_db = self.client["config"]
         if "databases" not in config_db.list_collection_names():
             logger.info("No data written to sharded cluster yet.")
--- a/lib/charms/mongodb/v0/mongos.py
+++ b/lib/charms/mongodb/v0/mongos.py
@@ -4,7 +4,7 @@
 
 import logging
 from dataclasses import dataclass
-from typing import Dict, Optional, Set
+from typing import Dict, Optional, Set, List
 from urllib.parse import quote_plus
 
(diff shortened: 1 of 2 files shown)


lib/charms/mongodb/v0/mongos.py
```diff
@@ -235,11 +216,37 @@ def _is_shard_draining(self, shard_name: str) -> bool:
...
+    def get_primary_shards(self) -> Optional[List[str]]:
+        """Returns the primary shards of the cluster.
+
+        In Sharded MongoDB clusters, mongos selects the primary shard when creating a new database
+        by picking the shard in the cluster that has the least amount of data. This means that:
+        1. There can be multiple primary shards in a cluster.
+        2. Until there is data written to the cluster there is effectively no primary shard.
+        """
+        databases_collection = self._get_databases_collection()
+        if databases_collection is None:
```
MiaAltieri, line 228: I know what you're thinking.... change this to `if databases_collection:` but this isn't possible due to
```
NotImplementedError: Collection objects do not implement truth value testing or bool(). Please compare with None instead: collection is not None	
```

lib/charms/mongodb/v0/mongos.py
```diff
@@ -183,37 +184,17 @@ def remove_shard(self, shard_name: str) -> None:
...
-        # possible with mongosh.
-        primary_shard = self.get_primary_shard()
-        if primary_shard:
-            # TODO Future PR, support removing Primary Shard if there are no unsharded collections
-            # on it. All sharded collections should perform `MovePrimary`
-            cannot_remove_primary_shard = (
-                f"Shard {shard_name} is the primary shard, cannot remove."
-            )
+        if shard_name in self.get_primary_shards():
+            database_names = self.get_databases_for_primary_shard(shard_name)
```
dmitry-ratushnyy: what happens if database names will be `None` in this case?
get_databases_for_primary_shard can return None

MiaAltieri: It will never be None, the definition of a primary shard is that it has databases. 

If ` self.get_databases_for_primary_shard(shard_name)` is `None` then, `shard_name in self.get_primary_shards()` is `False`

But that makes me think, It might be better to do:
```
if self.get_databases_for_primary_shard(shard_name):
        ....
````

**Review by delgod — DISMISSED** [review]

File: lib/charms/mongodb/v0/mongos.py [hunk 3]
```diff
@@ -235,11 +216,37 @@ def _is_shard_draining(self, shard_name: str) -> bool:
...
+        databases_collection = self._get_databases_collection()
+        if databases_collection is None:
+            return
+
+        return databases_collection.distinct("_id", {"primary": primary_shard})
+
+    def _get_databases_collection(self) -> collection.Collection:
+        """Returns the database collection if present."""
+        config_db = self.client["config"]
+        if "databases" not in config_db.list_collection_names():
```
dmitry-ratushnyy, line 239: why not use `config_db.get_collection("databasess")`?

MiaAltieri: because it is not guaranteed that the collection exists. The collection only gets created once a user writes data

lib/charms/mongodb/v0/mongos.py
```diff
@@ -235,11 +216,37 @@ def _is_shard_draining(self, shard_name: str) -> bool:
...
+            logger.info("No data written to sharded cluster yet.")
+            return None
+
+        return config_db["databases"]
```
dmitry-ratushnyy, line 243: here we are returning collection by accessing `config_db["databases"]`  just curious why do we need a call to `config_db.list_collection_names()` before what

lib/charms/mongodb/v0/mongos.py
dmitry-ratushnyy: in what case maybe yes 😄

lib/charms/mongodb/v0/mongos.py
MiaAltieri: `config_db.list_collection_names()` - lists all the collections present in the `config` database
`config_db["databases"]` - retrieves the collection named `databases` (which is not guaranteed to be present, this collection only gets created once data gets written to the cluster. I can add this to the docstring.)

dmitry-ratushnyy: i am just curious is it possible to do something like

if "databases" in config_db without calling to `config_db.list_collection_names()` (since we are returning it like config_db['databases'].

If not - no problem 😄

MiaAltieri: Sadly no, since `config_db ` is not an iterator we cannot do `config_db['databases']`, I tried something like this at first and got:

```
  File "/var/lib/juju/agents/unit-config-server-one-0/charm/lib/charms/mongodb/v0/mongos.py", line 245, in get_primary_shards
    if "databases" not in config_db:
TypeError: argument of type 'Database' is not iterable
```

dmitry-ratushnyy: Thank you for the explanations :)

MiaAltieri: of course! Thank you for your careful attention to detail!

dmitry-ratushnyy (approved):

delgod (approved):

PR #271, opened 2023-10-13
Fix libs

MiaAltieri:
## Issue
previous PR to fix libraries was misguided. More changes needed
- bumping LIB API requires adding a new director
- ^ requires updating paths


.github/workflows/ci.yaml
```diff
@@ -88,7 +88,7 @@ jobs:
     needs:
       - lint
       - unit-test
-      - lib-check
+      # - lib-check
```
dmitry-ratushnyy, line 91: just curious - why this is commented?

MiaAltieri: This is commented because it is a required check for CI, but for the sake of publishing libs for the **first time** which we are doing here, we cannot perform this check

dmitry-ratushnyy: Thank you for the explanation

dmitry-ratushnyy (approved):

delgod (approved):

PR #273, opened 2023-10-16
[DPE-2700] Move primary on shard removal

MiaAltieri:
## Issue
Code currently blocks the removal of primary

## Solution
permit removal of shard and movePrimary on database that use that shard as a primary shard


.github/workflows/ci.yaml
```diff
@@ -50,23 +50,20 @@ jobs:
...
-  #     - name: Checkout
-  #       uses: actions/checkout@v3
-  #       with:
-  #         fetch-depth: 0
-  #     - name: Check libs
-  #       uses: canonical/charming-actions/check-libraries@2.4.0
-  #       with:
-  #         credentials: "${{ secrets.CHARMHUB_TOKEN }}" # FIXME: current token will expire in 2023-07-04
-  #         github-token: "${{ secrets.GITHUB_TOKEN }}"
+  lib-check:
```
MiaAltieri, line 53: now that lib code has been merged and published by hand we can comment this back in

lib/charms/mongodb/v1/shards_interface.py
```diff
@@ -255,11 +245,6 @@ def remove_shards(self, departed_shard_id):
                 mongo.remove_shard(shard)
                 logger.info("Shard: %s, is now draining", shard)
 
-                if shard in mongo.get_shard_members():
```
MiaAltieri, line 258: this is now handled in mongos.py

lib/charms/mongodb/v1/shards_interface.py
```diff
@@ -253,12 +243,6 @@ def remove_shards(self, departed_shard_id):
                 self.charm.unit.status = MaintenanceStatus(f"Draining shard {shard}")
                 logger.info("Attempting to removing shard: %s", shard)
                 mongo.remove_shard(shard)
-                logger.info("Shard: %s, is now draining", shard)
```
MiaAltieri, line 256: we can remove this since `remove_shard` only succeeds on complete & successful removal/drainage

lib/charms/mongodb/v1/mongos.py
```diff
@@ -171,30 +175,44 @@ def remove_shard(self, shard_name: str) -> None:
...
         Raises:
             ConfigurationError, OperationFailure, NotReadyError,
-            RemovePrimaryShardError
+            NotEnoughSpaceError
         """
-        sc_status = self.client.admin.command("listShards")
-
         # It is necessary to call removeShard multiple times on a shard to guarantee removal.
         # Allow re-removal of shards that are currently draining.
+        sc_status = self.client.admin.command("listShards")
```
delgod, line 191: Can we also add a mandatory pre-flight check: Ensure the Balancer Process is Enabled?
https://www.mongodb.com/docs/manual/tutorial/remove-shards-from-cluster/#ensure-the-balancer-process-is-enabled

lib/charms/mongodb/v1/mongos.py
```diff
@@ -171,30 +175,44 @@ def remove_shard(self, shard_name: str) -> None:
...
+        remaining_chunks = self._retrieve_remaining_chunks(removal_info)
+        if remaining_chunks:
+            logger.info("Waiting for all chunks to be drained from %s.", shard_name)
+            raise NotDrainedError()
+
+        # MongoDB docs says to movePrimary only after all chunks have been drained from the shard.
+        logger.info("All chunks drained from shard: %s", shard_name)
+        databases_using_shard_as_primary = self.get_databases_for_shard(shard_name)
+        if not databases_using_shard_as_primary:
+            return
```
delgod: Should we call `removeShard` again here?

delgod (changes requested):

lib/charms/mongodb/v1/mongos.py
MiaAltieri: IMO no. We already check if we have remaining chunks after calling` removeShard`. If we do, then we raise `NotDrainedError` and then `defer`, once the deferred event runs again, we will call ` removeShard` again.

FYI draining shards can take several days depending how much data is present, so IMO deferring isn't detrimental to performance here

lib/charms/mongodb/v1/mongos.py
MiaAltieri: yup

lib/charms/mongodb/v1/mongos.py
```diff
@@ -286,3 +302,84 @@ def _log_removal_info(self, removal_info, shard_name):
...
+            # completes.
+            logger.info(
+                "Moving primary on %s database to new primary: %s. Do NOT write to %s database.",
+                database_name,
+                new_shard,
+                database_name,
+            )
+            # This command does not return until MongoDB completes moving all data. This can take
+            # a long time.
+            self.client.admin.command("movePrimary", database_name, to=new_shard)
```
dmitry-ratushnyy, line 371: Just curious: Why won't we put charm in a Maintenance status here, if the operation can take a long time?

MiaAltieri: `mongos.py` has no reference to `self.charm`. But the function that calls this goes into maintenance status:
[config server side](https://github.com/canonical/mongodb-operator/blob/7837e28be6f615e8017e24518542012f1c6405a8/lib/charms/mongodb/v1/shards_interface.py#L243)

And this is similarly reported on: 
[shard side](https://github.com/canonical/mongodb-operator/blob/7837e28be6f615e8017e24518542012f1c6405a8/lib/charms/mongodb/v1/shards_interface.py#L412)

dmitry-ratushnyy: Thank you for explanation 😄

dmitry-ratushnyy (dismissed):

delgod (dismissed):

dmitry-ratushnyy (approved):

delgod (approved):

PR #275, opened 2023-10-19
[DPE-2763] POC Status reporting shard side

MiaAltieri:
## Issue
Statuses are not properly reported for shards in update status hook

## Solution
Report statuses

## Testing
```
# deploy charms 

# speed up frequency of status checks
juju model-config update-status-hook-interval=10s

# monitor `juju status --watch 1s`
Unit                  Workload  Agent  Machine  Public address  Ports            Message
config-server-one/0*  active    idle   2        10.61.64.75     27017-27018/tcp  Primary
shard-one/0*          blocked   idle   0        10.61.64.126    27017/tcp        missing relation to config server
shard-two/0*          blocked   idle   1        10.61.64.216    27017/tcp        missing relation to config server

# relate application

# monitor `juju status --watch 1s`
Unit                  Workload  Agent  Machine  Public address  Ports            Message
config-server-one/0*  active    idle   2        10.61.64.75     27017-27018/tcp  Primary
shard-one/0*          active    idle   0        10.61.64.126    27017/tcp        Shard connected to config-server: config-server-one
shard-two/0*          active    idle   1        10.61.64.216    27017/tcp        Shard connected to config-server: config-server-one

# remove a relation to shard
juju remove-relation config-server-one:config-server shard-two:sharding

# monitor `juju status --watch 1s`
config-server-one/0*  active    idle   0        10.61.64.50     27017-27018/tcp  Primary
shard-one/0*          active    idle   1        10.61.64.235    27017/tcp        Shard connected to config-server: config-server-one
shard-two/0*          active    idle   2        10.61.64.128    27017/tcp        Shard drained from cluster, ready for removal

# add non supported relation
cd mongodb-operator/tests/integration/relation_tests/new_relations/application-charm
charmcraft pack
juju deploy ./*charm
juju integrate shard-one application

# monitor `juju status --watch 1s`
application/0*        active    idle   3        10.61.64.203
config-server-one/0*  active    idle   0        10.61.64.43     27017-27018/tcp
shard-one/0*          blocked   idle   1        10.61.64.31     27017/tcp        Sharding roles do not support mongodb_client interface.
shard-two/0*          active    idle   2        10.61.64.239    27017/tcp
```


lib/charms/mongodb/v1/shards_interface.py
```diff
@@ -564,13 +611,48 @@ def _update_relation_data(self, relation_id: int, data: dict) -> None:
...
+
     def has_config_server(self) -> bool:
         """Returns True if currently related to config server."""
         return len(self.charm.model.relations[self.relation_name]) > 0
 
-    def get_related_config_server(self) -> List[str]:
+    def get_related_config_server(self) -> str:
         """Returns the related config server."""
-        return [rel.app.name for rel in self.charm.model.relations[self.relation_name]]
+        config_server = [rel.app.name for rel in self.charm.model.relations[self.relation_name]]
```
dmitry-ratushnyy: is this possible that this list will be empty?

lib/charms/mongodb/v1/shards_interface.py
```diff
@@ -459,6 +468,44 @@ def wait_for_draining(self, mongos_hosts: List[str]):
 
                 break
 
+    def get_shard_status(self):
```
dmitry-ratushnyy: nits: return type is missing here

lib/charms/mongodb/v1/helpers.py
```diff
@@ -201,6 +201,35 @@ def generate_keyfile() -> str:
     return "".join([secrets.choice(choices) for _ in range(1024)])
 
 
+def prioritise_statuses(mongodb_status, pbm_status, shard_status, config_server_status):
```
dmitry-ratushnyy: nits: return type is missing here

lib/charms/mongodb/v1/shards_interface.py
```diff
@@ -564,13 +611,48 @@ def _update_relation_data(self, relation_id: int, data: dict) -> None:
...
         return len(self.charm.model.relations[self.relation_name]) > 0
 
-    def get_related_config_server(self) -> List[str]:
+    def get_related_config_server(self) -> str:
         """Returns the related config server."""
-        return [rel.app.name for rel in self.charm.model.relations[self.relation_name]]
+        config_server = [rel.app.name for rel in self.charm.model.relations[self.relation_name]]
+
+        # metadata.yaml prevents having multiple config servers
+        return config_server[0]
```
dmitry-ratushnyy: is there any possible issues if we return "wrong" config sever? Meaning: we always return first one, no matter what

lib/charms/mongodb/v1/shards_interface.py
dmitry-ratushnyy: because we always return first element of the list, why do we need to build the list?
why not something like `self.charm.model.relations[self.relation_name][0].app.name`? 😄

lib/charms/mongodb/v1/mongodb_provider.py
```diff
@@ -82,6 +84,34 @@ def __init__(self, charm: CharmBase, substrate="k8s", relation_name: str = "data
             self.database_provides.on.database_requested, self._on_relation_event
         )
 
+    def pass_hook_checks(self) -> bool:
+        """Runs the pre-hooks checks for MongoDBProvider, returns True if all pass."""
+        if self.charm.is_role(Config.Role.SHARD) or self.charm.is_role(Config.Role.CONFIG_SERVER):
+            self.charm.unit.status = BlockedStatus(
+                "Sharding roles do not support mongodb_client interface."
```
Mehdi-Bendriss: This and other statuses should probably be externalized in a constant file

lib/charms/mongodb/v1/mongodb_provider.py
```diff
@@ -82,6 +84,34 @@ def __init__(self, charm: CharmBase, substrate="k8s", relation_name: str = "data
             self.database_provides.on.database_requested, self._on_relation_event
         )
 
+    def pass_hook_checks(self) -> bool:
+        """Runs the pre-hooks checks for MongoDBProvider, returns True if all pass."""
+        if self.charm.is_role(Config.Role.SHARD) or self.charm.is_role(Config.Role.CONFIG_SERVER):
+            self.charm.unit.status = BlockedStatus(
```
Mehdi-Bendriss: imo, this method should not apply statuses, but rather just stick to returning booleans, and when calling it - depending on the return value to call a method with the exhaustive list of conditions + status. 
i.e: 
```
if not self.pass_hook_checks():
    logger.info("Skipping %s: hook checks did not pass", type(event))
    self.apply_status_if_needed()
    return

self.clear_status()  # this only targets relation related statuses
```

lib/charms/mongodb/v1/mongodb_vm_legacy_provider.py
```diff
@@ -64,6 +66,16 @@ def _on_legacy_relation_created(self, event):
             )
             return
 
+        if self.charm.is_role(Config.Role.SHARD) or self.charm.is_role(Config.Role.CONFIG_SERVER):
```
Mehdi-Bendriss: This code block is similar to the one above, would probably be an argument in favor of externalizing rel related statuses in a separate method?

lib/charms/mongodb/v1/mongos.py
```diff
@@ -334,6 +334,36 @@ def _log_removal_info(self, removal_info, shard_name):
...
+            allow server time to start up.
+
+        Raises:
+            ConfigurationError, ConfigurationError, OperationFailure
+        """
+        try:
+            for attempt in Retrying(stop=stop_after_delay(60), wait=wait_fixed(3)):
+                with attempt:
+                    # The ping command is cheap and does not require auth.
+                    self.client.admin.command("ping")
```
Mehdi-Bendriss, line 352: Des this call when ping failing?

lib/charms/mongodb/v1/shards_interface.py
```diff
@@ -349,6 +354,10 @@ def _on_relation_changed(self, event):
             logger.info("Skipping relation joined event: hook checks re not passed")
             return
 
+        # if re-using an old shard, re-set drained flag.
+        self.charm.app_peer_data["drained"] = json.dumps(False)
```
Mehdi-Bendriss: nice

lib/charms/mongodb/v1/shards_interface.py
```diff
@@ -459,6 +468,44 @@ def wait_for_draining(self, mongos_hosts: List[str]):
...
+            return BlockedStatus("missing relation to config server")
+
+        if not self.model.get_relation(self.relation_name) and self.charm.drained:
+            return ActiveStatus("Shard drained from cluster, ready for removal")
+
+        if not self._is_mongos_reachable():
+            return BlockedStatus("Config server unreachable")
+
+        if not self._is_added_to_cluster():
+            self.charm.unit.status = MaintenanceStatus("Adding shard to config-server")
```
Mehdi-Bendriss: is applying the status here and **not** returning  expected?

src/charm.py
```diff
@@ -574,20 +575,25 @@ def _on_update_status(self, event: UpdateStatusEvent):
...
-            self.unit.status = pbm_status
+        shard_status = self.shard.get_shard_status() if self.is_role(Config.Role.SHARD) else None
+        config_server_status = (
+            self.config_server.get_config_server_status()
+            if self.is_role(Config.Role.CONFIG_SERVER)
+            else None
+        )
+        pbm_status = (
+            self.backups.get_pbm_status() if self.model.get_relation(S3_RELATION) else None
+        )
```
Mehdi-Bendriss: would it make sense to put all these in a `get_status()` or `compute_unit_status()` method? where the priorisation also happens?

Mehdi-Bendriss (commented):
Great work Mia! just a few points / questions to clarify

lib/charms/mongodb/v1/mongodb_provider.py
MiaAltieri: I am happy to do that, but i think it would end up having the function copy pasted, just with/without status. I personally don't like this, but if you feel strongly about that implementation/style I will gladly do it :)

lib/charms/mongodb/v1/mongodb_provider.py
MiaAltieri: I think you bring up a good point to have status messages as constants. This would effect multiple files throughout the charm (and require its own library) so I think that this is out of scope for this PR, but I made an issue for it on [the Jira backlog.](https://warthogs.atlassian.net/browse/DPE-2847)

Note to self:
It isn't super straightforward, since we have another constants file that is outside of the shared lib `src/charm.py`, so it might be worthwhile to dissect that file and create two constant files, one just for VM/k8s charm specific code and one for the remaining libs.

lib/charms/mongodb/v1/mongos.py
MiaAltieri: When ping fails an exception is raised :)

lib/charms/mongodb/v1/shards_interface.py
MiaAltieri: good catch! thank you! 🙇

src/charm.py
MiaAltieri: Great idea! Since they use `self` I'll move `prioritise_status` to the charm code

lib/charms/mongodb/v1/mongodb_vm_legacy_provider.py
MiaAltieri: This is an idea I am in favour of, we have many checks like this throughout the code. Let me think of a nice way to do this

lib/charms/mongodb/v1/shards_interface.py
MiaAltieri: No, since it is impossible to be related to multiple config-servers. Juju prevents this since we declare the number of possible config servers in `metadata.yaml`

lib/charms/mongodb/v1/shards_interface.py
MiaAltieri: yes good point! updating

MiaAltieri: ah, we should do some error checking, so let me add that, then add your suggestion

lib/charms/mongodb/v1/mongodb_vm_legacy_provider.py
MiaAltieri: think I got it :)

Mehdi-Bendriss (dismissed):

dmitry-ratushnyy (dismissed):

Mehdi-Bendriss (dismissed):

dmitry-ratushnyy (approved):

Mehdi-Bendriss (approved):

PR #285, opened 2023-10-27
update ci/workflows for snap error

MiaAltieri:
## Issue
CI has an occasional issue with snap-install, which is fixed in new data-platform workflows

## Solution
Update workflows + deps

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -72,11 +72,11 @@ description = Run charm integration tests
 pass_env =
     {[testenv]pass_env}
     CI
-    CI_PACKED_CHARMS
 deps =
     pytest
     juju==3.2.0.1
     pytest-operator
+    git+https://github.com/canonical/data-platform-workflows@v5.1.2\#subdirectory=python/pytest_plugins/pytest_operator_cache
     pytest-mock
     -r {tox_root}/requirements.txt
 commands =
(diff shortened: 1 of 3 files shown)


tox.ini
```diff
@@ -72,11 +72,11 @@ description = Run charm integration tests
 pass_env =
     {[testenv]pass_env}
     CI
-    CI_PACKED_CHARMS
 deps =
     pytest
     juju==3.2.0.1
     pytest-operator
+    git+https://github.com/canonical/data-platform-workflows@v5.1.2\#subdirectory=python/pytest_plugins/pytest_operator_cache
```
dmitry-ratushnyy, line 79: is this comment required? 😄

tox.ini
```diff
@@ -87,11 +87,11 @@ description = Run high availability integration tests
 pass_env =
     {[testenv]pass_env}
     CI
-    CI_PACKED_CHARMS
 deps =
     pytest
     juju==3.2.0.1
     pytest-mock
+    git+https://github.com/canonical/data-platform-workflows@v5.1.2\#subdirectory=python/pytest_plugins/pytest_operator_cache
```
dmitry-ratushnyy, line 94: is this comment (if it is a comment) required?

MiaAltieri: No, it is not a comment which is why we have `\` :) 

I copied this from [these instructions](https://github.com/canonical/data-platform-workflows/blob/v4.0.0/.github/workflows/build_charms_with_cache.md#step-2-install-plugin-for-pytest-operator) with the help of @carlcsaposs-canonical

dmitry-ratushnyy (approved):

delgod (approved):

PR #290, opened 2023-10-30
[DPE-2776] Status reporting config server side

MiaAltieri:
## Issue
No statuses reported for config server

## Solution
Report statuses on config sever side

### DIFF
--- a/lib/charms/mongodb/v1/mongos.py
+++ b/lib/charms/mongodb/v1/mongos.py
@@ -163,11 +163,6 @@ class MongosConnection:
         shard_hosts = [f"{host}:{shard_port}" for host in shard_hosts]
         shard_hosts = ",".join(shard_hosts)
         shard_url = f"{shard_name}/{shard_hosts}"
-        # TODO Future PR raise error when number of shards currently adding are higher than the
-        # number of secondaries on the primary shard. This will be challenging, as there is no
-        # MongoDB command to retrieve the primary shard. Will likely need to be done via
-        # mongosh
-
         if shard_name in self.get_shard_members():
             logger.info("Skipping adding shard %s, shard is already in cluster", shard_name)
             return
@@ -444,3 +439,13 @@ class MongosConnection:
                 candidate_free_space = current_free_space
 
         return (candidate_shard, candidate_free_space)
+
+    def get_draining_shards(self) -> List[str]:
+        """Returns a list of the shards currently draining."""
+        sc_status = self.client.admin.command("listShards")
+        draining_shards = []
+        for shard in sc_status["shards"]:
+            if shard.get("draining", False):
+                draining_shards.append(shard["_id"])
+
+        return draining_shards
--- a/lib/charms/mongodb/v1/mongos.py
+++ b/lib/charms/mongodb/v1/mongos.py
@@ -21,7 +21,7 @@ LIBAPI = 1
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 2
+LIBPATCH = 3
 
 # path to store mongodb ketFile
 logger = logging.getLogger(__name__)
--- a/lib/charms/mongodb/v1/shards_interface.py
+++ b/lib/charms/mongodb/v1/shards_interface.py
@@ -32,6 +32,7 @@ from ops.framework import Object
 from ops.model import (
     ActiveStatus,
     BlockedStatus,
+    ErrorStatus,
     MaintenanceStatus,
     StatusBase,
     WaitingStatus,
@@ -268,7 +269,7 @@ class ShardingProvider(Object):
         for relation in self.charm.model.relations[self.relation_name]:
             self._update_relation_data(relation.id, {HOSTS_KEY: json.dumps(self.charm._unit_ips)})
 
-    def get_config_server_status(self):
+    def get_config_server_status(self) -> Optional[StatusBase]:
         """Returns the current status of the config-server."""
         if not self.charm.is_role(Config.Role.CONFIG_SERVER):
             logger.info("skipping status check, charm is not running as a shard")
@@ -285,7 +286,7 @@ class ShardingProvider(Object):
             return BlockedStatus(f"relation {REL_NAME} to shard not supported.")
 
         if not self.is_mongos_running():
-            return BlockedStatus("Internal mongos is not running.")
+            return ErrorStatus("Internal mongos is not running.")
 
         shard_draining = self.get_draining_shards()
         if shard_draining:
@@ -297,12 +298,10 @@ class ShardingProvider(Object):
 
         unreachable_shards = self.get_unreachable_shards()
         if unreachable_shards:
-            unreachable_shards = "".join(unreachable_shards)
-            return BlockedStatus(f"Shards {unreachable_shards} are unreachable.")
+            unreachable_shards = ", ".join(unreachable_shards)
+            return ErrorStatus(f"Shards {unreachable_shards} are unreachable.")
 
-        shards = self.get_related_shards()
-        shards = ", ".join(shards) if len(shards) < 5 else f"{len(shards)} shards"
-        return ActiveStatus(f"config-server connected to {shards}")
+        return ActiveStatus()
 
     def _update_relation_data(self, relation_id: int, data: dict) -> None:
         """Updates a set of key-value pairs in the relation.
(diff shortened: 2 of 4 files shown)


lib/charms/mongodb/v1/mongos.py
```diff
@@ -163,11 +163,6 @@ def add_shard(self, shard_name, shard_hosts, shard_port=Config.MONGODB_PORT):
         shard_hosts = [f"{host}:{shard_port}" for host in shard_hosts]
         shard_hosts = ",".join(shard_hosts)
         shard_url = f"{shard_name}/{shard_hosts}"
-        # TODO Future PR raise error when number of shards currently adding are higher than the
```
MiaAltieri, line 166: mongodb actually limits the amount of total migrations occuring so there is no need to do this

lib/charms/mongodb/v1/shards_interface.py
```diff
@@ -217,9 +217,6 @@ def add_shards(self, departed_shard_id):
         with MongosConnection(self.charm.mongos_config) as mongo:
             cluster_shards = mongo.get_shard_members()
             relation_shards = self._get_shards_from_relations(departed_shard_id)
-
-            # TODO Future PR, limit number of shards add at a time, based on the number of
```
MiaAltieri, line 221: mongodb actually limits the amount of total migrations occuring so there is no need to do this

lib/charms/mongodb/v1/shards_interface.py
```diff
@@ -325,6 +354,44 @@ def get_related_shards(self) -> List[str]:
         """Returns a list of related shards."""
         return [rel.app.name for rel in self.charm.model.relations[self.relation_name]]
 
+    def get_unreachable_shards(self) -> List[str]:
```
delgod, line 356: about `-> List[str]`, I don't see a `return unreachable_hosts` in this function.
also sometimes we return a bool instead of List

lib/charms/mongodb/v1/shards_interface.py
```diff
@@ -272,8 +269,40 @@ def update_mongos_hosts(self):
...
+            return None
+
+        if self.model.relations[LEGACY_REL_NAME]:
+            return BlockedStatus(f"relation {LEGACY_REL_NAME} to shard not supported.")
+
+        if self.model.relations[REL_NAME]:
+            return BlockedStatus(f"relation {REL_NAME} to shard not supported.")
+
+        if not self.is_mongos_running():
+            return BlockedStatus("Internal mongos is not running.")
```
delgod: Blocked is when we wait for action from the user. Maybe the Error status can be better.

lib/charms/mongodb/v1/shards_interface.py
```diff
@@ -272,8 +269,40 @@ def update_mongos_hosts(self):
...
+            shard_draining = ",".join(shard_draining)
+            return MaintenanceStatus(f"Draining shard {shard_draining}")
+
+        if not self.model.relations[self.relation_name]:
+            return BlockedStatus("missing relation to shard(s)")
+
+        unreachable_shards = self.get_unreachable_shards()
+        if unreachable_shards:
+            unreachable_shards = "".join(unreachable_shards)
+            return BlockedStatus(f"Shards {unreachable_shards} are unreachable.")
```
delgod: Blocked is when we wait for action from the user. Maybe the Error status can be better.

delgod (changes requested):

lib/charms/mongodb/v1/shards_interface.py
MiaAltieri: great idea

lib/charms/mongodb/v1/shards_interface.py
MiaAltieri: great idea

lib/charms/mongodb/v1/shards_interface.py
MiaAltieri: oh gosh you are totally correct 🤦 lets get some unit tests added ASAP

lib/charms/mongodb/v1/shards_interface.py
```diff
@@ -272,8 +269,40 @@ def update_mongos_hosts(self):
...
+
+        shards = self.get_related_shards()
+        shards = ", ".join(shards) if len(shards) < 5 else f"{len(shards)} shards"
+        return ActiveStatus(f"config-server connected to {shards}")
```
delgod: when app is fully started and functioning well, we shouldn't show anything in status. (except "Primary" status for primary)
```suggestion
        return ActiveStatus()
```

MiaAltieri: Ah, okay let me update this (and I will update this for the shard side as well)

lib/charms/mongodb/v1/shards_interface.py
```diff
@@ -272,8 +269,40 @@ def update_mongos_hosts(self):
             self._update_relation_data(relation.id, {HOSTS_KEY: json.dumps(self.charm._unit_ips)})
 
     def get_config_server_status(self):
```
dmitry-ratushnyy: nits: Looks like a return type is missing

lib/charms/mongodb/v1/shards_interface.py
```diff
@@ -325,6 +354,44 @@ def get_related_shards(self) -> List[str]:
         """Returns a list of related shards."""
         return [rel.app.name for rel in self.charm.model.relations[self.relation_name]]
 
+    def get_unreachable_shards(self) -> List[str]:
+        """Returns a list of unreable shard hosts."""
+        if not self.model.relations[self.relation_name]:
+            logger.info("shards are not reachable, none related to config-sever")
+            return False
```
dmitry-ratushnyy: why we are returning `False` if  return type List[str] ? 😄

lib/charms/mongodb/v1/shards_interface.py
```diff
@@ -325,6 +354,44 @@ def get_related_shards(self) -> List[str]:
...
+
+        unreachable_hosts = []
+        for shard_name in self.get_related_shards():
+            shard_hosts = self._get_shard_hosts(shard_name)
+            if not shard_hosts:
+                return False
```
dmitry-ratushnyy: why we are returning False if return type List[str] ? 😄

lib/charms/mongodb/v1/shards_interface.py
```diff
@@ -636,7 +702,10 @@ def _is_mongos_reachable(self) -> bool:
 
     def _is_added_to_cluster(self) -> bool:
         """Returns True if the shard has been added to the cluster."""
-        return json.loads(self.charm.app_peer_data.get("added_to_cluster", "False"))
+        if "added_to_cluster" not in self.charm.app_peer_data:
+            return False
+
+        return json.loads(self.charm.app_peer_data.get("added_to_cluster"))
```
dmitry-ratushnyy, line 708: Looks like return type is missing

lib/charms/mongodb/v1/shards_interface.py
MiaAltieri: covered in Mykolas comment

lib/charms/mongodb/v1/shards_interface.py
MiaAltieri: covered in Mykolas comment

lib/charms/mongodb/v1/shards_interface.py
MiaAltieri: I dont think so?

lib/charms/mongodb/v1/shards_interface.py
```diff
@@ -271,9 +269,39 @@ def update_mongos_hosts(self):
...
+        if shard_draining:
+            shard_draining = ",".join(shard_draining)
+            return MaintenanceStatus(f"Draining shard {shard_draining}")
+
+        if not self.model.relations[self.relation_name]:
+            return BlockedStatus("missing relation to shard(s)")
+
+        unreachable_shards = self.get_unreachable_shards()
+        if unreachable_shards:
+            unreachable_shards = "".join(unreachable_shards)
```
dmitry-ratushnyy: on line 293 `join` uses "," while here it is not. I think it is better to use comma here as well

delgod (dismissed):

lib/charms/mongodb/v1/shards_interface.py
```diff
@@ -271,9 +269,39 @@ def update_mongos_hosts(self):
...
             self._update_relation_data(relation.id, {HOSTS_KEY: json.dumps(self.charm._unit_ips)})
 
-    def get_config_server_status(self):
-        """TODO: Implement this function in a separate PR."""
-        return None
+    def get_config_server_status(self) -> StatusBase:
+        """Returns the current status of the config-server."""
+        if not self.charm.is_role(Config.Role.CONFIG_SERVER):
+            logger.info("skipping status check, charm is not running as a shard")
+            return None
```
dmitry-ratushnyy, line 276: nits: should it be Optional[StatusBase] as a return type  then?

lib/charms/mongodb/v1/shards_interface.py
delgod: can we fix this line?

dmitry-ratushnyy (approved):

dmitry-ratushnyy (approved):

delgod (approved):

PR #293, opened 2023-11-01
[DPE-2780] Test shard for unpermitted relations

MiaAltieri:
## Issue
No tests for relations that are not allowed for shards

## Solution
add tests for checking unsupported relations

## extras
create a dummy legacy application. This should NOT be used inside the legacy integration test suite as those tests should be done with production legacy charms


tests/integration/sharding_tests/test_sharding_relations.py
```diff
@@ -0,0 +1,139 @@
...
+        f"{CONFIG_SERVER_ONE_APP_NAME}:{CONFIG_SERVER_REL_NAME}",
+    )
+
+    multiple_config_server_rels_allowed = True
+    try:
+        await ops_test.model.integrate(
+            f"{SHARD_ONE_APP_NAME}:{SHARD_REL_NAME}",
+            f"{CONFIG_SERVER_TWO_APP_NAME}:{CONFIG_SERVER_REL_NAME}",
+        )
+    except JujuAPIError as e:
```
juditnovak: Why not to use `pytest.raises()` here instead? (https://docs.pytest.org/en/7.1.x/reference/reference.html#pytest.raises)
If I get it right, we want to check if the error occurs, with the expected message.

tests/integration/sharding_tests/test_sharding_relations.py
```diff
@@ -0,0 +1,139 @@
...
+        f"{SHARD_ONE_APP_NAME}:{LEGACY_RELATION_NAME}",
+        f"{LEGACY_APP_CHARM_NAME}:{LEGACY_RELATION_NAME}",
+    )
+
+    await ops_test.model.wait_for_idle(
+        apps=[SHARD_ONE_APP_NAME],
+        idle_period=20,
+        raise_on_blocked=False,
+        timeout=TIMEOUT,
+    )
```
juditnovak, line 140: If I get it right, the above two commands are only to clean up.
Why not to indicate that with a comment?
(Just to make it clear that we didn't expect any specific outcome to verify after setting up this state.)

tests/integration/sharding_tests/test_sharding_relations.py
```diff
@@ -0,0 +1,139 @@
...
+
+
+async def test_cannot_use_legacy_db_relation(ops_test: OpsTest) -> None:
+    """Verify that a shard cannot use the DB relation."""
```
juditnovak: Docsting update -- we're testing something different than in the test above :-)

juditnovak (changes requested):
In principal all good, but the `try...except` -> evaluation of variable set inside really doesn't look pretty in a test...

May I ask to pls-pls-pls change that one...?

dmitry-ratushnyy (approved):

juditnovak (approved):
LGTM, thx much for the changes :-)

tests/integration/sharding_tests/conftest.py
```diff
@@ -0,0 +1,33 @@
...
+    charm_path = "tests/integration/relation_tests/new_relations/application-charm"
+    charm = await ops_test.build_charm(charm_path)
+    return charm
+
+
+@pytest.fixture(scope="module")
+async def database_charm(ops_test: OpsTest):
+    """Build the database charm."""
+    charm = await ops_test.build_charm(".")
+    return charm
```
zmraul, line 33: Note for future housekeeping: you don't need to have these fixtures on each folder that uses them. They can just be declared on `tests/integration/conftest.py`.

MiaAltieri: let me do this on the next PR, it is a great suggestion

PR #298, opened 2023-11-03
[DPE-2781] Test replica relations to sharding interface

MiaAltieri:
## Issue
No tests for incorrect usage of MongoDB charm running as replica set

## Solution
Add tests that show:
- using a replica set as a config-server leads to blocked
- using a replica set as a shard leads to blocked

## Context
A mongodb charm can run as a Shard, Config-Server, or as a plain Replica set. Sharding relations are not supported with replica sets and so we must test for those and show informative status messages

## Extra Charm Code
Adding these tests highlighted holes in the code and so those mistakes have been patched


lib/charms/mongodb/v1/shards_interface.py
```diff
@@ -295,7 +319,7 @@ def get_config_server_status(self) -> Optional[StatusBase]:
             return BlockedStatus(f"relation {REL_NAME} to shard not supported.")
 
         if not self.is_mongos_running():
-            return ErrorStatus("Internal mongos is not running.")
```
MiaAltieri, line 298: it is actually not permitted to set ErrorStatus like we had before, it results in a broken charm and we get the following traceback:
```
  File "/var/lib/juju/agents/unit-shard-0/charm/venv/ops/model.py", line 2895, in status_set
    self._run('status-set', f'--application={is_app}', status, message)
  File "/var/lib/juju/agents/unit-shard-0/charm/venv/ops/model.py", line 2695, in _run
    raise ModelError(e.stderr)
ops.model.ModelError: ERROR invalid status "error", expected one of [maintenance blocked waiting active]
```

delgod (dismissed):

lib/charms/mongodb/v1/shards_interface.py
```diff
@@ -231,10 +239,25 @@ def add_shards(self, departed_shard_id):
...
                     mongo.add_shard(shard, shard_hosts)
                 except PyMongoError as e:
+                    # raise exception after trying to add the remaining shards, as to not prevent
+                    # adding other shards
                     logger.error("Failed to add shard %s to the config server, error=%r", shard, e)
-                    raise
+                    failed_to_add_shard = (e, shard)
+
+        if not failed_to_add_shard:
+            self.charm.unit.status = ActiveStatus("")
```
dmitry-ratushnyy, line 248: maybe we can move this 
```
self.charm.unit.status = ActiveStatus("")
return
```
under this 
```
mongo.add_shard(shard, shard_hosts)
```
in that case we won't need if and we can move the processing of error code under `except`.

What do you think?

MiaAltieri: You always offer such good advise when it comes to restructuring, thank you Dmitry. I like where your head is at!

But unfortunately in this scenario this will not work. It is possible that we may want to add multiple shards at a time (this line of code is part of this for loop `for shard in relation_shards - cluster_shards:`) so returning would prevent the other shards from being added.

dmitry-ratushnyy: Thank you for the explanation 😄

delgod (approved):

dmitry-ratushnyy (approved):

PR #309, opened 2023-11-28
[DPE-2992] Part 1 - Provide Mongos with keyfile

MiaAltieri:
## Issue
DPE-2992 is to start the mongos subordinate charm with the provided config server. In order for the config server and for mongos to be connected they both must have the same keyfile

## Solution
Provide that key file

## Follow up PRs
1. A follow up PR on mongos charm that uses this library
2. Updating the library to actually start the mongos service with the information of the config server
3. Integration tests

## Testing
```
juju deploy ./*charm --config role="config-server" config-server
juju deploy application
juju deploy mongos
juju integrate application mongos
juju integrate mongos:cluster mongodb:cluster
juju ssh mongos
sudo cat KEY_FILE_PATH
```

### DIFF
--- a/lib/charms/mongodb/v0/config_server_interface.py
+++ b/lib/charms/mongodb/v0/config_server_interface.py
@@ -8,7 +8,6 @@ shards.
 """
 import logging
 
-
 from ops.charm import CharmBase, EventBase
 from ops.framework import Object
 from ops.model import WaitingStatus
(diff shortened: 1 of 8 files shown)


lib/charms/mongodb/v1/helpers.py
```diff
@@ -267,3 +267,23 @@ def process_pbm_status(pbm_status: str) -> StatusBase:
         return WaitingStatus("waiting to sync s3 configurations.")
 
     return ActiveStatus()
+
+
+def add_args_to_env(var: str, args: str):
```
MiaAltieri, line 272: moved from `src/machine_helpers.py`

lib/charms/mongodb/v0/config_server_interface.py
```diff
@@ -0,0 +1,150 @@
+# Copyright 2023 Canonical Ltd.
```
MiaAltieri, line 1: When I did `charmcraft create-lib` it created this lib in `v0` ... ideally all the sharding stuff would be in `v1`. What I can do is after this lib is published I could bump the LIPAPI, but that feels messy...

dmitry-ratushnyy: I don't think that there is a way around it, most likely chamrcraft will complain about library version.

dmitry-ratushnyy (approved):

delgod (approved):

PR #313, opened 2023-12-05
Reset libpatch to fix publishing

MiaAltieri:
## Issue
Charm has not successfully released the recent changes to 6/edge. This means library `config_server_interface.py` has never been publish so lib patch must be reset to 0

## Solution
Change lib patch


dmitry-ratushnyy (approved):

delgod (approved):

MiaAltieri:
Out of Date

PR #324, opened 2024-01-04
[DPE-3083] Make integration tests work with existing cluster

dmitry-ratushnyy:
## Issue
It should be possible to run integration tests agains the existing cluster (if provided by a user)

## Solution
Update integration tests so they will support existing user clusters

## Future PR
Make these changes for sharding tests (if needed) + add these changes to the k8s charm


tests/integration/backup_tests/test_backups.py
```diff
@@ -105,7 +111,8 @@ async def test_blocked_incorrect_conf(ops_test: OpsTest) -> None:
 @pytest.mark.abort_on_fail
 async def test_ready_correct_conf(ops_test: OpsTest) -> None:
     """Verifies charm goes into active status when s3 config and creds options are correct."""
-    db_app_name = await helpers.app_name(ops_test)
+    db_app_name = await get_app_name(ops_test)
+    await helpers.set_credentials(ops_test, cloud="AWS")
```
Mehdi-Bendriss: Should `AWS` be parametrized here?

tests/integration/ha_tests/helpers.py
```diff
@@ -46,33 +52,33 @@ class ProcessRunningError(Exception):
     """Raised when a process is running when it is not expected to be."""
 
 
-def replica_set_client(replica_ips: List[str], password: str, app=APP_NAME) -> MongoClient:
+def replica_set_client(replica_ips: List[str], password: str, app_name=None) -> MongoClient:
```
Mehdi-Bendriss: I believe the `app_name` cannot be None

```suggestion
def replica_set_client(replica_ips: List[str], password: str, app_name: str) -> MongoClient:
```

tests/integration/ha_tests/helpers.py
```diff
@@ -46,33 +52,33 @@ class ProcessRunningError(Exception):
...
     hosts = ["{}:{}".format(replica_ip, PORT) for replica_ip in replica_ips]
     hosts = ",".join(hosts)
 
-    replica_set_uri = f"mongodb://operator:" f"{password}@" f"{hosts}/admin?replicaSet={app}"
+    replica_set_uri = f"mongodb://operator:" f"{password}@" f"{hosts}/admin?replicaSet={app_name}"
     return MongoClient(replica_set_uri)
 
 
-async def fetch_replica_set_members(replica_ips: List[str], ops_test: OpsTest):
+async def fetch_replica_set_members(replica_ips: List[str], ops_test: OpsTest, app_name=None):
```
Mehdi-Bendriss: Similar to the above comment, I don't think `app_name` can be `None`

tests/integration/tls_tests/helpers.py
```diff
@@ -43,7 +54,7 @@ async def mongo_tls_command(ops_test: OpsTest) -> str:
     )
 
 
-async def check_tls(ops_test: OpsTest, unit: ops.model.Unit, enabled: bool) -> bool:
+async def check_tls(ops_test: OpsTest, unit: ops.model.Unit, enabled: bool, app_name=None) -> bool:
```
Mehdi-Bendriss, line 57: Similar to the above, the app_name doesn't seem like it can be as an optional

Mehdi-Bendriss (commented):
Looks good, thank you! I just have a couple of remarks and should be good to go

tests/integration/backup_tests/test_backups.py
dmitry-ratushnyy: parametrised in what sense?

Mehdi-Bendriss: With regards to the cloud substrate (like you do in other tests). I guess my question is: is this test supposed to only run on AWS or other clouds

dmitry-ratushnyy: This one does not really care. We can improve the test to add "GCP", but I think it is better to do that a a part of separate PR (specifically for tests updates)

dmitry-ratushnyy: This one does not really care. We can improve the test to add "GCP", but I think it is better to do that a a part of separate PR (specifically for tests updates)

Mehdi-Bendriss: Sure, if SQA doesn't need to test on GCP - we can push it to another PR

MiaAltieri: @Mehdi-Bendriss testing with GCP + AWS is done in the other tests. When @delgod and I first designed this test suite we intended for the first tests to perform basic operations on AWS, and deemed it was not necessary to do both

However, it seems to me that this line of code is not necessary, as in. the previous test we set AWS credentials, so I will remove this line

tests/integration/ha_tests/helpers.py
MiaAltieri: agreed, what do you think @dmitry-ratushnyy ?

tests/integration/ha_tests/helpers.py
MiaAltieri: agreed, what do you think @dmitry-ratushnyy ?

tests/integration/ha_tests/helpers.py
dmitry-ratushnyy: Yes, "None" was added as tmp solution, we can remove it. Thank you 😄

tests/integration/ha_tests/helpers.py
dmitry-ratushnyy: Yes, "None" was added as tmp solution, we can remove it. Thank you 😄

tests/integration/tls_tests/helpers.py
MiaAltieri: @Mehdi-Bendriss from what I can tell `app_name` can be None here, since in the line that it is used (`73`) we make the call `mongod_tls_check = await mongo_tls_command(ops_test, app_name=app_name)` and `mongo_tls_command` does not require the option `app_name` and will infer it in the case of `None`

Mehdi-Bendriss (dismissed):

dmitry-ratushnyy:
LGTM

MiaAltieri (approved):

Mehdi-Bendriss (approved):

PR #334, opened 2024-01-11
[DPE-3182] mongodb remove user

MiaAltieri:
## Issue
When relation is broken between `mongos` charm and `config-server` charm the user created for the `mongos` application is still accessible 

## Solution
When relation is broken between mongos charm and config-sever process users 

## Tested by hand
```
 # deploy both charms
cd ~/mongos-operator
# juju add-model test-0
tox -e build
juju deploy ./*charm

cd ~/mongos-operator/tests/integration/application
charmcraft pack 
juju deploy ./*charm -n2

# get shared cluster up and running
cd ~/mongodb-operator
charmcraft pack
juju deploy ./*charm --config role="config-server" config-server -n2
juju deploy ./*charm --config role="shard" shard

# integrate
juju integrate mongos application
juju integrate config-server:config-server shard:sharding
juju integrate config-server:cluster mongos:cluster

# get URI
juju show-unit application/0
juju show-secret --reveal <secret URI>

# access user
juju ssh application/0
sudo charmed-mongodb.mongosh <URI>

# scale application and repeat access user step
juju remove-unit config-server/1
juju remove-unit application/0

# remove relation and  repeat access user step (except this time verify we cannot access the user)
juju remove-relation  config-server:cluster mongos:cluster
```

## Future PRs
1. [user int tests](https://warthogs.atlassian.net/browse/DPE-3292)
2. [have mongos sub-charm remove connection info for host application]()

### DIFF
--- a/lib/charms/mongodb/v0/config_server_interface.py
+++ b/lib/charms/mongodb/v0/config_server_interface.py
@@ -35,7 +35,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 2
+LIBPATCH = 3
 
 
 class ClusterProvider(Object):
@@ -54,8 +54,13 @@ class ClusterProvider(Object):
             charm.on[self.relation_name].relation_changed, self._on_relation_changed
         )
 
-        # TODO Future PRs handle scale down
-        # TODO Future PRs handle changing of units/passwords to be propagated to mongos
+        self.framework.observe(
+            charm.on[self.relation_name].relation_departed,
+            self.charm.check_relation_broken_or_scale_down,
+        )
+        self.framework.observe(
+            charm.on[self.relation_name].relation_broken, self._on_relation_broken
+        )
 
     def pass_hook_checks(self, event: EventBase) -> bool:
         """Runs the pre-hooks checks for ClusterProvider, returns True if all pass."""
@@ -96,6 +101,26 @@ class ClusterProvider(Object):
                     CONFIG_SERVER_DB_KEY: config_server_db,
                 },
             )
+
+    def _on_relation_broken(self, event) -> None:
+        # Only relation_deparated events can check if scaling down
+        departed_relation_id = event.relation.id
+        if not self.charm.has_departed_run(departed_relation_id):
+            logger.info(
+                "Deferring, must wait for relation departed hook to decide if relation should be removed."
+            )
+            event.defer()
+            return
+
+        if not self.pass_hook_checks(event):
+            logger.info("Skipping relation broken event: hook checks did not pass")
+            return
+
+        if not self.charm.proceed_on_broken_event(event):
+            logger.info("Skipping relation broken event, broken event due to scale down")
+            return
+
+        self.charm.client_relations.oversee_users(departed_relation_id, event)
 
     def update_config_server_db(self, event):
         """Provides related mongos applications with new config server db."""
@@ -157,6 +182,7 @@ class ClusterRequirer(Object):
         self.framework.observe(
             charm.on[self.relation_name].relation_changed, self._on_relation_changed
         )
+
         # TODO Future PRs handle scale down
 
     def _on_database_created(self, event) -> None:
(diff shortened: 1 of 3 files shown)


lib/charms/mongodb/v1/shards_interface.py
```diff
@@ -148,28 +148,6 @@ def pass_hook_checks(self, event: EventBase) -> bool:
 
         return True
 
-    def _proceed_on_broken_event(self, event) -> int:
```
MiaAltieri, line 151: move this to `src.charm` since it is used by other libs

lib/charms/mongodb/v0/config_server_interface.py
```diff
@@ -97,6 +102,26 @@ def _on_relation_changed(self, event) -> None:
...
+        if not self.charm.has_departed_run(departed_relation_id):
+            logger.info(
+                "Deferring, must wait for relation departed hook to decide if relation should be removed."
+            )
+            event.defer()
+            return
+
+        if not self.pass_hook_checks(event):
+            logger.info("Skipping relation broken event: hook checks did not pass")
+            return
```
dmitry-ratushnyy, line 117: why we are not deferring event here?

lib/charms/mongodb/v0/config_server_interface.py
```diff
@@ -97,6 +102,26 @@ def _on_relation_changed(self, event) -> None:
...
+
+        if not self.charm.proceed_on_broken_event(event):
+            logger.info("Skipping relation broken event, broken event due to scale down")
```
dmitry-ratushnyy, line 120: why we are not deferring event here?

dmitry-ratushnyy, line 120: why we are not deferring event here?

lib/charms/mongodb/v0/config_server_interface.py
MiaAltieri: please read the function `pass_hook_checks `, it defers when necessary

lib/charms/mongodb/v0/config_server_interface.py
MiaAltieri: please read the function `pass_hook_checks `, it defers when necessary

MiaAltieri: please read the function `pass_hook_checks `, it defers when necessary

dmitry-ratushnyy (approved):

src/charm.py
```diff
@@ -1319,10 +1319,32 @@ def set_scaling_down(self, event: RelationDepartedEvent) -> bool:
         # check if relation departed is due to current unit being removed. (i.e. scaling down the
         # application.)
         rel_departed_key = self._generate_relation_departed_key(event.relation.id)
-        scaling_down = json.dumps(event.departing_unit == self.unit)
-        self.unit_peer_data[rel_departed_key] = scaling_down
+        scaling_down = event.departing_unit == self.unit
+        self.unit_peer_data[rel_departed_key] = json.dumps(scaling_down)
         return scaling_down
 
+    def proceed_on_broken_event(self, event) -> int:
```
Mehdi-Bendriss, line 1326: The name of this method is not very intuitive imo. It's quite verbose, but what about something along the lines of (or something better): 
`def broken_relation_id_if_departed(self, event: ..) -> Optional[int]`

Mehdi-Bendriss (commented):
LGTM - I have one comment regarding a method name

src/charm.py
MiaAltieri: You have some points here. I think given the way that it is used in other areas of the code it should be kept as is, i.e. it is used in other parts of the code as:

```
    if not self.charm.proceed_on_broken_event(event):
            logger.info("Skipping relation broken event, broken event due to scale down")
            return
```

But I agree with what you are saying here. Instead I could make this function return a `boolean` instead and process the relation id in a separte function (or even in the calling function.) What do you think?

Mehdi-Bendriss: That sounds fair - this can be part of a next PR

Mehdi-Bendriss (approved):

PR #335, opened 2024-01-12
[DPE-3316] Fix issue with mongodb.log

dmitry-ratushnyy:
## Issue
mongodb.log was disabled on VM charm 

## Solution
Fix mogngodb.log creation


MiaAltieri (commented):
This looks good, thanks for catching this. I am wondering should we pass the `--logRotate` option as well?
@delgod

delgod (approved):

delgod:
> This looks good, thanks for catching this. I am wondering should we pass the `--logRotate` option as well?

we will have another task for log rotation later. prefer to keed it as it is for now

PR #336, opened 2024-01-15
[DPE-3314] remove secrets on broken

MiaAltieri:
## FYI
most of the work here is done on the `mongos` side, but some changes are needed in the libs that are owned by this charm

## Issue
Secretes associated with username, password, uri are still present on the `host-application` and `mongos sub-charm` relation after relation with `config-server` and `mongos sub-charm` is broken

## Solution
Call appropriate functions from mongos charm

## Future PRs
1. Implementation + Integration tests added on mongos side

### DIFF
--- a/lib/charms/mongodb/v0/config_server_interface.py
+++ b/lib/charms/mongodb/v0/config_server_interface.py
@@ -35,7 +35,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 3
+LIBPATCH = 4
 
 
 class ClusterProvider(Object):
@@ -182,8 +182,13 @@ class ClusterRequirer(Object):
         self.framework.observe(
             charm.on[self.relation_name].relation_changed, self._on_relation_changed
         )
-
-        # TODO Future PRs handle scale down
+        self.framework.observe(
+            charm.on[self.relation_name].relation_departed,
+            self.charm.check_relation_broken_or_scale_down,
+        )
+        self.framework.observe(
+            charm.on[self.relation_name].relation_broken, self._on_relation_broken
+        )
 
     def _on_database_created(self, event) -> None:
         if not self.charm.unit.is_leader():
@@ -227,6 +232,31 @@ class ClusterRequirer(Object):
             return
 
         self.charm.unit.status = ActiveStatus()
+
+    def _on_relation_broken(self, event) -> None:
+        # Only relation_deparated events can check if scaling down
+        departed_relation_id = event.relation.id
+        if not self.charm.has_departed_run(departed_relation_id):
+            logger.info(
+                "Deferring, must wait for relation departed hook to decide if relation should be removed."
+            )
+            event.defer()
+            return
+
+        if not self.charm.proceed_on_broken_event(event):
+            logger.info("Skipping relation broken event, broken event due to scale down")
+            return
+
+        self.charm.stop_mongos_service()
+        logger.info("Stopped mongos daemon")
+
+        if not self.charm.unit.is_leader():
+            return
+
+        logger.info("Database and user removed for mongos application")
+        self.charm.remove_secret(Config.Relations.APP_SCOPE, Config.Secrets.USERNAME)
+        self.charm.remove_secret(Config.Relations.APP_SCOPE, Config.Secrets.PASSWORD)
+        self.charm.remove_connection_info()
 
     # BEGIN: helper functions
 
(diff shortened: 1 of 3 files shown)


lib/charms/mongodb/v1/shards_interface.py
```diff
@@ -146,6 +146,18 @@ def pass_hook_checks(self, event: EventBase) -> bool:
             event.defer()
             return False
 
+        if isinstance(event, RelationBrokenEvent):
```
MiaAltieri, line 149: move this due to comments on previous PR

src/charm.py
```diff
@@ -1334,16 +1334,16 @@ def proceed_on_broken_event(self, event) -> int:
                 "Deferring, must wait for relation departed hook to decide if relation should be removed."
             )
             event.defer()
-            return
+            return False
```
dmitry-ratushnyy, line 1335: Could you please update the function return type as well (not it is "int")

lib/charms/mongodb/v1/shards_interface.py
```diff
@@ -159,9 +171,7 @@ def _on_relation_event(self, event):
 
         departed_relation_id = None
         if isinstance(event, RelationBrokenEvent):
-            departed_relation_id = self.charm.proceed_on_broken_event(event)
-            if not departed_relation_id:
-                return
+            departed_relation_id = event.relation.id
```
dmitry-ratushnyy, line 173: Looks like here if the event is NOT  `RelationBrokenEvent`  then `departed_relation_id` will be None, so maybe in that case we need to return from function earlier?

lib/charms/mongodb/v0/config_server_interface.py
```diff
@@ -228,6 +233,31 @@ def _on_relation_changed(self, event) -> None:
 
         self.charm.unit.status = ActiveStatus()
 
+    def _on_relation_broken(self, event) -> None:
```
Mehdi-Bendriss: nit: typed args

lib/charms/mongodb/v0/config_server_interface.py
```diff
@@ -228,6 +233,31 @@ def _on_relation_changed(self, event) -> None:
...
+        # Only relation_deparated events can check if scaling down
+        departed_relation_id = event.relation.id
```
Mehdi-Bendriss: this variable seems to be only used once on the subsequent condition - safe to drop and refer to `event.relation.id`

lib/charms/mongodb/v1/shards_interface.py
```diff
@@ -146,6 +146,18 @@ def pass_hook_checks(self, event: EventBase) -> bool:
             event.defer()
             return False
 
+        if isinstance(event, RelationBrokenEvent):
+            departed_relation_id = event.relation.id
```
Mehdi-Bendriss: same as previous comment

Mehdi-Bendriss (commented):
Good work! I only have a few minor comments

lib/charms/mongodb/v1/shards_interface.py
MiaAltieri: > Looks like here if the event is NOT RelationBrokenEvent then departed_relation_id will be None

That is correct. I am not 100% sure what you mean by:
> in that case we need to return from function earlier

But, we should **not** return if `departed_relation_id` is `None`

In this function we pass `departed_relation_id` to subsequent functions and those functions handle the cases in which `departed_relation_id` is `None`, namely:
```
            self.add_shards(departed_relation_id)
            self.remove_shards(departed_relation_id)
 ```
 
 does that answer your question? Or am I misunderstanding?

Comment (dmitry-ratushnyy) [reply lib/charms/mongodb/v1/shards_interface.py:173]: Why would we want to do his call ?

```
self.add_shards(None)
self.remove_shards(None)
```

Comment (MiaAltieri) [reply lib/charms/mongodb/v1/shards_interface.py:173]: Please read the functions `add_shards` and `remove_shards` 

ideally from those function signatures it should be obvious why we would want to do this. If it isn't obvious then there is a big problem in the readability of this file and that needs to be updated ASAP

File: src/charm.py [hunk 7]
```diff
@@ -1323,7 +1323,7 @@ def set_scaling_down(self, event: RelationDepartedEvent) -> bool:
         self.unit_peer_data[rel_departed_key] = json.dumps(scaling_down)
         return scaling_down
 
-    def proceed_on_broken_event(self, event) -> int:
+    def proceed_on_broken_event(self, event) -> None:
```
dmitry-ratushnyy: Maybe it should be bool? 🤔

dmitry-ratushnyy (approved):

Mehdi-Bendriss (approved):

PR #339, opened 2024-01-18
Update helpers lib: add function for k8s charm

dmitry-ratushnyy:
## Issue
To update helpers library on K8s charm we need to make sure that the library contains all the functions that being used in K8s charm

## Solution
Add missing functions


MiaAltieri (approved):
thank you for this

delgod (approved):

PR #346, opened 2024-01-31
[DPE-3456] add messages for incorrect rels to mongos

MiaAltieri:
## Issue
Right now Charmed MongoDB applications that are running as shard or replication can be integrated with mongos , this goes against the proposed architecture from the [sharding spec doc](https://docs.google.com/document/d/1kx5HUnMR2tiQQpCoWIyxJRUO2fdX4hhzVlo86noh-ts/edit?usp=sharingConnect)

 
## Solution
for cases in which either a Charmed MongoDB instance is in the role of `shard` or `replication` and a relation has been made to `mongos` then go into BlockedState

### DIFF
--- a/lib/charms/mongodb/v0/config_server_interface.py
+++ b/lib/charms/mongodb/v0/config_server_interface.py
@@ -69,12 +69,12 @@ class ClusterProvider(Object):
             event.defer()
             return False
 
-        if not self.charm.is_role(Config.Role.CONFIG_SERVER):
+        is_integrated_to_mongos = len(
+            self.charm.model.relations[Config.Relations.CLUSTER_RELATIONS_NAME]
+        )
+        if not self.charm.is_role(Config.Role.CONFIG_SERVER) and is_integrated_to_mongos:
             logger.info(
-                "Skipping %s. ShardingProvider is only be executed by config-server", type(event)
-            )
-            self.charm.unit.status = BlockedStatus(
-                "Relation to mongos not supported, config role must be config-server"
+                "Skipping %s. ClusterProvider is only be executed by config-server", type(event)
             )
             return False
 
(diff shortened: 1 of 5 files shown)


lib/charms/mongodb/v0/config_server_interface.py
```diff
@@ -208,7 +211,6 @@ def _on_relation_changed(self, event) -> None:
             event.relation.id, CONFIG_SERVER_DB_KEY
         )
         if not key_file_contents or not config_server_db:
-            event.defer()
```
MiaAltieri, line 211: this leads to the `mongos` charm forever waiting. Will need to merge this then fectch the lib from `mongos` side

dmitry-ratushnyy (dismissed):

lib/charms/mongodb/v0/config_server_interface.py
```diff
@@ -80,9 +83,20 @@ def pass_hook_checks(self, event: EventBase) -> bool:
 
         return True
 
+    def set_blocked_status_impossible_integration(self) -> None:
```
Mehdi-Bendriss: The core condition of this method is duplicated between 3 methods: `set_blocked_status_impossible_integration`, `pass_hook_checks` and `get_invalid_integration_status`.
 
Could we change this method (set_blocked...) to something along the lines of: `is_valid_mongos_integration -> bool` and take out the status setting to `_on_relation_changed` directly, and we refer to it in the 2 other methods relying on it

lib/charms/mongodb/v1/shards_interface.py
```diff
@@ -539,6 +539,9 @@ def pass_hook_checks(self, event):
             logger.info("skipping %s is only be executed by shards", type(event))
             return False
 
+        if not event.relation.app:
```
Mehdi-Bendriss, line 544: When would this happen?

Mehdi-Bendriss (commented):
Thanks Mia, I have a couple of comments

lib/charms/mongodb/v0/config_server_interface.py
MiaAltieri: good idea, ty

lib/charms/mongodb/v1/shards_interface.py
MiaAltieri: Good question - I am embarrassed to say that I do not recall, but that I experienced and error related to this last week when testing by hand. If I had to guess it would be that unit_changed events occured resulting in a `relation_changed` event where the app is not present

dmitry-ratushnyy (dismissed):

dmitry-ratushnyy (dismissed):

Mehdi-Bendriss (dismissed):

dmitry-ratushnyy (approved):

Mehdi-Bendriss (approved):

PR #351, opened 2024-02-09
[DPE-3500] Share backup password + create&list backups

MiaAltieri:
## Issue
Backups are not feasible in sharded clusters

## Solution
Make backups possible in sharded cluster + add tests. 

Necessary changes to make backups feasible in sharded cluster:
1. pbm password shared across cluster
2. shards restart with new pbm uri

## Other Changes
pbm was incorrectly configured to the wrong URI, according to the documentation it should use standalone URI


lib/charms/mongodb/v0/mongodb.py
```diff
@@ -66,6 +69,14 @@ def uri(self):
         auth_source = ""
         if self.database != "admin":
             auth_source = "&authSource=admin"
+
+        if self.standalone:
```
MiaAltieri, line 73: PBM was incorrectly configured previously (this only began causing issues in sharding) it should use the standalone URI

lib/charms/mongodb/v1/mongodb_backups.py
```diff
@@ -57,7 +57,7 @@
 REMAPPING_PATTERN = r"\ABackup doesn't match current cluster topology - it has different replica set names. Extra shards in the backup will cause this, for a simple example. The extra/unknown replica set names found in the backup are: ([^,\s]+)([.] Backup has no data for the config server or sole replicaset)?\Z"
 PBM_STATUS_CMD = ["status", "-o", "json"]
 MONGODB_SNAP_DATA_DIR = "/var/snap/charmed-mongodb/current"
-BACKUP_RESTORE_MAX_ATTEMPTS = 5
+BACKUP_RESTORE_MAX_ATTEMPTS = 10
```
MiaAltieri, line 60: It was the case that we needed to wait longer to confirm backups / restores so I increased the max attempts

src/config.py
```diff
@@ -17,7 +17,7 @@ class Config:
     MONGODB_SNAP_DATA_DIR = "/var/snap/charmed-mongodb/current"
     MONGOD_CONF_DIR = f"{MONGODB_SNAP_DATA_DIR}/etc/mongod"
     MONGOD_CONF_FILE_PATH = f"{MONGOD_CONF_DIR}/mongod.conf"
-    SNAP_PACKAGES = [("charmed-mongodb", "6/edge", 93)]
+    SNAP_PACKAGES = [("charmed-mongodb", "6/edge", 111)]
```
MiaAltieri, line 20: new snap wahoo 🎉

dmitry-ratushnyy (dismissed):

src/charm.py
```diff
@@ -225,7 +225,9 @@ def monitor_config(self) -> MongoDBConfiguration:
     def backup_config(self) -> MongoDBConfiguration:
         """Generates a MongoDBConfiguration object for backup."""
         self._check_or_set_user_password(BackupUser)
-        return self._get_mongodb_config_for_user(BackupUser, BackupUser.get_hosts())
+        return self._get_mongodb_config_for_user(
+            BackupUser, BackupUser.get_hosts(), standalone=True
```
Mehdi-Bendriss, line 229: Should `standalone` be always set to `True` ?

MiaAltieri: Yes it is necessary for the configuration of Percona Backup MongoDB. Is it worth it to add a comment in line 229 to make that more clear.

[from documentation:](https://docs.percona.com/percona-backup-mongodb/details/authentication.html)
> The pbm-agent processes should connect to their localhost mongod with a standalone type of connection.

Mehdi-Bendriss (dismissed):

lib/charms/mongodb/v0/mongodb.py
```diff
@@ -30,11 +30,13 @@
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 7
+LIBPATCH = 8
 
 # path to store mongodb ketFile
 logger = logging.getLogger(__name__)
 
+MONGODB_PORT = 27017
```
delgod: up to @Mehdi-Bendriss, maybe it can be `Config.MONGODB_PORT` like in `lib/charms/mongodb/v1/mongos.py`
```suggestion
from config import Config
```

percona-release_latest.generic_all.deb
delgod, line 1: I think we should remove this file

tests/integration/sharding_tests/test_sharding_backups.py
```diff
@@ -0,0 +1,127 @@
...
+    leader_unit = await backup_helpers.get_leader_unit(
+        ops_test, db_app_name=CONFIG_SERVER_APP_NAME
+    )
+    await backup_helpers.set_credentials(ops_test, github_secrets, cloud="AWS")
+    # verify backup list works
+    action = await leader_unit.run_action(action_name="list-backups")
+    list_result = await action.wait()
+    backups = list_result.results["backups"]
+    assert backups, "backups not outputted"
+
```
delgod, line 111: I think in a separate PR, we need to test that backups actually work across shards to avoid situations when the backup is restored only to one shard. Also, we should test that the backup can be restored to the new cluster.
The test can have the following logic:
1. write some data into a shared collection.
2. check the amount of data on each shard
3. make a backup
4. write more data.
5. check the amount of data grown on each shard.
6. stop writing. make a restore
7. check the amount of data on each shard; should be equal to step 2

delgod (dismissed):

tests/integration/sharding_tests/test_sharding_backups.py
Mehdi-Bendriss: Thank you! we'll track both tests on [DPE-3608](https://warthogs.atlassian.net/browse/DPE-3608) and [DPE-3609](https://warthogs.atlassian.net/browse/DPE-3609)

lib/charms/mongodb/v0/mongodb.py
MiaAltieri: sounds good to me, but this will have to be done on the MongoDB side, since this charm does not own the lib :)

percona-release_latest.generic_all.deb
MiaAltieri: 🤦‍♀️ most certainly yes

tests/integration/sharding_tests/test_sharding_backups.py
MiaAltieri: We already have a JIRA task for restores [3501](https://warthogs.atlassian.net/browse/DPE-3501) - I think merging the test logic for DPE-3608 into that tasks will definitely be feasible, perhaps for `DPE-3609` we will need the separate task

Mehdi-Bendriss (dismissed):

delgod (approved):

Mehdi-Bendriss (approved):

PR #356, opened 2024-02-14
fix (tests): ensure that the leader is managing secrets

tonyandrewmeyer:
## Issue

As of ops 2.10+ `Harness` applies the same access controls to secrets as Juju does in production. This causes tests to fail that were ignoring these access restrictions (deliberately - they have docstrings that call it out).

## Solution

Set the unit to be leader in each of the applicable tests. It might be that you'd rather have tests for both cases, leader and (failing appropriately) non-leader - I can adjust to do that if you prefer.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,7 +4,7 @@ cosl==0.0.7
 importlib-resources==5.10.2
 tenacity==8.1.0
 pymongo==4.3.3
-ops==2.4.1
+ops==2.12
 jsonschema==4.17.3
 cryptography==38.0.4
 pure-sasl==0.6.2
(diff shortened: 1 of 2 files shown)


Mehdi-Bendriss:
Hello @tonyandrewmeyer, thanks for the PR! 

Apologies for the delay, we're finally getting around tackling these issues as we're [upgrading all dependencies](https://github.com/canonical/mongodb-operator/pull/395).

Could you please add the test case you mentioned? (failure in case non-leader attempts to manage secrets)?

tonyandrewmeyer:
> Could you please add the test case you mentioned? (failure in case non-leader attempts to manage secrets)?

Sure - done!

Mehdi-Bendriss (approved):
Thank you!

MiaAltieri (approved):
Looks good to me. Thank you very much @tonyandrewmeyer

PR #359, opened 2024-02-21
[DPE-3501] - add restore test for sharded cluster

MiaAltieri:
## Issue
Uncertain that restores are supported in current backup implemenation

## Solution
Verify their functionality via an integration test

1. write some data into a shared collection.
2. check the amount of data on each shard
3. make a backup
4. write more data.
5. check the amount of data grown on each shard.
6. stop writing. make a restore
7. check the amount of data on each shard; should be equal to step 2

### DIFF
--- a/tests/integration/sharding_tests/test_sharding_backups.py
+++ b/tests/integration/sharding_tests/test_sharding_backups.py
@@ -8,7 +8,7 @@ import time
 
 import pytest
 from pytest_operator.plugin import OpsTest
-from tenacity import RetryError, Retrying, stop_after_delay, wait_fixed
+from tenacity import Retrying, stop_after_delay, wait_fixed
 
 from ..backup_tests import helpers as backup_helpers
 
@@ -116,19 +116,16 @@ async def test_set_credentials_in_cluster(ops_test: OpsTest, github_secrets) -> 
 
 @pytest.mark.group(1)
 @pytest.mark.abort_on_fail
-async def test_create_and_list_backups_in_cluster(ops_test: OpsTest, github_secrets) -> None:
+async def test_create_and_list_backups_in_cluster(ops_test: OpsTest) -> None:
     """Tests that sharded cluster can successfully create and list backups."""
-    leader_unit = await backup_helpers.get_leader_unit(
-        ops_test, db_app_name=CONFIG_SERVER_APP_NAME
-    )
-    await backup_helpers.set_credentials(ops_test, github_secrets, cloud="AWS")
     # verify backup list works
-    action = await leader_unit.run_action(action_name="list-backups")
-    list_result = await action.wait()
-    backups = list_result.results["backups"]
+    backups = await backup_helpers.get_backup_list(ops_test, db_app_name=CONFIG_SERVER_APP_NAME)
     assert backups, "backups not outputted"
 
     # verify backup is started
+    leader_unit = await backup_helpers.get_leader_unit(
+        ops_test, db_app_name=CONFIG_SERVER_APP_NAME
+    )
     action = await leader_unit.run_action(action_name="create-backup")
     backup_result = await action.wait()
     assert "backup started" in backup_result.results["backup-status"], "backup didn't start"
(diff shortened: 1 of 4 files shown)


tests/integration/sharding_tests/test_sharding_backups.py
```diff
@@ -200,3 +216,89 @@ async def test_rotate_backup_password(ops_test: OpsTest) -> None:
                 assert backups == 2
     except RetryError:
         assert backups == 2, "Backup not created after password rotation."
+
+
+@pytest.mark.group(1)
+@pytest.mark.abort_on_fail
+async def test_restore_backup(ops_test: OpsTest, add_writes_to_db) -> None:
```
phvalguima, line 222: @MiaAltieri you should have tests for removing/readding relation + deleting the cluster and redeploying a new one, then restoring

phvalguima (dismissed):
LGTM
After discussing, the missing restore in a new cluster test will be covered as part of [DPE-3609](https://warthogs.atlassian.net/browse/DPE-3609)

tests/integration/sharding_tests/test_sharding_backups.py
```diff
@@ -200,3 +216,89 @@ async def test_rotate_backup_password(ops_test: OpsTest) -> None:
...
+    assert first_backup.status == "completed", "First backup not started."
+
+    # verify that backup was made on the bucket
+    try:
+        for attempt in Retrying(stop=stop_after_delay(4), wait=wait_fixed(5)):
+            with attempt:
+                backups = await backup_helpers.count_logical_backups(leader_unit)
+                assert backups == prev_backups + 1, "Backup not created."
+    except RetryError:
+        assert backups == prev_backups + 1, "Backup not created."
```
Mehdi-Bendriss: is this needed?

tests/integration/sharding_tests/test_sharding_backups.py
```diff
@@ -200,3 +216,89 @@ async def test_rotate_backup_password(ops_test: OpsTest) -> None:
...
+    )
+    assert (
+        new_total_writes["total_writes"] > cluster_writes["total_writes"]
+    ), "No writes to be cleared after restoring."
+
+    # find most recent backup id and restore
+    action = await leader_unit.run_action(action_name="list-backups")
+    list_result = await action.wait()
+    list_result = list_result.results["backups"]
+    most_recent_backup = list_result.split("\n")[-1]
```
Mehdi-Bendriss: For a future PR, this seems to be quite common / duplicated across multiple helpers, let's extract it into it's own `list_backups` function

tests/integration/sharding_tests/test_sharding_backups.py
```diff
@@ -200,3 +216,89 @@ async def test_rotate_backup_password(ops_test: OpsTest) -> None:
...
+                assert (
+                    restored_total_writes["total_writes"] == cluster_writes["total_writes"]
+                ), "writes not correctly restored to whole cluster"
+                assert (
+                    restored_total_writes[SHARD_ONE_APP_NAME] == cluster_writes[SHARD_ONE_APP_NAME]
+                ), f"writes not correctly restored to {SHARD_ONE_APP_NAME}"
+                assert (
+                    restored_total_writes[SHARD_TWO_APP_NAME] == cluster_writes[SHARD_TWO_APP_NAME]
+                ), f"writes not correctly restored to {SHARD_TWO_APP_NAME}"
+    except RetryError:
```
Mehdi-Bendriss: Same as above, is this needed?

Mehdi-Bendriss (commented):
Thanks Mia, it looks good - I have a couple of questions

tests/integration/sharding_tests/test_sharding_backups.py
MiaAltieri: Yes - the reason is I don't want to just assert failure when the `RetryError` occurs, I want to know exactly what the error was and what the count was, what was the expected count, and a helpful message.

tests/integration/sharding_tests/test_sharding_backups.py
MiaAltieri: Yes - the reason is I don't want to just assert failure when the `RetryError` occurs, I want to know exactly what the error was and what the count was, what was the expected count, and a helpful message.

tests/integration/sharding_tests/test_sharding_backups.py
MiaAltieri: In my current work on migration restore I started moving things over to use the pre-existing `create_and_list_backups` function. Let me make this chagne here.

tests/integration/sharding_tests/test_sharding_backups.py
Mehdi-Bendriss: Yes, you essentially want to propagate the assertion failure instead of an obscure `RetryError`. 
You can achieve this by adding `reraise=True` in the `Retrying` context, and you can drop the `except RetryError`

Mehdi-Bendriss (approved):

phvalguima (approved):

PR #363, opened 2024-02-29
[DPE-3651] fix race condition with cluster auth

### LINKED ISSUE #348: Flakey sharding tests
## Steps to reproduce

1. 

## Versions

Operating system: 

Juju CLI: 

Juju agent: 

Charm revision: 

LXD: 

## Log output

Juju debug log:

MiaAltieri:
## Issue
Flakey sharding tests would show that a cluster component would *temporarily* go into error due to unauthorised access in update_status, only for the for the unit to get resolved in the next hook execution

## Reasoning
this is due to improper sync of passwords across cluster

Fixes #348 

## Solution
add a check in update status to notify the user why the cluster component is not yet connected

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -548,6 +548,17 @@ class MongodbOperatorCharm(CharmBase):
                 self.unit.status = WaitingStatus("Waiting for MongoDB to start")
                 return
 
+        # Cannot check more advanced MongoDB statuses if the cluster doesn't have passwords synced
+        # this can occur in two cases:
+        # 1. password rotation
+        # 2. race conditions when a new shard is addeded.
+        if (
+            not self.shard.cluster_password_synced()
+            or not self.config_server.cluster_password_synced()
+        ):
+            self.unit.status = WaitingStatus("Waiting to sync passwords across the cluster")
+            return
+
         # leader should periodically handle configuring the replica set. Incidents such as network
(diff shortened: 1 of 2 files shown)


dmitry-ratushnyy (dismissed):

src/charm.py
```diff
@@ -548,6 +548,17 @@ def _on_update_status(self, event: UpdateStatusEvent):
                 self.unit.status = WaitingStatus("Waiting for MongoDB to start")
                 return
 
+        # Cannot check more advanced MongoDB statuses if the cluster doesn't have passwords synced
+        # this can occur in two cases:
+        # 1. password rotation
+        # 2. race conditions when a new shard is addeded.
+        if (
+            not self.shard.cluster_password_synced()
+            or not self.config_server.cluster_password_synced()
```
Mehdi-Bendriss, line 557: Is update_status the only hook where this must be handled?

Mehdi-Bendriss (dismissed):

src/charm.py
MiaAltieri: It is the only place I have seen the error occur, but I am surprised I have not seen it elsewhere.

I could be easily convinced to add it elsewhere in the code, but I believe this is sufficent for the time-being

dmitry-ratushnyy (approved):

Mehdi-Bendriss (approved):

PR #367, opened 2024-03-04
Request wide SANs in the internal CSR

delgod:
to avoid unnecessary restarts when adding/removing shards or units.


MiaAltieri:
solved with #369

PR #370, opened 2024-03-07
[DPE-3650] TLS now supported on sharded deployments

MiaAltieri:
Future PR:
 Add Useful TLS error messages i.e.
        1. Shard has TLS but config-server does not 
        2. Config-server has TLS but shard does not 
        3. CA mismatch

### DIFF
--- a/lib/charms/mongodb/v0/mongodb_tls.py
+++ b/lib/charms/mongodb/v0/mongodb_tls.py
@@ -218,8 +218,8 @@ class MongoDBTLS(Object):
             return
 
         logger.debug("Generating a new Certificate Signing Request.")
-        key = self.charm.get_tls_secret(internal, Config.TLS.SECRET_KEY_LABEL).encode("utf-8")
-        old_csr = self.charm.get_tls_secret(internal, Config.TLS.SECRET_CSR_LABEL).encode("utf-8")
+        key = self.get_tls_secret(internal, Config.TLS.SECRET_KEY_LABEL).encode("utf-8")
+        old_csr = self.get_tls_secret(internal, Config.TLS.SECRET_CSR_LABEL).encode("utf-8")
         new_csr = generate_csr(
             private_key=key,
             subject=self.get_host(self.charm.unit),
--- a/lib/charms/mongodb/v0/mongodb_tls.py
+++ b/lib/charms/mongodb/v0/mongodb_tls.py
@@ -76,17 +76,18 @@ class MongoDBTLS(Object):
         """Set the TLS private key, which will be used for requesting the certificate."""
         logger.debug("Request to set TLS private key received.")
         try:
-            self._request_certificate(event.params.get("external-key", None), internal=False)
-            self._request_certificate(event.params.get("internal-key", None), internal=True)
+            self.request_certificate(event.params.get("external-key", None), internal=False)
+            self.request_certificate(event.params.get("internal-key", None), internal=True)
             logger.debug("Successfully set TLS private key.")
         except ValueError as e:
             event.fail(str(e))
 
-    def _request_certificate(
+    def request_certificate(
         self,
         param: Optional[str],
         internal: bool,
     ):
+        """Request TLS certificate."""
         if param is None:
             key = generate_private_key()
         else:
@@ -94,8 +95,8 @@ class MongoDBTLS(Object):
 
         csr = generate_csr(
             private_key=key,
-            subject=self.get_host(self.charm.unit),
-            organization=self.charm.app.name,
+            subject=self._get_subject_name(),
+            organization=self._get_subject_name(),
             sans=self._get_sans(),
             sans_ip=[str(self.charm.model.get_binding(self.peer_relation).network.bind_address)],
         )
@@ -124,8 +125,8 @@ class MongoDBTLS(Object):
 
     def _on_tls_relation_joined(self, _: RelationJoinedEvent) -> None:
         """Request certificate when TLS relation joined."""
-        self._request_certificate(None, internal=True)
-        self._request_certificate(None, internal=False)
+        self.request_certificate(None, internal=True)
+        self.request_certificate(None, internal=False)
 
     def _on_tls_relation_broken(self, event: RelationBrokenEvent) -> None:
         """Disable TLS when TLS relation broken."""
@@ -149,7 +150,7 @@ class MongoDBTLS(Object):
 
         if ext_csr and event.certificate_signing_request.rstrip() == ext_csr.rstrip():
             logger.debug("The external TLS certificate available.")
-            internal = False  # external crs
+            internal = False
         elif int_csr and event.certificate_signing_request.rstrip() == int_csr.rstrip():
             logger.debug("The internal TLS certificate available.")
             internal = True
@@ -222,8 +223,8 @@ class MongoDBTLS(Object):
         old_csr = self.get_tls_secret(internal, Config.TLS.SECRET_CSR_LABEL).encode("utf-8")
         new_csr = generate_csr(
             private_key=key,
-            subject=self.get_host(self.charm.unit),
-            organization=self.charm.app.name,
+            subject=self._get_subject_name(),
+            organization=self._get_subject_name(),
             sans=self._get_sans(),
             sans_ip=[str(self.charm.model.get_binding(self.peer_relation).network.bind_address)],
         )
@@ -293,3 +294,15 @@ class MongoDBTLS(Object):
         scope = "int" if internal else "ext"
         label_name = f"{scope}-{label_name}"
         return self.charm.get_secret(UNIT_SCOPE, label_name)
+
+    def _get_subject_name(self) -> str:
+        """Generate the subject name for CSR."""
+        # In sharded MongoDB deployments it is a requirement that all subject names match across
+        # all cluster components
+        if self.charm.is_role(Config.Role.SHARD):
+            # until integrated with config-server use current app name as
+            # subject name
+            subject_name = self.charm.shard.get_config_server_name() or self.charm.app.name
+            return subject_name
+
+        return self.charm.app.name
--- a/lib/charms/mongodb/v0/mongodb_tls.py
+++ b/lib/charms/mongodb/v0/mongodb_tls.py
@@ -8,7 +8,6 @@ and expose needed information for client connection via fields in
 external relation.
 """
 import base64
-import json
 import logging
 import re
 import socket
@@ -106,8 +105,8 @@ class MongoDBTLS(Object):
         self.set_tls_secret(internal, Config.TLS.SECRET_CERT_LABEL, None)
 
         label = "int" if internal else "ext"
-        self.charm.unit_peer_data[f"{label}_certs_subject"] = json.dumps(self._get_subject_name())
-        self.charm.unit_peer_data[f"{label}_certs_subject"] = json.dumps(self._get_subject_name())
+        self.charm.unit_peer_data[f"{label}_certs_subject"] = self._get_subject_name()
+        self.charm.unit_peer_data[f"{label}_certs_subject"] = self._get_subject_name()
 
         if self.charm.model.get_relation(Config.TLS.TLS_PEER_RELATION):
             self.certs.request_certificate_creation(certificate_signing_request=csr)
--- a/lib/charms/mongodb/v0/mongodb_tls.py
+++ b/lib/charms/mongodb/v0/mongodb_tls.py
@@ -39,7 +39,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 8
+LIBPATCH = 9
 
 logger = logging.getLogger(__name__)
 
--- a/lib/charms/mongodb/v0/mongodb_tls.py
+++ b/lib/charms/mongodb/v0/mongodb_tls.py
@@ -8,6 +8,7 @@ and expose needed information for client connection via fields in
 external relation.
 """
 import base64
+import json
 import logging
 import re
 import socket
@@ -100,10 +101,13 @@ class MongoDBTLS(Object):
             sans=self._get_sans(),
             sans_ip=[str(self.charm.model.get_binding(self.peer_relation).network.bind_address)],
         )
-
         self.set_tls_secret(internal, Config.TLS.SECRET_KEY_LABEL, key.decode("utf-8"))
         self.set_tls_secret(internal, Config.TLS.SECRET_CSR_LABEL, csr.decode("utf-8"))
         self.set_tls_secret(internal, Config.TLS.SECRET_CERT_LABEL, None)
+
+        label = "int" if internal else "ext"
+        self.charm.unit_peer_data[f"{label}_certs_subject"] = json.dumps(self._get_subject_name())
+        self.charm.unit_peer_data[f"{label}_certs_subject"] = json.dumps(self._get_subject_name())
 
         if self.charm.model.get_relation(Config.TLS.TLS_PEER_RELATION):
             self.certs.request_certificate_creation(certificate_signing_request=csr)
@@ -166,7 +170,7 @@ class MongoDBTLS(Object):
         self.set_tls_secret(internal, Config.TLS.SECRET_CERT_LABEL, event.certificate)
         self.set_tls_secret(internal, Config.TLS.SECRET_CA_LABEL, event.ca)
 
-        if self._waiting_for_certs():
+        if self.waiting_for_certs():
             logger.debug(
                 "Defer till both internal and external TLS certificates available to avoid second restart."
             )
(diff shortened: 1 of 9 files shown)


lib/charms/mongodb/v1/helpers.py
```diff
@@ -121,12 +121,40 @@ def get_mongos_args(
         f"--configdb {config_server_db}",
         # config server is already using 27017
         f"--port {Config.MONGOS_PORT}",
-        f"--keyFile={full_conf_dir}/{KEY_FILE}",
-        "\n",
     ]
 
-    # TODO Future PR: support TLS on mongos
+    # TODO : generalise these into functions to be re-used
```
MiaAltieri, line 126: lets do this in a future PR

lib/charms/mongodb/v0/mongodb_tls.py
```diff
@@ -293,3 +298,15 @@ def get_tls_secret(self, internal: bool, label_name: str) -> str:
...
+
+    def _get_subject_name(self) -> str:
+        """Generate the subject name for CSR."""
+        # In sharded MongoDB deployments it is a requirement that all subject names match across
+        # all cluster components
+        if self.charm.is_role(Config.Role.SHARD):
+            # until integrated with config-server use current app name as
+            # subject name
+            subject_name = self.charm.shard.get_config_server_name() or self.charm.app.name
+            return subject_name
```
Mehdi-Bendriss: nit: 
```suggestion
            return self.charm.shard.get_config_server_name() or self.charm.app.name
```

lib/charms/mongodb/v1/helpers.py
```diff
@@ -121,12 +121,40 @@ def get_mongos_args(
...
+                f"--tlsCAFile={full_conf_dir}/{TLS_EXT_CA_FILE}",
+                f"--tlsCertificateKeyFile={full_conf_dir}/{TLS_EXT_PEM_FILE}",
+                # allow non-TLS connections
+                "--tlsMode=preferTLS",
+                "--tlsDisabledProtocols=TLS1_0,TLS1_1",
+            ]
+        )
+
+    # internal TLS can be enabled only in external is enabled
+    if config.tls_internal and config.tls_external:
```
Mehdi-Bendriss, line 139: Can we ever have `config.tls_internal and not config.tls_external`?

lib/charms/mongodb/v1/shards_interface.py
```diff
@@ -541,28 +558,96 @@ def _handle_changed_secrets(self, event) -> None:
             event.defer()
             return
 
+        # FUTURE PR: if config-server does not have TLS enabled log a useful message and go into
+        # blocked in relation_changed and other status checks
+
+    def get_membership_auth_modes(self, event) -> Tuple:
```
Mehdi-Bendriss: nit: ``
```suggestion
    def get_membership_auth_modes(self, event: RelationChangedEvent) -> Tuple[bool, bool]:
```

lib/charms/mongodb/v1/shards_interface.py
```diff
@@ -541,28 +558,96 @@ def _handle_changed_secrets(self, event) -> None:
...
+        """Returns the available authentication membership forms."""
+        key_file_contents = self.database_requires.fetch_relation_field(
+            event.relation.id, KEYFILE_KEY
+        )
+        tls_ca = self.database_requires.fetch_relation_field(event.relation.id, INT_TLS_CA_KEY)
+        return (key_file_contents is not None, tls_ca is not None)
+
+    def update_member_auth(self, event, membership_auth) -> None:
```
Mehdi-Bendriss:
```suggestion
    def update_member_auth(
        self, event: RelationChangedEvent, membership_auth: Tuple[bool, bool]
    ) -> None:
```

lib/charms/mongodb/v1/shards_interface.py
```diff
@@ -541,28 +558,96 @@ def _handle_changed_secrets(self, event) -> None:
...
+        """Updates the shard to have the same membership auth as the config-server."""
+        (key_file_contents, cluster_tls_ca) = membership_auth
```
Mehdi-Bendriss: perhaps?
```suggestion
        is_key_file_contents_auth, is_tls_auth = membership_auth
```

lib/charms/mongodb/v1/shards_interface.py
```diff
@@ -937,3 +1014,19 @@ def get_mongos_hosts(self) -> List[str]:
             return
 
         return json.loads(config_server_relation.data[config_server_relation.app].get(HOSTS_KEY))
+
+    def has_requested_cluster_certs(self) -> bool:
+        """Returns if the shard has already requested the certificates for internal-membership."""
+        if (
+            "int_certs_subject" not in self.charm.unit_peer_data
+            or "ext_certs_subject" not in self.charm.unit_peer_data
```
Mehdi-Bendriss: where are these ever set?

lib/charms/mongodb/v1/shards_interface.py
```diff
@@ -937,3 +1014,19 @@ def get_mongos_hosts(self) -> List[str]:
...
+        ):
+            logger.info("Unit hasn't requested necessary all certs for cluster TLS.")
+            return False
+
+        int_subject = json.loads(self.charm.unit_peer_data["int_certs_subject"])
+        ext_subject = json.loads(self.charm.unit_peer_data["ext_certs_subject"])
```
Mehdi-Bendriss: this won't work unless `int/ext_certs_subject` are booleans

Mehdi-Bendriss: perhaps:
```
def _should_request_new_certs(self) -> bool:
    """Returns if the shard has already requested the certificates for internal-membership."""
    int_subject = self.charm.unit_peer_data.get("int_certs_subject")
    ext_subject = self.charm.unit_peer_data.get("ext_certs_subject")
    return {int_subject, ext_subject} != {self.get_config_server_name()}
```

lib/charms/mongodb/v1/shards_interface.py
```diff
@@ -541,28 +558,96 @@ def _handle_changed_secrets(self, event) -> None:
...
+        # have the correct cluster KeyFile readily available.
+        key_file_contents = self.database_requires.fetch_relation_field(
+            event.relation.id, KEYFILE_KEY
+        )
+        self.update_keyfile(key_file_contents=key_file_contents)
+
+    # Future PR - status updating for inconsistencies with TLS (i.e. shard has TLS but
+    # config-server does not and vice versa or CA-mismatch)
+
+    def get_cluster_passwords(self, event) -> Tuple:
```
Mehdi-Bendriss:
```suggestion
    def get_cluster_passwords(self, event: RelationChangedEvent) -> Tuple[Optional[str], Optional[str]]:
```

lib/charms/mongodb/v1/shards_interface.py
```diff
@@ -541,28 +558,96 @@ def _handle_changed_secrets(self, event) -> None:
...
+        """Retrieves shared cluster passwords."""
+        operator_password = self.database_requires.fetch_relation_field(
+            event.relation.id, OPERATOR_PASSWORD_KEY
+        )
+        backup_password = self.database_requires.fetch_relation_field(
+            event.relation.id, BACKUP_PASSWORD_KEY
+        )
+        return (operator_password, backup_password)
+
+    def update_cluster_passwords(self, event, operator_password, backup_password):
```
Mehdi-Bendriss:
```suggestion
    def update_cluster_passwords(
        self, event: RelationChangedEvent, operator_password: str, backup_password: str
    ) -> None:
```

lib/charms/mongodb/v1/shards_interface.py
```diff
@@ -541,28 +558,96 @@ def _handle_changed_secrets(self, event) -> None:
...
+        try:
+            self.update_password(
+                username=OperatorUser.get_username(), new_password=operator_password
+            )
+            self.update_password(BackupUser.get_username(), new_password=backup_password)
+        except RetryError:
+            self.charm.unit.status = BlockedStatus("Shard not added to config-server")
+            logger.error(
+                "Shard could not be added to config server, failed to set operator password."
+            )
```
Mehdi-Bendriss: Could you explain this? it's not immediately clear to me from the `update_password` logic

lib/charms/mongodb/v1/shards_interface.py
```diff
@@ -893,6 +966,10 @@ def cluster_password_synced(self) -> bool:
             if e.code == 18:  # Unauthorized Error - i.e. password is not in sync
                 return False
             raise
+        except ServerSelectionTimeoutError as e:
+            if e.code == 111:  # Connection refused, - i.e. TLS certs not in sync
```
Mehdi-Bendriss: Could you explain what "not in sync" means here?

Mehdi-Bendriss (commented):
Thanks Mia! I left a few comments / questions

lib/charms/mongodb/v1/shards_interface.py
MiaAltieri: Yes I will add to the comment. This error happens when internal membership is not working across the cluster.

i.e. shard has incorrect internal membership certs and so the config-server cannot communicate with it and vice versa

lib/charms/mongodb/v1/shards_interface.py
MiaAltieri: oooh yes - thank you Mehdi 🙇

lib/charms/mongodb/v1/shards_interface.py
MiaAltieri: In TLS class when the cert is requested. When we request the cert we save the name of the subject that we used upon requesting it

lib/charms/mongodb/v1/helpers.py
MiaAltieri: YES - But, its a long story. 

Context:
A key part of the legacy MongoDB relations is that they expect [very specific fields](https://github.com/canonical/mongodb-operator/blob/c2744cf25faed01a03d63c0b22cb8728a12c7d12/lib/charms/mongodb/v1/mongodb_vm_legacy_provider.py#L95) - in a way that if external TLS is enabled we would not be able to tell clients using the legacy relations these external connection secrets, and then they wouldn't be able to connect :(

Moving Forward:
Now that we have planned to cut support of `5/edge`, are we also cutting legacy relation support? If so, there is no need to handle this edge case.

lib/charms/mongodb/v1/shards_interface.py
MiaAltieri: which exact part? 🙈 

I actually just moved this from the `relation_changed` function since I thought that `relation_changed`  was becoming too big. 

The idea is that if the shard cannot update its passwords to mirror those of the config-server, that it isn't added to the sharded cluster. Would renaming the function to: `sync_cluster_passwords` be better? I could update the log message to say:
`logger.error("Failed to sync cluster passwords from config-server to shard. Shard cannot be added to config-server, deferring event and retrying.")`

lib/charms/mongodb/v1/helpers.py
```diff
@@ -121,12 +121,40 @@ def get_mongos_args(
...
-    # TODO Future PR: support TLS on mongos
+    # TODO : generalise these into functions to be re-used
+    if not config.tls_internal:
+        # keyFile cannot be used without auth and cannot be used in tandem with internal TLS
+        cmd.extend(
+            [
+                "--clusterAuthMode=keyFile",
+                f"--keyFile={full_conf_dir}/{KEY_FILE}",
+            ]
+        )
```
delgod: Will any auth be configured `if config.tls_internal and not config.tls_external:`?
I believe or keyFile or TLS auth should be configured.

lib/charms/mongodb/v1/shards_interface.py
```diff
@@ -576,34 +669,20 @@ def _on_relation_changed(self, event):
...
-            )
-            event.defer()
-            return
+        self.sync_cluster_passwords(event, operator_password, backup_password)
 
         # after updating the password of the backup user, restart pbm with correct password
         self.charm._connect_pbm_agent()
         self.charm.app_peer_data["mongos_hosts"] = json.dumps(self.get_mongos_hosts())
 
+        self.charm.unit.status = ActiveStatus("")
```
delgod: It looks like don't matter what status we had before; we changed it to Active here.
In any case, status handling up to Mehdi

lib/charms/mongodb/v1/shards_interface.py
```diff
@@ -541,28 +560,102 @@ def _handle_changed_secrets(self, event) -> None:
...
             return
 
         # if re-using an old shard, re-set drained flag.
         self.charm.unit_peer_data["drained"] = json.dumps(False)
 
-        # TODO: Future PR better status message behavior
-        self.charm.unit.status = MaintenanceStatus("Adding shard to config-server")
+        # relation-changed events can be used for other purposes (not only adding the shard), i.e.
+        # password rotation, secret rotation, mongos hosts rotation
+        if not self._is_mongos_reachable() or not self._is_added_to_cluster():
```
delgod: It looks like if any mongos fail to happen, we notify the user that the charm `Adding shard to config-server`.
I wouldn't say that it is always true.
In any case, status handling up to Mehdi

lib/charms/mongodb/v1/shards_interface.py
```diff
@@ -541,28 +560,102 @@ def _handle_changed_secrets(self, event) -> None:
...
+            event.relation.id, KEYFILE_KEY
+        )
+        tls_ca = self.database_requires.fetch_relation_field(event.relation.id, INT_TLS_CA_KEY)
+        return (key_file_contents is not None, tls_ca is not None)
+
+    def update_member_auth(
+        self, event: RelationChangedEvent, membership_auth: Tuple[bool, bool]
+    ) -> None:
+        """Updates the shard to have the same membership auth as the config-server."""
+        cluster_auth_keyfile, cluster_auth_tls = membership_auth
```
delgod, line 576: feel free to ignore: Is there any reason why a Tuple is needed as input if we convert it to two variables later?

lib/charms/mongodb/v1/shards_interface.py
```diff
@@ -541,28 +560,102 @@ def _handle_changed_secrets(self, event) -> None:
...
+        tls_integrated = self.charm.model.get_relation(Config.TLS.TLS_PEER_RELATION)
+
+        # Edge case: shard has TLS enabled before having connected to the config-server. For TLS in
+        # sharded MongoDB clusters it is necessary that the subject and organisation name are the
+        # same in their CSRs. Re-requesting a cert after integrated with the config-server
+        # regenerates the cert with the appropriate configurations needed for sharding.
+        if cluster_auth_tls and tls_integrated and not self._should_request_new_certs():
```
delgod: maybe we should `request_certificate()` if `_should_request_new_certs==True`
```suggestion
        if cluster_auth_tls and tls_integrated and self._should_request_new_certs():
```

delgod (changes requested):

lib/charms/mongodb/v1/shards_interface.py
MiaAltieri: Hmm yes I agree, I will revert this change

lib/charms/mongodb/v1/shards_interface.py
MiaAltieri: I agree - let me improve

lib/charms/mongodb/v1/shards_interface.py
MiaAltieri: I wanted to be consistent with it's "sister" function `get_membership_auth_modes` which returns a `Tuple`

lib/charms/mongodb/v1/shards_interface.py
MiaAltieri: better?

lib/charms/mongodb/v1/shards_interface.py
MiaAltieri: better?

lib/charms/mongodb/v1/helpers.py
Mehdi-Bendriss: I understand now, thank you. Yes, for `6/stable`, we officially stop:
1. supporting legacy relations
2. supporting `5/`

lib/charms/mongodb/v1/shards_interface.py
Mehdi-Bendriss: Yes, in general, i believe we should move away from setting `ActiveStatus` to rather "clearing" the specific statuses we're targeting. ActiveStatus is too intrusive in the sense that it overrides any component's status (even though unrelated with the current)

lib/charms/mongodb/v1/shards_interface.py
delgod: thanks

delgod (approved):

Mehdi-Bendriss (approved):

lib/charms/mongodb/v1/helpers.py
MiaAltieri: Thanks for clearing this up! I knew about `5/` but wasn't sure about legacy. Thank you 🙇

PR #371, opened 2024-03-12
[DPE-3627] - Sanity Checks for TLS in sharded cluster 

MiaAltieri:
## Issue
There are no safegaurds + informative messages to notify users that they have incorrectly configured TLS in a sharded cluster

## Solution
Add sanity checks + error reporting for these three cases:
1. Shard has TLS enabled but config-server does not have TLS enabled
2. Config-server has TLS  enabled but shard does not have TLS enabled
3. Both components have TLS enabled, but they use different CA's

### DIFF
--- a/lib/charms/mongodb/v1/shards_interface.py
+++ b/lib/charms/mongodb/v1/shards_interface.py
@@ -641,7 +641,7 @@ class ConfigServerRequirer(Object):
         self.update_member_auth(event, (key_file_enabled, tls_enabled))
 
         # restart on high loaded databases can be very slow (e.g. up to 10-20 minutes).
-        with MongoDBConnection(self.charm.mongodb_config, "localhost", direct=True) as mongo:
+        with MongoDBConnection(self.charm.mongodb_config) as mongo:
             if not mongo.is_ready:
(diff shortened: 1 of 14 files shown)


lib/charms/mongodb/v1/shards_interface.py
```diff
@@ -540,26 +549,10 @@ def _handle_changed_secrets(self, event) -> None:
...
-            self.update_password(BackupUser.get_username(), new_password=backup_password)
-        except (NotReadyError, PyMongoError):
-            self.charm.unit.status = BlockedStatus("Failed to rotate cluster secrets")
-            logger.error("Shard failed to rotate cluster secrets.")
-            event.defer()
-            return
-
-        # FUTURE PR: if config-server does not have TLS enabled log a useful message and go into
-        # blocked in relation_changed and other status checks
+        self.sync_cluster_passwords(event, operator_password, backup_password)
```
MiaAltieri, line 555: re-using functions from last PR

delgod (dismissed):

Mehdi-Bendriss (dismissed):

Mehdi-Bendriss (approved):

delgod (approved):

PR #373, opened 2024-03-13
[DPE-3802] make sure backup-restore tests writes on all shards

MiaAltieri:
## Issue
Backup - Restore tests should ensure there are writes to clear / restore on all shards

## Solution
Add writes to all shards

### DIFF
--- a/tests/integration/sharding_tests/test_sharding_backups.py
+++ b/tests/integration/sharding_tests/test_sharding_backups.py
@@ -11,10 +11,9 @@ from pytest_operator.plugin import OpsTest
 from tenacity import Retrying, stop_after_delay, wait_fixed
 
 from ..backup_tests import helpers as backup_helpers
-
-# from .writes_helpers import writes_helpers
 from ..helpers import get_leader_id, get_password, set_password
 from . import writes_helpers
+from .helpers import generate_mongodb_client, write_data_to_mongodb
 
 S3_APP_NAME = "s3-integrator"
 SHARD_ONE_APP_NAME = "shard-one"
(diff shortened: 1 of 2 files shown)


tests/integration/sharding_tests/test_sharding_backups.py
```diff
@@ -24,19 +23,43 @@
...
+        mongos_client,
+        db_name=SHARD_TWO_DB_NAME,
+        coll_name=SHARD_TWO_COLL_NAME,
+        content={"horse-breed": "unicorn", "real": True},
+    )
+    mongos_client.admin.command("movePrimary", SHARD_TWO_DB_NAME, to=SHARD_TWO_APP_NAME)
+
     yield
     await writes_helpers.clear_db_writes(ops_test)
+    await writes_helpers.remove_db_writes(
```
Mehdi-Bendriss: I believe these are exact duplicates, could we extend `clear_db_writes` with default arguments (`db`, `collection`) and make it more generic?

tests/integration/sharding_tests/test_sharding_backups.py
```diff
@@ -232,13 +255,28 @@ async def test_rotate_backup_password(ops_test: OpsTest) -> None:
...
+    shard_two_writes = await writes_helpers.count_shard_writes(
+        ops_test,
+        shard_app_name=SHARD_TWO_APP_NAME,
+        db_name=SHARD_TWO_DB_NAME,
+        collection_name=SHARD_TWO_COLL_NAME,
+    )
+
+    assert cluster_writes["total_writes"], "no writes to backup"
+    assert shard_one_writes, "no writes to backup for shard one"
+    assert shard_two_writes, "no writes to backup for shard two"
```
Mehdi-Bendriss, line 274: Can we add something along the lines of: 
```suggestion
   assert shard_one_writes + shard_two_writes >= cluster_writes["total_writes"], "writes not synced"
```

tests/integration/sharding_tests/writes_helpers.py
```diff
@@ -57,6 +57,26 @@ async def clear_db_writes(ops_test: OpsTest, config_server_name=APP_NAME) -> boo
     client.close()
 
 
+async def remove_db_writes(
```
Mehdi-Bendriss, line 43: same comment as above about making clear_db_writes generic

Mehdi-Bendriss (commented):
Thanks Mia! I left a few comments

tests/integration/sharding_tests/test_sharding_backups.py
MiaAltieri: `clear_db_writes` is poorly named - it also stops the continous writes process. 

Let me re-name the functions and consoldate :)

tests/integration/sharding_tests/test_sharding_backups.py
MiaAltieri: yes! changed it to `==` but that was because I changed the count function

Mehdi-Bendriss (approved):

PR #377, opened 2024-03-14
[DPE-3813] make cluster password sync more robust

MiaAltieri:
## Issue
During sharding integration tests, sharding components will occasionally go into error state when executing `update_status`. However when the hook gets fired again the component goes out of error. The errors are either `SeverSelectionError` or `OperationFailure` (code 18) indicating the cluster is still syncing either password or internal membership. The current check for `cluster_password_synced` is not robust enough to catch these.

## Solution
Update `cluster_password_synced` and its dependent functions to catch these errors right away.

### DIFF
--- a/lib/charms/mongodb/v1/shards_interface.py
+++ b/lib/charms/mongodb/v1/shards_interface.py
@@ -451,14 +451,13 @@ class ShardingProvider(Object):
         if not self.charm.is_role(Config.Role.CONFIG_SERVER):
             return True
 
-        # base case: no cluster relation
-        if not self.model.relations[self.relation_name]:
-            return True
-
         try:
-            # check our ability to use connect to cluster
+            # check our ability to use connect to mongos
             with MongosConnection(self.charm.mongos_config) as mongos:
                 mongos.get_shard_members()
+            # check our ability to use connect to mongod
+            with MongoDBConnection(self.charm.mongodb_config) as mongod:
+                mongod.get_replset_status()
         except OperationFailure as e:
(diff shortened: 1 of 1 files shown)


delgod (approved):

lib/charms/mongodb/v1/shards_interface.py
```diff
@@ -1017,7 +1020,7 @@ def cluster_password_synced(self) -> bool:
 
         try:
             # check our ability to use connect to both mongos and our current replica set.
-            mongos_reachable = self._is_mongos_reachable()
+            mongos_reachable = self._is_mongos_reachable(with_auth=True)
             with MongoDBConnection(self.charm.mongodb_config) as mongo:
                 mongod_reachable = mongo.is_ready
         except OperationFailure as e:
```
Mehdi-Bendriss, line 1026: Is there are reason why we `raise` 3 lines below this one? as opposed to returning `False` (seeing that this method is only called in `update_status`)

Similary on `ShardingProvider.cluster_password_synced`

Mehdi-Bendriss (commented):
Thanks Mia! I have a question

lib/charms/mongodb/v1/shards_interface.py
MiaAltieri: Yes - errors outside of: Authentication and Failures with Internal membership should be reported + raised - because they are related to something different entirely.

This is used in update status only to report Authentication Failures with Failures Internal membership

Mehdi-Bendriss (approved):

PR #381, opened 2024-03-20
[DPE-3839] restoring backups with name mismatch

MiaAltieri:
## Issue
Migrating a cluster to a cluster with a new name is not supported

## Solution
support that via passing an option for it

### DIFF
--- a/actions.yaml
+++ b/actions.yaml
@@ -5,7 +5,8 @@ get-primary:
   description: Report primary replica
 
 get-password:
-  description: Fetch the password of the provided internal user of the charm, used for internal charm operations. 
+  description:
+    Fetch the password of the provided internal user of the charm, used for internal charm operations.
     It is for internal charm users only, and SHOULD NOT be used by applications.
   params:
     username:
@@ -39,6 +40,11 @@ restore:
     backup-id:
       type: string
       description: A backup-id to identify the backup to restore. Format of <%Y-%m-%dT%H:%M:%SZ>
+    remap-pattern:
+      type: string
+      description:
+        Optional, a pattern used to remap cluster component names when performing a restore.
+        Format of old_config_server_name=new_config_server_name,old_shard_name=new_shard_name
 
 set-tls-private-key:
   description: Set the privates key, which will be used for certificate signing requests (CSR). Run for each unit separately.
--- a/lib/charms/mongodb/v1/mongodb_backups.py
+++ b/lib/charms/mongodb/v1/mongodb_backups.py
@@ -141,13 +141,13 @@ class MongoDBBackups(Object):
             return
 
         if not self.charm.db_initialised:
-            self._defer_action_with_info_log(
+            self._defer_event_with_info_log(
                 event, action, "Set PBM credentials, MongoDB not ready."
             )
             return
 
         if not self.charm.has_backup_service():
-            self._defer_action_with_info_log(
+            self._defer_event_with_info_log(
                 event, action, "Set PBM configurations, pbm-agent service not found."
             )
             return
@@ -181,7 +181,7 @@ class MongoDBBackups(Object):
             return
 
         if isinstance(pbm_status, WaitingStatus):
-            self._defer_action_with_info_log(
+            self._fail_action_with_error_log(
                 event,
                 action,
                 "Sync-ing configurations needs more time, must wait before creating a backup.",
@@ -215,7 +215,7 @@ class MongoDBBackups(Object):
         self.charm.unit.status = pbm_status
 
         if isinstance(pbm_status, WaitingStatus):
-            self._defer_action_with_info_log(
+            self._fail_action_with_error_log(
                 event,
                 action,
                 "Sync-ing configurations needs more time, must wait before listing backups.",
@@ -244,7 +244,7 @@ class MongoDBBackups(Object):
         # sometimes when we are trying to restore pmb can be resyncing, so we need to retry
         try:
             backup_id = event.params.get("backup-id")
-            self._try_to_restore(backup_id, remapping_args=event.params.get("remap-pattern"))
+            self._restore(backup_id, remapping_args=event.params.get("remap-pattern"))
             self.charm.unit.status = MaintenanceStatus(
                 f"restore started/running, backup id:'{backup_id}'"
             )
@@ -284,7 +284,7 @@ class MongoDBBackups(Object):
             return False
 
         if isinstance(pbm_status, WaitingStatus):
-            self._defer_action_with_info_log(
+            self._fail_action_with_error_log(
                 event,
                 action,
                 "Sync-ing configurations needs more time, must wait before restoring.",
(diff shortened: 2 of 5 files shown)


actions.yaml
```diff
@@ -39,6 +40,11 @@ restore:
     backup-id:
       type: string
       description: A backup-id to identify the backup to restore. Format of <%Y-%m-%dT%H:%M:%SZ>
+    remap-pattern:
```
Mehdi-Bendriss, line 43: Could you please update the backup/restore spec with this change?

lib/charms/mongodb/v1/mongodb_backups.py
```diff
@@ -733,3 +754,20 @@ def process_pbm_error(self, pbm_status: Optional[_StrOrBytes]) -> str:
         elif "status code: 301" in error_message:
             message = "s3 configurations are incompatible."
         return message
+
+    def _needs_provided_remap_arguments(self, backup_id: str) -> bool:
+        """Returns true if remap arguments are needed to perform a restore command."""
+        pbm_status = self.charm.run_pbm_command(["status", "--out=json"])
+        pbm_status = json.loads(pbm_status)
+        backups = pbm_status["backups"]["snapshot"] or []
```
Mehdi-Bendriss: Would "snapshot" be `None` here or would the "snapshot" key be missing altogether? if the latter I suggest using `pbm_status["backups"].get("snapshot", [])`

lib/charms/mongodb/v1/mongodb_backups.py
```diff
@@ -733,3 +754,20 @@ def process_pbm_error(self, pbm_status: Optional[_StrOrBytes]) -> str:
...
+        for backup in backups:
+            if not backup_id == backup["name"]:
+                continue
+
+            backup_status = backup.get("error", "")
+            break
```
Mehdi-Bendriss: Perhaps:
```suggestion
            if backup_id == backup["name"]:
                backup_status = backup.get("error", "")
                break
```

lib/charms/mongodb/v1/mongodb_backups.py
```diff
@@ -733,3 +754,20 @@ def process_pbm_error(self, pbm_status: Optional[_StrOrBytes]) -> str:
...
+
+        # When a charm is running as a Replica set it can generate its own remapping arguments
+        return self._backup_from_different_cluster(backup_status) and self.charm.is_role(
```
Mehdi-Bendriss: perhaps it would be more clear to prefix `_backup_from_different_cluster` with an `is_`

lib/charms/mongodb/v1/mongodb_backups.py
```diff
@@ -536,7 +554,7 @@ def _backup_from_different_cluster(self, backup_status: str) -> bool:
         """Returns if a given backup was made on a different cluster."""
         return re.search(REMAPPING_PATTERN, backup_status) is not None
 
-    def _try_to_restore(self, backup_id: str) -> None:
+    def _try_to_restore(self, backup_id: str, remapping_args: Optional[str] = None) -> None:
```
Mehdi-Bendriss: we could perhaps get away with renaming it to `_restore`, right?

actions.yaml
```diff
@@ -5,7 +5,8 @@ get-primary:
   description: Report primary replica
 
 get-password:
-  description: Fetch the password of the provided internal user of the charm, used for internal charm operations. 
+  description:
```
dmitry-ratushnyy, line 8: Do we need new line here or it was accidental formatting change?

lib/charms/mongodb/v1/mongodb_backups.py
```diff
@@ -54,7 +54,7 @@
     "storage-class": "storage.s3.storageClass",
 }
 S3_RELATION = "s3-credentials"
-REMAPPING_PATTERN = r"\ABackup doesn't match current cluster topology - it has different replica set names. Extra shards in the backup will cause this, for a simple example. The extra/unknown replica set names found in the backup are: ([^,\s]+)([.] Backup has no data for the config server or sole replicaset)?\Z"
+REMAPPING_PATTERN = r"\ABackup doesn't match current cluster topology - it has different replica set names. Extra shards in the backup will cause this, for a simple example. The extra/unknown replica set names found in the backup are: ([\w\d\-,\s]+)([.] Backup has no data for the config server or sole replicaset)?\Z"
```
dmitry-ratushnyy, line 57: I would be helpful for future us to have an example of what this re matches 😄

lib/charms/mongodb/v1/mongodb_backups.py
```diff
@@ -733,3 +754,20 @@ def process_pbm_error(self, pbm_status: Optional[_StrOrBytes]) -> str:
         elif "status code: 301" in error_message:
             message = "s3 configurations are incompatible."
         return message
+
+    def _needs_provided_remap_arguments(self, backup_id: str) -> bool:
```
Mehdi-Bendriss, line 751: Potentially for another PR: it seems that a big portion of `_remap_replicaset()` and this method are duplicated, would be nice to unify them and reuse

Mehdi-Bendriss (commented):
Thanks Mia - i have a couple of questions

lib/charms/mongodb/v1/mongodb_backups.py
dmitry-ratushnyy: is it always guaranteed that `pbm_status` will contains these keys and that `pbm_status["backups"]` will contain "snapshot" key or it is possible that this will trigger KeyError exception?

actions.yaml
MiaAltieri: my formatter did this

lib/charms/mongodb/v1/mongodb_backups.py
MiaAltieri: easy! done

lib/charms/mongodb/v1/mongodb_backups.py
MiaAltieri: At the point in which this is executed it should always have both. But I made Mehdi's suggestion for "snapshot"

lib/charms/mongodb/v1/mongodb_backups.py
MiaAltieri: ah - yes

lib/charms/mongodb/v1/mongodb_backups.py
```diff
@@ -141,13 +141,13 @@ def _on_s3_credential_changed(self, event: CredentialsChangedEvent):
...
         if not self.charm.db_initialised:
-            self._defer_action_with_info_log(
+            self._defer_event_with_info_log(
                 event, action, "Set PBM credentials, MongoDB not ready."
             )
             return
 
         if not self.charm.has_backup_service():
-            self._defer_action_with_info_log(
+            self._defer_event_with_info_log(
```
dmitry-ratushnyy, line 150: do we need to test here that event is not an action?

lib/charms/mongodb/v1/mongodb_backups.py
```diff
@@ -141,13 +141,13 @@ def _on_s3_credential_changed(self, event: CredentialsChangedEvent):
             return
 
         if not self.charm.db_initialised:
-            self._defer_action_with_info_log(
+            self._defer_event_with_info_log(
```
dmitry-ratushnyy, line 144: do we need to test here that event is not an action?

lib/charms/mongodb/v1/mongodb_backups.py
MiaAltieri: IMO no - the function that this event is a triggered by a hook firing from the s3-integrator , not an action

lib/charms/mongodb/v1/mongodb_backups.py
MiaAltieri: IMO no - the function that this event is a triggered by a hook firing from the s3-integrator , not an action

dmitry-ratushnyy (approved):

Mehdi-Bendriss (approved):

PR #384, opened 2024-03-21
[DPE-3388] Write logs to files

dmitry-ratushnyy:
## Issue
We need to write mongod logs and audit logs into files instead of syslog.

## Solution
Write logs to files located on a separate mount point
Screenshot 2024-04-05 at 10 47 21

### DIFF
--- a/lib/charms/mongodb/v1/helpers.py
+++ b/lib/charms/mongodb/v1/helpers.py
@@ -67,6 +67,7 @@ def _get_audit_log_settings(snap_install: bool) -> List[str]:
         f"--auditFormat={Config.AuditLog.FORMAT}",
     ]
 
+
 # noinspection GrazieInspection
 def get_create_user_cmd(config: MongoDBConfiguration, mongo_path=MONGO_SHELL) -> List[str]:
     """Creates initial admin user for MongoDB.
@@ -189,7 +190,7 @@ def get_mongod_args(
         # for simplicity we run the mongod daemon on shards, configsvrs, and replicas on the same
         # port
         f"--port={Config.MONGODB_PORT}",
-        "--setParameter processUmask=037", # required for log files perminission (g+r)
+        "--setParameter processUmask=037",  # required for log files perminission (g+r)
         logging_options,
     ]
     cmd.extend(audit_log_settings)
--- a/lib/charms/mongodb/v1/helpers.py
+++ b/lib/charms/mongodb/v1/helpers.py
@@ -46,20 +46,26 @@ MONGO_SHELL = "charmed-mongodb.mongosh"
 
 DATA_DIR = "/var/lib/mongodb"
 LOG_DIR = "/var/log/mongodb"
-LOG_TO_SYSLOG = True
 CONF_DIR = "/etc/mongod"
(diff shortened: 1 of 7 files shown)


lib/charms/mongodb/v1/helpers.py
```diff
@@ -46,19 +46,27 @@
...
 DATA_DIR = "/var/lib/mongodb"
 LOG_DIR = "/var/log/mongodb"
-LOG_TO_SYSLOG = True
 CONF_DIR = "/etc/mongod"
 MONGODB_LOG_FILENAME = "mongodb.log"
 logger = logging.getLogger(__name__)
 
 
 def _get_logging_options(snap_install: bool) -> str:
-    # TODO sending logs to syslog until we have a separate mount point for logs
```
Mehdi-Bendriss, line 56: Could you add a short docstring to this function?

lib/charms/mongodb/v1/helpers.py
```diff
@@ -46,19 +46,27 @@
...
-    # the output of the container. To enable logging to a file it must be set explicitly
-    return f"--logpath={LOG_DIR}/{MONGODB_LOG_FILENAME}" if snap_install else ""
+    log_path = f"{LOG_DIR}/{MONGODB_LOG_FILENAME}"
+    if snap_install:
+        log_path = f"{MONGODB_COMMON_DIR}{log_path}"
+    return f"--logpath={log_path}"
+
+
+def _get_audit_log_settings(snap_install: bool) -> List[str]:
+    audit_log_path = f"{LOG_DIR}/{Config.AuditLog.FILE_NAME}"
```
Mehdi-Bendriss, line 72: Could you add a short docstring here?

Mehdi-Bendriss (dismissed):
Thank you! I left a couple of minor comments

lib/charms/mongodb/v1/helpers.py
MiaAltieri: ➕

MiaAltieri (dismissed):

metadata.yaml
```diff
@@ -34,6 +34,9 @@ storage:
   mongodb:
     type: filesystem
     location: /var/snap/charmed-mongodb/common
+  logs:
+    type: filesystem
```
delgod: is juju refresh possible to version with changed storages list?

delgod (changes requested):

metadata.yaml
phvalguima: @delgod, indeed juju will not accept that. It runs a mountpoint validation [here](https://github.com/juju/juju/blob/1c4005149780c7d359c9b3a38c4b0371f9eeb1ed/state/filesystem.go#L1476-L1481). If one mount point contains the other, it will fail.

The only alternative is to update the `mongodb` mount point to a subfolder that does not contain logs's. However, that will be a breaking change between revisions. The upper side is keeping IOPS of logging and data separated.

dmitry-ratushnyy: Removed the mount point for now

delgod (approved):

Mehdi-Bendriss (approved):

MiaAltieri (approved):

PR #386, opened 2024-03-25
[DPE-3753] Support mongos tls

MiaAltieri:
## Issue
mongos charm does not support TLS and we want to re-use some of the shared libs owned by the mongodb chamr

## Solution
update the shared lib owned by the mongodb charm so that mongos charm can easily support TLS 

## Future PRs:
1. add libs to mongos + add necessary changes to mongos src/charm
2. add sanity checks for mongos checking valid TLS configuration
3. Int tests

### DIFF
--- a/lib/charms/mongodb/v0/config_server_interface.py
+++ b/lib/charms/mongodb/v0/config_server_interface.py
@@ -180,7 +180,7 @@ class ClusterProvider(Object):
             if new_ca is None:
                 self.database_provides.delete_relation_data(relation.id, {INT_TLS_CA_KEY: new_ca})
             else:
-                self._update_relation_data(relation.id, {INT_TLS_CA_KEY: new_ca})
+                self.database_provides.update_relation_data(relation.id, {INT_TLS_CA_KEY: new_ca})
 
 
 class ClusterRequirer(Object):
--- a/lib/charms/mongodb/v0/config_server_interface.py
+++ b/lib/charms/mongodb/v0/config_server_interface.py
@@ -18,6 +18,7 @@ from ops.charm import CharmBase, EventBase, RelationBrokenEvent
 from ops.framework import Object
 from ops.model import ActiveStatus, BlockedStatus, MaintenanceStatus, WaitingStatus
 
+from typing import Optional
 from config import Config
 
 logger = logging.getLogger(__name__)
@@ -26,6 +27,7 @@ KEY_FILE = "keyFile"
 HOSTS_KEY = "host"
 CONFIG_SERVER_DB_KEY = "config-server-db"
 MONGOS_SOCKET_URI_FMT = "%2Fvar%2Fsnap%2Fcharmed-mongodb%2Fcommon%2Fvar%2Fmongodb-27018.sock"
+INT_TLS_CA_KEY = f"int-{Config.TLS.SECRET_CA_LABEL}"
 
 # The unique Charmhub library identifier, never change it
 LIBID = "58ad1ccca4974932ba22b97781b9b2a0"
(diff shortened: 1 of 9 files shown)


lib/charms/mongodb/v1/mongos.py
```diff
@@ -61,7 +61,7 @@ def uri(self):
         # Auth DB should be specified while user connects to application DB.
         auth_source = ""
         if self.database != "admin":
-            auth_source = "&authSource=admin"
+            auth_source = "authSource=admin"
```
dmitry-ratushnyy, line 64: why the "&" is removed?

MiaAltieri: when `admin` isn't the database this `&` causes immediate failure due to PyMongo URI parsing and the charm becomes unrecoverable.

dmitry-ratushnyy: got it, thank you for the explanation

dmitry-ratushnyy (dismissed):

lib/charms/mongodb/v0/config_server_interface.py
```diff
@@ -106,16 +108,22 @@ def _on_relation_changed(self, event) -> None:
...
+        }
+
+        # if tls enabled
+        int_tls_ca = self.charm.tls.get_tls_secret(
+            internal=True, label_name=Config.TLS.SECRET_CA_LABEL
+        )
+        if int_tls_ca:
+            relation_data[INT_TLS_CA_KEY] = int_tls_ca
+
         if self.charm.unit.is_leader():
```
Mehdi-Bendriss: should we perhaps check if not-leader and `return` before line 111 (where you create relation_data)

lib/charms/mongodb/v0/config_server_interface.py
```diff
@@ -319,4 +335,12 @@ def update_keyfile(self, key_file_contents: str) -> bool:
 
         return True
 
+    def get_config_server_name(self) -> Optional[str]:
+        """Returns the name of the Juju Application that mongos is using as a config server."""
+        if not self.model.get_relation(self.relation_name):
+            return None
+
+        # metadata.yaml prevents having multiple config servers
+        return self.charm.model.relations[self.relation_name][0].app.name
```
Mehdi-Bendriss: perhaps:
```suggestion
        return self.model.get_relation(self.relation_name).app.name
```

lib/charms/mongodb/v0/mongodb_tls.py
```diff
@@ -187,13 +209,12 @@ def _on_certificate_available(self, event: CertificateAvailableEvent) -> None:
...
 
-        with MongoDBConnection(self.charm.mongodb_config) as mongo:
-            if not mongo.is_ready:
-                self.charm.unit.status = WaitingStatus("Waiting for MongoDB to start")
-            else:
-                self.charm.unit.status = ActiveStatus()
+        if not self.charm.is_db_service_ready():
+            self.charm.unit.status = WaitingStatus("Waiting for MongoDB to start")
+        else:
+            self.charm.unit.status = ActiveStatus()
```
Mehdi-Bendriss, line 218: I think we should clear the status `Waiting for MongoDB to start` (if set) instead of setting ActiveStatus (this could override an existing status set by another component)

src/charm.py
```diff
@@ -1486,6 +1486,29 @@ def is_sharding_component(self) -> bool:
...
+        with MongoDBConnection(self.mongodb_config) as mongod:
+            mongod_ready = mongod.is_ready
+
+        if not self.is_role(Config.Role.CONFIG_SERVER):
+            return mongod_ready
+
+        with MongoDBConnection(self.mongos_config) as mongos:
+            mongos_ready = mongos.is_ready
+
+        return mongod_ready and mongos_ready
```
Mehdi-Bendriss: perhaps:
```suggestion
            return mongod_ready and mongos.is_ready
```

Mehdi-Bendriss (commented):
Thanks Mia! just a couple of remarks

lib/charms/mongodb/v0/config_server_interface.py
MiaAltieri: ah actually it isn't needed to check since its done in `pass_hooks_checks` - I will remove this

Mehdi-Bendriss (approved):

dmitry-ratushnyy (approved):

PR #388, opened 2024-03-25
[DPE-3753] mongos must check tls inconsistencies

MiaAltieri:
## Issue
mongos charm must check + report tls inconsistencies

## Solution
add a check in the shared lib code

### DIFF
--- a/lib/charms/mongodb/v0/config_server_interface.py
+++ b/lib/charms/mongodb/v0/config_server_interface.py
@@ -298,7 +298,7 @@ class ClusterRequirer(Object):
     # BEGIN: helper functions
     def pass_hook_checks(self, event):
         """Runs the pre-hooks checks for ClusterRequirer, returns True if all pass."""
-        if self.is_mongos_tls_needed():
+        if self.is_mongos_tls_missing():
             logger.info(
                 "Deferring %s. Config-server uses TLS, but mongos does not. Please synchronise encryption methods.",
                 str(type(event)),
@@ -306,7 +306,7 @@ class ClusterRequirer(Object):
             event.defer()
             return False
 
-        if self.is_config_server_tls_needed():
+        if self.is_config_server_tls_missing():
             logger.info(
                 "Deferring %s. mongos uses TLS, but config-server does not. Please synchronise encryption methods.",
                 str(type(event)),
@@ -366,10 +366,10 @@ class ClusterRequirer(Object):
 
     def get_tls_statuses(self) -> Optional[StatusBase]:
         """Returns statuses relevant to TLS."""
-        if self.is_mongos_tls_needed():
+        if self.is_mongos_tls_missing():
             return BlockedStatus("mongos requires TLS to be enabled.")
 
-        if self.is_config_server_tls_needed():
+        if self.is_config_server_tls_missing():
             return BlockedStatus("mongos has TLS enabled, but config-server does not.")
 
         if not self.is_ca_compatible():
@@ -404,12 +404,12 @@ class ClusterRequirer(Object):
         )
 
         # base-case: missing one or more CA's to compare
-        if not config_server_tls_ca or not mongos_tls_ca:
+        if not config_server_tls_ca and not mongos_tls_ca:
             return True
 
         return config_server_tls_ca == mongos_tls_ca
 
-    def is_mongos_tls_needed(self) -> bool:
+    def is_mongos_tls_missing(self) -> bool:
         """Returns true if the config-server has TLS enabled but mongos does not."""
         config_server_relation = self.charm.model.get_relation(self.relation_name)
         if not config_server_relation:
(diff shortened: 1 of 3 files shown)


lib/charms/mongodb/v0/config_server_interface.py
```diff
@@ -334,4 +388,57 @@ def get_config_server_name(self) -> Optional[str]:
...
+            internal=True, label_name=Config.TLS.SECRET_CA_LABEL
+        )
+
+        # base-case: missing one or more CA's to compare
+        if not config_server_tls_ca or not mongos_tls_ca:
+            return True
+
+        return config_server_tls_ca == mongos_tls_ca
+
+    def is_mongos_tls_needed(self) -> bool:
```
Mehdi-Bendriss: perhaps we should rename this method to: `def is_mongos_tls_missing(self) -> bool:` ?

lib/charms/mongodb/v0/config_server_interface.py
```diff
@@ -334,4 +388,57 @@ def get_config_server_name(self) -> Optional[str]:
...
+            config_server_relation.id, INT_TLS_CA_KEY
+        )
+
+        mongos_tls_ca = self.charm.tls.get_tls_secret(
+            internal=True, label_name=Config.TLS.SECRET_CA_LABEL
+        )
+
+        # base-case: missing one or more CA's to compare
+        if not config_server_tls_ca or not mongos_tls_ca:
+            return True
```
Mehdi-Bendriss, line 408: should this perhaps be: `and`? I assume if one CA set and not the other the method should return False?

lib/charms/mongodb/v0/config_server_interface.py
```diff
@@ -326,6 +364,22 @@ def update_keyfile(self, key_file_contents: str) -> bool:
...
+        if self.is_config_server_tls_needed():
+            return BlockedStatus("mongos has TLS enabled, but config-server does not.")
+
+        if not self.is_ca_compatible():
+            logger.error(
+                "mongos is integrated to a different CA than the config server. Please use the same CA for all cluster components."
+            )
+            return BlockedStatus("mongos CA and Config-Server CA don't match.")
+
+        return
```
Mehdi-Bendriss, line 381: nit:
```suggestion
        return None
```

lib/charms/mongodb/v0/config_server_interface.py
```diff
@@ -334,4 +388,57 @@ def get_config_server_name(self) -> Optional[str]:
...
+        config_server_has_tls = (
+            self.database_requires.fetch_relation_field(config_server_relation.id, INT_TLS_CA_KEY)
+            is not None
+        )
+        if config_server_has_tls and not mongos_has_tls:
+            return True
+
+        return False
+
+    def is_config_server_tls_needed(self) -> bool:
```
Mehdi-Bendriss: perhaps we should rename this method to: `def is_config_server_tls_missing(self) -> bool:` ?

lib/charms/mongodb/v0/config_server_interface.py
```diff
@@ -286,6 +296,34 @@ def _on_relation_broken(self, event: RelationBrokenEvent) -> None:
...
 
     # BEGIN: helper functions
+    def pass_hook_checks(self, event):
+        """Runs the pre-hooks checks for ClusterRequirer, returns True if all pass."""
+        if self.is_mongos_tls_needed():
+            logger.info(
+                "Deferring %s. Config-server uses TLS, but mongos does not. Please synchronise encryption methods.",
+                str(type(event)),
+            )
+            event.defer()
```
Mehdi-Bendriss, line 306: nit: given this and all subsequent conditions evaluating to false trigger an event deferral, would it make sense to move the event.defer to the hook itself ?

Mehdi-Bendriss (commented):
Thank you Mia! I have a couple of questions

lib/charms/mongodb/v0/config_server_interface.py
MiaAltieri: Sure!

lib/charms/mongodb/v0/config_server_interface.py
MiaAltieri: hmmm - let me see, IIRC there was a reason for the OR, but you have a solid argument here 

I will test this change in the `mongos` charm and confirm that it works here :)

lib/charms/mongodb/v0/config_server_interface.py
MiaAltieri: we could, but in the future we may need to change it back, if there ends up being a case where an event warrants not needing to defer.

We have other `pass_hook` functions in other areas of Charmed MongoDB, for example in`lib/charms/mongodb/v1/sharding.py`. In that lib they are responsible for defering events when necessary. I would rather have code consistency across the multiple `pass_hook` functions in the code base than clean up the deferals in this one case. 

If you feel strongly though I am more than happy to, just let me know :)

dmitry-ratushnyy (approved):

Mehdi-Bendriss (approved):

PR #392, opened 2024-04-04
Bump snap revision to add ulimit value

Mehdi-Bendriss:
## Issue
Bump snap revision to incorporate new version of snap with Ulimit values set


delgod (approved):

MiaAltieri (approved):
thanks

PR #395, opened 2024-04-10
[DPE-4136] Fix TLS tests + Migrate to self-signed-certificates + cleanup

Mehdi-Bendriss:
## Issue
- The `tls-certificate-operator` and tls v1 interface is deprecated
- we're using many old versions of python dependencies and charm libs 
- multiple parts of the TLS integ. tests are not asserted / awaited / correct, effectively not validating the behavior of TLS

## Solution
- removes dependencies to the deprecated `tls-certificate-operator` + v1 interface
- introduces the `self-signed-certificates operator`  + v3 interface
- upgrades charm libs 
- upgrade python dependencies 
- upgrade dp action libs to the latest
- fix  tests:
    - CSR retrievals from secrets
    - CA cert retrieval from units
    - missing `assert`s and `await`s on coroutines


requirements.txt
```diff
@@ -1,20 +1,21 @@
...
 pyyaml==6.0.1
-zipp==3.11.0
-pyOpenSSL==22.1.0
-typing-extensions==4.5.0 
+zipp==3.18.1
+pyOpenSSL==24.1.0
+typing-extensions==4.11.0
 parameterized==0.9.0
-pydantic==1.10.7
+poetry==1.8.2
```
MiaAltieri, line 22: thank you for this - we should make a jira ticket for upgrading to poetry

tests/integration/sharding_tests/test_sharding_tls.py
```diff
@@ -185,30 +185,30 @@ async def check_cluster_tls_disabled(ops_test: OpsTest) -> None:
...
-        )
+        ), f"Mongos TLS not disabled in unit {unit.name}"
 
 
 async def check_cluster_tls_enabled(ops_test: OpsTest) -> None:
     # check each replica set is running with TLS enabled
     for cluster_component in CLUSTER_COMPONENTS:
         for unit in ops_test.model.applications[cluster_component].units:
-            await tls_helpers.check_tls(
+            assert await tls_helpers.check_tls(
```
MiaAltieri, line 202: oof - cannot believe this was missing, thank you

tests/integration/tls_tests/helpers.py
```diff
@@ -122,35 +128,26 @@ def process_ls_time(ls_output):
...
     "ActiveEnterTimestamp=Thu 2022-09-22 10:00:00 UTC"
     time_as_str = "T".join(systemctl_output.split("=")[1].split(" ")[1:3])
-    d = datetime.strptime(time_as_str, "%Y-%m-%dT%H:%M:%S")
-    return d
+    return datetime.strptime(time_as_str, "%Y-%m-%dT%H:%M:%S")
 
 
-async def scp_file_preserve_ctime(ops_test: OpsTest, unit_name: str, path: str) -> int:
-    """Returns the unix timestamp of when a file was created on a specified unit."""
+async def scp_file_preserve_ctime(ops_test: OpsTest, unit_name: str, path: str) -> str:
```
MiaAltieri: I could be wrong - but it looks like this function is actually just used to read a file via copying over the contents. Instead I think we could remove it and adopt my suggestions below

tests/integration/tls_tests/helpers.py
```diff
@@ -161,55 +158,58 @@ async def check_certs_correctly_distributed(
...
-    internal_item = [
+    external_item = [
         data
         for data in certificates_data
-        if data["certificate_signing_request"].rstrip() == app_current_crt.rstrip()
+        if data["certificate_signing_request"].rstrip() == external_unit_csr.rstrip()
     ][0]
 
     # Get a local copy of the external cert
     external_copy_path = await scp_file_preserve_ctime(ops_test, unit.name, EXTERNAL_CERT_PATH)
```
MiaAltieri:
```suggestion
external_contents_file = await get_file_content(ops_test, unit.name, EXTERNAL_CERT_PATH)
```

tests/integration/tls_tests/helpers.py
```diff
@@ -161,55 +158,58 @@ async def check_certs_correctly_distributed(
...
 
     # Get the external cert value from the relation
-    relation_external_cert = "\n".join(external_item["chain"])
+    relation_external_cert = "\n".join(external_item["chain"]).strip()
 
     # CHECK: Compare if they are the same
     with open(external_copy_path) as f:
         external_contents_file = f.read()
```
MiaAltieri:
```suggestion
```

tests/integration/tls_tests/helpers.py
```diff
@@ -161,55 +158,58 @@ async def check_certs_correctly_distributed(
...
-        assert relation_external_cert == external_contents_file
+        assert (
+            relation_external_cert == external_contents_file
+        ), f"Relation Content:\n{relation_external_cert}\nFile Content:\n{external_contents_file}\nMismatch."
```
MiaAltieri:
```suggestion
assert (
      relation_external_cert == external_contents_file
 ), f"Relation Content:\n{relation_external_cert}\nFile Content:\n{external_contents_file}\nMismatch."
```

tests/integration/tls_tests/helpers.py
```diff
@@ -161,55 +158,58 @@ async def check_certs_correctly_distributed(
...
 
     # Get a local copy of the internal cert
     internal_copy_path = await scp_file_preserve_ctime(ops_test, unit.name, INTERNAL_CERT_PATH)
```
MiaAltieri: and then taking the same approach here

MiaAltieri (changes requested):

requirements.txt
Mehdi-Bendriss: Tracked in [DPE-4059](https://warthogs.atlassian.net/browse/DPE-4059)

tests/integration/tls_tests/helpers.py
Mehdi-Bendriss: You're right! I'll rename the helper to just `scp_file` and keep it there in case we need in in the future, but I'll switch to your suggestion

tests/integration/tls_tests/helpers.py
Mehdi-Bendriss: I just grouped them together in a loop since they're identical

requirements.txt
MiaAltieri: 🙏

MiaAltieri (dismissed):

MiaAltieri (approved):

PR #397, opened 2024-04-11
[DPE-3940] Support Upgrade

MiaAltieri:
## Issue
`juju refresh` is not supported

## Solution
support `juju refresh` by implementing functions from upgrade lib


requirements.txt
```diff
@@ -17,4 +17,6 @@ zipp==3.11.0
 pyOpenSSL==22.1.0
 typing-extensions==4.5.0 
 parameterized==0.9.0
+# Future PR - use poetry in MongoDB Charm
+poetry==1.8.2
```
MiaAltieri, line 21: this is a temp fix for the unit tests - in the future we should have a separate PR to bring the whole MongoDB repo to use poetry

src/config.py
```diff
@@ -17,21 +17,16 @@ class Config:
...
     DEPENDENCIES = {
         "mongod_service": {
             "dependencies": {},
             "name": "mongod",
-            # this should be">4,<8" - but we get pydantic.error_wrappers.ValidationError - resolve
-            # in DPE-3940
-            "upgrade_supported": ">4",
-            # this should be "6.0.6-5" - but we get pydantic.error_wrappers.ValidationError -
-            # resolve in DPE-3940
+            "upgrade_supported": ">=6.0.0,<7",
```
marcoppenheimer: **nitpick:** I think you can actually just do `^6` to say the same thing.

lib/charms/mongodb/v0/upgrade.py
```diff
@@ -65,30 +82,91 @@ def pre_upgrade_check(self) -> None:
             raise ClusterNotReadyError(message=default_message, cause="Cluster is not healthy")
 
         if not self.is_cluster_able_to_read_write():
-            raise ClusterNotReadyError(message=default_message, cause="Cluster cannot read/write")
+            raise ClusterNotReadyError(
+                message=default_message, cause="Cluster cannot read/write - please check logs"
```
marcoppenheimer, line 96: **suggestion:** I believe that `ClusterNotReadyError` also takes `resolution` as an arg. I think here - and in some of the others - you can help the user along with a resolution recommendation, e.g 'Run the check on the config leader' or whatever.

lib/charms/mongodb/v0/upgrade.py
```diff
@@ -65,30 +82,91 @@ def pre_upgrade_check(self) -> None:
...
     @override
     def build_upgrade_stack(self) -> list[int]:
+        """Builds an upgrade stack, specifying the order of nodes to upgrade."""
+        if self.charm.is_role(Config.Role.CONFIG_SERVER):
+            # TODO implement in a future PR a stack for shards and config server
+            pass
+        elif self.charm.is_role(Config.Role.REPLICATION):
+            return self.get_replica_set_upgrade_stack()
+
+    def get_replica_set_upgrade_stack(self) -> list[int]:
```
marcoppenheimer, line 123: **question:** Is the reason this is a separate method because you will eventually have it split by `Config.Role`?
Like:
```python
get_replica_set_upgrade_stack()
get_config_server_upgrade_stack()
...
```

src/charm.py
```diff
@@ -943,7 +947,7 @@ def _open_ports_tcp(self, ports: int) -> None:
                 logger.exception("failed opening port: %s", str(e))
                 raise
 
-    def _install_snap_packages(self, packages: List[str]) -> None:
+    def install_snap_packages(self, packages: List[str]) -> None:
```
marcoppenheimer: **todo:** The type-hint is wrong here I believe.

It should be `install_snap_packages(self, packages: list[tuple[str, str, str]]): -> None`

marcoppenheimer: Or, probably better, you add a type for it.

```python
# config.py
from typing import TypeAlias

Package: TypeAlias = tuple[str, str, str]


# charm.py
from config import Package

def install_snap_packages(self, packages: list[Package]): -> None
...

File: lib/charms/mongodb/v0/upgrade.py [hunk 6]
```diff
@@ -65,30 +82,91 @@ def pre_upgrade_check(self) -> None:
...
+            logger.error("Unable to install Snap")
+            self.set_unit_failed()
+            return
+
+        logger.info(f"{self.charm.unit.name} upgrading service...")
+        self.charm.restart_charm_services()
+
+        try:
+            logger.debug("Running post-upgrade check...")
+            for attempt in Retrying(
```
marcoppenheimer: **nitpick:** could this retrying block decorate `post_upgrade_check`? Or is this a preference thing?

marcoppenheimer (dismissed):
Left some minor comments, mostly unimportant but hopefully useful.

One thing that I didn't have the time to check thoroughly, are you ensuring that you're blocking events that CANNOT run mid-upgrade?
An obvious example is `restart` if you have such an event, but I'm sure there are others. 

A check would be `if not self.upgrade.idle: return|event.defer()`

lib/charms/mongodb/v0/upgrade.py
MiaAltieri: yes exactly!

lib/charms/mongodb/v0/upgrade.py
MiaAltieri: Its a actually an optimisation thing, restartiing mongod can be very costly and so its better to wrap just `post_upgrade_check`.

Great question!

lib/charms/mongodb/v0/upgrade.py
MiaAltieri: great suggestion! I will add wherever I can provide useful resolutions :)

src/charm.py
MiaAltieri: Ah yes, let me do this. Thank you Marc 🙇

MiaAltieri:
@marcoppenheimer  - thank you for reminder about checking for `idle`, let me do this in my hooks :)

lib/charms/mongodb/v0/upgrade.py
marcoppenheimer: What I meant was, you can decorate the actual function - https://github.com/canonical/kafka-operator/blob/002289f3952e9ab190c922a58be4e01f079d1b30/src/workload.py#L93-L104

MiaAltieri: Ah smart 🧠

src/charm.py
```diff
@@ -368,14 +368,15 @@ def _on_config_changed(self, event: ConfigChangedEvent) -> None:
...
-        if self.is_role_changed():
-            logger.error(
-                f"cluster migration currently not supported, cannot change from { self.model.config['role']} to {self.role}"
-            )
-            raise ShardingMigrationError(
-                f"Migration of sharding components not permitted, revert config role to {self.role}"
-            )
+        if self.upgrade.idle:
+            # TODO in the future (24.04) support migration of components
+            if self.is_role_changed():
```
```suggestion
        if self.upgrade.idle and self.is_role_changed():
            # TODO in the future (24.04) support migration of components
```

Mehdi-Bendriss (dismissed):
Thank you!

Mehdi-Bendriss (dismissed):

Mehdi-Bendriss (approved):

marcoppenheimer (approved):
🔥 💪🏾 🚀

PR #399, opened 2024-04-15
[DPE-3944] Add tests for upgrades

MiaAltieri:
## Issue
Tests for upgrades missing

## Solution
Add tests for upgrades

Integration tests made to test upgrading from latest stable, but since stable does not have the ability to upgrade we cannot run these tests (hence the `pytest.skip`)

### DIFF
--- a/tests/integration/upgrade/test_upgrade.py
+++ b/tests/integration/upgrade/test_upgrade.py
@@ -25,7 +25,6 @@ async def continuous_writes(ops_test: OpsTest):
 
 
 @pytest.mark.group(1)
-@pytest.mark.skip(reason="Must wait for stable channel to support upgrades before testing")
 async def test_build_and_deploy(ops_test: OpsTest) -> None:
     """Build and deploy one unit of MongoDB."""
     # it is possible for users to provide their own cluster for testing. Hence check if there
@@ -34,14 +33,16 @@ async def test_build_and_deploy(ops_test: OpsTest) -> None:
     if app_name:
         return await check_or_scale_app(ops_test, app_name)
 
-    await ops_test.model.deploy("mongodb", channel="stable", num_units=3)
+    # TODO: When `6/stable` track supports upgrades deploy and test that revision instead.
+    await ops_test.model.deploy("mongodb", channel="edge", num_units=3)
+
     await ops_test.model.wait_for_idle(
         apps=["mongodb"], status="active", timeout=1000, idle_period=120
     )
+    app_name = await get_app_name(ops_test)
 
 
 @pytest.mark.group(1)
-@pytest.mark.skip(reason="Must wait for stable channel to support upgrades before testing")
 async def test_upgrade(ops_test: OpsTest, continuous_writes) -> None:
     """Verifies that the upgrade can run successfully."""
     app_name = await get_app_name(ops_test)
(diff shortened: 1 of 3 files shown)


dmitry-ratushnyy (dismissed):

dmitry-ratushnyy (dismissed):

tests/integration/upgrade/test_upgrade.py
```diff
@@ -31,12 +34,41 @@ async def test_build_and_deploy(ops_test: OpsTest) -> None:
...
-    await ops_test.model.deploy(my_charm, num_units=3)
-    await ops_test.model.wait_for_idle()
+    await ops_test.model.deploy("mongodb", channel="stable", num_units=3)
+    await ops_test.model.wait_for_idle(
+        apps=["mongodb"], status="active", timeout=1000, idle_period=120
+    )
+
+
+@pytest.mark.group(1)
+@pytest.mark.skip(reason="Must wait for stable channel to support upgrades before testing")
```
Mehdi-Bendriss: Can we rely on `edge` so these tests get run?

tests/integration/upgrade/test_upgrade.py
```diff
@@ -31,12 +34,41 @@ async def test_build_and_deploy(ops_test: OpsTest) -> None:
...
+        apps=[app_name], status="active", timeout=1000, idle_period=120
+    )
+
+    new_charm = await ops_test.build_charm(".")
+    app_name = await get_app_name(ops_test)
+    await ops_test.model.applications[app_name].refresh(path=new_charm)
+    await ops_test.model.wait_for_idle(
+        apps=[app_name], status="active", timeout=1000, idle_period=120
+    )
+
```
Mehdi-Bendriss, line 65: I believe we should call `test_consistency_between_workload_and_metadata` here

Mehdi-Bendriss (commented):
Thanks Mia - I have a few comments with regards to the used channel for upgrades. In addition to this - I believe we should have a rollback test

tests/integration/upgrade/test_upgrade.py
MiaAltieri: `test_consistency_between_workload_and_metadata` is a sanity check that tests that:

when the developer updates the snap revision, that they should have upgraded the version in the variable DEPENDENCIES. If they do not then they will be mismatched and upgrades will not be checking feasible revisions properly. 

I know this is kind of confusing, so if a call would be better let me know

tests/integration/upgrade/test_upgrade.py
MiaAltieri: In theory yes, but we only support upgrades to `stable` revisions, I believe our tests should reflect that.

tests/integration/upgrade/test_upgrade.py
Mehdi-Bendriss: Sorry, I misread that function for a helper. Please disregard

tests/integration/upgrade/test_upgrade.py
MiaAltieri: yes  you are right! I will add a comment saying that we should use stable ASAP though :)

Mehdi-Bendriss (approved):

dmitry-ratushnyy (approved):

PR #404, opened 2024-04-19
[DPE-4155] Copy over upgrade code

MiaAltieri:
## Issue
Before implementing upgrades we need the "bones" of the upgrade code implemented by Carl

## Solution
1. Copy over bones from carl
2. change opensearch specific changes to VM specific changes
3. add notes for future PRs

## Future PRs 
1. Implement basic preflight check and tests
2. Implement upgrades 
3. Add int-tests in for upgrades

### DIFF
--- a/src/machine_upgrade.py
+++ b/src/machine_upgrade.py
@@ -1,0 +1,145 @@
+# Copyright 2023 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+"""In-place upgrades on machines
+
+Derived from specification: DA058 - In-Place Upgrades - Kubernetes v2
+(https://docs.google.com/document/d/1tLjknwHudjcHs42nzPVBNkHs98XxAOT2BXGGpP7NyEU/)
+"""
+import json
+import logging
+import time
+import typing
+
+import charms.opensearch.v0.constants_charm as constants_charm
+import ops
+
+import upgrade
+from opensearch import OpenSearchSnap
+
+logger = logging.getLogger(__name__)
+
+_SNAP_REVISION = str(constants_charm.OPENSEARCH_SNAP_REVISION)
+
+
+class Upgrade(upgrade.Upgrade):
+    """In-place upgrades on machines"""
+
+    @property
+    def unit_state(self) -> typing.Optional[str]:
+        if (
+            self._unit_workload_version is not None
+            and self._unit_workload_version != self._app_workload_version
+        ):
+            logger.debug("Unit upgrade state: outdated")
+            return "outdated"
+        return super().unit_state
+
+    @unit_state.setter
+    def unit_state(self, value: str) -> None:
+        if value == "healthy":
+            # Set snap revision on first install
+            self._unit_databag["snap_revision"] = _SNAP_REVISION
+            logger.debug(f"Saved {_SNAP_REVISION} in unit databag while setting state healthy")
+        # Super call
+        upgrade.Upgrade.unit_state.fset(self, value)
+
+    def _get_unit_healthy_status(
+        self, *, workload_status: typing.Optional[ops.StatusBase]
+    ) -> ops.StatusBase:
+        if self._unit_workload_version == self._app_workload_version:
+            if isinstance(workload_status, ops.WaitingStatus):
+                return ops.WaitingStatus(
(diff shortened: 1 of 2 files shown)


Mehdi-Bendriss (approved):
Thank you!

src/machine_upgrade.py
```diff
@@ -137,7 +140,9 @@ def authorized(self) -> bool:
                 return False
         return False
 
-    def upgrade_unit(self, *, snap: OpenSearchSnap) -> None:
+    def upgrade_unit(self, *, snap) -> None:
+        """Runs the upgrade procedure."""
+        # TODO: Future PR - run mongodb specific checks for upgrade
```
carlcsaposs-canonical, line 145: fyi, preupgrade check will be added to upgrade.py in opensearch PR soon

src/upgrade.py
```diff
@@ -25,21 +24,21 @@
...
 def unit_number(unit_: ops.Unit) -> int:
-    """Get unit number"""
+    """Get unit number."""
     return int(unit_.name.split("/")[-1])
 
 
-class PeerRelationNotReady(Exception):
-    """Upgrade peer relation not available (to this unit)"""
+class PeerRelationNotReadyError(Exception):
+    """Upgrade peer relation not available (to this unit)."""
```
carlcsaposs-canonical, line 32: nit: it's not an error

it's an expected exceptional case (during initial start)

carlcsaposs-canonical: also, where possible, I would avoid changing the public interface of upgrade.py if not needed, since it'll reduce reusability if we extract upgrade.py into a shared lib/package in the future

src/upgrade.py
```diff
@@ -85,6 +85,7 @@ def is_compatible(self) -> bool:
             key: poetry_version.Version.parse(value) for key, value in current_version_strs.items()
         }
         try:
+            # TODO Future PR: change this > sign to support downgrades
```
carlcsaposs-canonical, line 88: FYI, it appears mongodb only supports single version downgrades (https://www.mongodb.com/docs/manual/release-notes/6.0-downgrade/)
(and doesn't support downgrades after 7.0 https://www.mongodb.com/docs/manual/release-notes/7.0-downgrade/)

you may not want to remove this check completely

src/upgrade.py
```diff
@@ -120,24 +122,26 @@ def in_progress(self) -> bool:
...
         self, *, workload_status: typing.Optional[ops.StatusBase]
     ) -> ops.StatusBase:
-        """Status shown during upgrade if unit is healthy"""
+        """Status shown during upgrade if unit is healthy."""
 
     def get_unit_juju_status(
         self, *, workload_status: typing.Optional[ops.StatusBase]
     ) -> typing.Optional[ops.StatusBase]:
+        """Returns the unit status for the upgrade."""
         # TODO: revise status handling & priority
```
carlcsaposs-canonical, line 138: fyi, it's a bit harder to track the diff since opensearch was used for the first commit instead of router (and opensearch is in active development)

for example, this line isn't in router & was removed from opensearch (but doesn't show in the diff https://github.com/canonical/mongodb-operator/pull/404/files/85e769ce8bfee89606bb42f7bf3033cabd65a3f8..HEAD)

src/machine_upgrade.py
```diff
@@ -137,7 +140,9 @@ def authorized(self) -> bool:
...
         logger.debug(f"Upgrading {self.authorized=}")
         self.unit_state = "upgrading"
         snap.install()
```
carlcsaposs-canonical, line 148: in future PR, remember to stop & restart mongodb before refreshing snap

src/machine_upgrade.py
```diff
@@ -11,22 +11,22 @@
...
 import ops
 
 import upgrade
-from opensearch import OpenSearchSnap
+from config import Config
 
 logger = logging.getLogger(__name__)
 
-_SNAP_REVISION = str(constants_charm.OPENSEARCH_SNAP_REVISION)
+_SNAP_REVISION = str(Config.SNAP_PACKAGES[0][3])
```
carlcsaposs-canonical:
```suggestion
_SNAP_REVISION = str(Config.SNAP_PACKAGES[0][2])
```
think this might be off by 1

Issue #398, opened 2024-04-12
Import of big data sets


dmitry-ratushnyy:
## Steps to reproduce
1. Deploy MongoDB cluster, for example on LXD
2. Obtain connection URI
3. Install mongosh on host machine in case of LXD or copy your data set into juju unit
4. Using "charmed-mongodb.mongoimport" (from inside of juju unit) or mongoimport (from host machine) start import data
5. Periodically check free space on disk. 

## Expected behavior
Import of 70 GB of data should not consume 500+GB of disk space

## Actual behavior
Disk space is being consumed very quickly. This most likely happens due to audit logs. 
Log rotation and  allowing users to opt out from audit logging  should solve the problem

## Versions
stable/164

Operating system: 

Juju CLI: 

Juju agent: 

Charm revision: 

LXD: 

## Log output

Juju debug log:

dmitry-ratushnyy:
We are working on resolving this issue

PR #409, opened 2024-04-23
Add log rotation with "logrotate"

dmitry-ratushnyy:
## Issue
[ISSUE-398](https://github.com/canonical/mongodb-operator/issues/398)
At the current moment Charmed MongoDB logs, can grow very fast, especially audit log.
For example when a user imports data sets.

## Solution
Support log rotation to avoid consuming too much space with log files.
Log will be rotated if they grow bigger than 100M, the check for log files size will be performed every minute

## Result of log rotation
Screenshot 2024-04-23 at 18 59 47


src/machine_helpers.py
```diff
@@ -28,3 +35,26 @@ def update_mongod_service(
...
+    with open("templates/logrotate.j2", "r") as file:
+        template = jinja2.Template(file.read())
+
+    rendered = template.render(
+        logs_directory=f"{MONGODB_COMMON_DIR}/{LOG_DIR}",
+        mongo_user=MONGO_USER,
+    )
+
+    with open("/etc/logrotate.d/mongodb", "w") as file:
+        file.write(rendered)
```
MiaAltieri: noob question. What is this? Can you add comments for it 🙏 

Specifically answering these questions:
1. what is the significance of `templates/logrotate.j2`, what is its orgin? 
2. what happens during rendering

src/machine_helpers.py
```diff
@@ -28,3 +35,26 @@ def update_mongod_service(
...
+
+    cron = (
+        "* 1-23 * * * root logrotate -f /etc/logrotate.d/mongodb\n"
+        "1-59 0 * * * root logrotate -f /etc/logrotate.d/mongodb\n"
+    )
```
MiaAltieri: more noob questions from someone who doesn't understand cron, can you add comments explaining:
1. what each of these two are
2. the significance of `1-23` and `1-59`

src/machine_helpers.py
```diff
@@ -28,3 +35,26 @@ def update_mongod_service(
     if role == Config.Role.CONFIG_SERVER:
         mongos_start_args = get_mongos_args(config, snap_install=True)
         add_args_to_env("MONGOS_ARGS", mongos_start_args)
+
+
+def setup_logrotate_and_cron() -> None:
+    """Create and write the logrotate config file."""
```
MiaAltieri: In this doc string can you specify:
1. how often the cron job runs (I believe this is 1min)?
2. at what point do we rotate (I belive after 100MB?)

templates/logrotate.j2
```diff
@@ -0,0 +1,16 @@
...
+  size 100M
+  missingok
+  notifempty
+  create 0600 {{mongo_user}} {{mongo_user}}
+  compress
+  nomail
+  nocopytruncate
+  sharedscripts
+  postrotate
+    sudo /bin/kill -SIGUSR1 $(sudo systemctl show -p MainPID --value snap.charmed-mongodb.mongod.service)
```
MiaAltieri: From what I understand about Unix / Linux log rotate it works when the kill -SIGUSR1 signal is sent to the process that writes logs(in this case mongod) and then the logs are rotated according to what settings are specified in `/etc/logrotate.d/mongodb`. 

Is this correct?

If so I have concerns about us killing `mongod` and in some cases forcing a primary re-election. I understand that this is similar to what [percona does](https://stackoverflow.com/questions/20162176/centos-linux-setting-logrotate-to-maximum-file-size-for-all-logs)

But is that what [mysql does](https://github.com/canonical/mysql-operator/blob/14e7dd5662a8b4fe224b59bc453fb3bda4738466/templates/logrotate.j2)? I genuinely can't tell what their script does (without doing a deep dive) 

Either way, I think we can do better. If we have this cron job running anyway, why not just delete the log file manually and then let mongodb write a new one, rather than sending SIGKILL? This would give us better performance regarding primary elections.

I could be naive here and missing a bigger picture though. But I am curious what your thoughts are @delgod and @Mehdi-Bendriss

MiaAltieri (changes requested):
Looks good, thank you so much for your work here it is great and a big work on your end. I am very grateful for it. 

There are a few comments, some nits and a bigger picture question (regarding sigkill). If we decide to address the bigger picture question later I won't have an issue merging.

But I do have a few other questions:

1. does `mongos` write to its own log file? If so then we should do similar operations to what you have done here for `mongod` logs
2. Is it possible to add an integration test for this, (maybe with copying over large data). We could add  `@pytest.mark.skip("Test imports big files, do not run on CI")` to the top so it doesn't run on CI and take up big resources.

src/machine_helpers.py
dmitry-ratushnyy: During rendering the template variables will be replaced with actual values as you can see.

`logrotate.j2`  is a template for configuration for logrotate based on different docs about log rotation for mongodb.

This approach (rendering template) and updating the cron job is the same as in MySQL

templates/logrotate.j2
dmitry-ratushnyy: This command does  [not kill mongodb process, it only sends singal to it to rotate the logs](https://www.mongodb.com/docs/manual/tutorial/rotate-log-files/#forcing-a-log-rotation-with-sigusr1), that's it. So no re-election will happen.

dmitry-ratushnyy: Mongodb will rotate logs according to [these](https://github.com/canonical/mongodb-operator/pull/409/files#diff-094641a8acb9699cf7cfb0984347bf0e0abf254adf46b8080311fc973a03daffR142) settings,

and `/etc/logrotate.d/mongodb` is configuration for `logrotate` only and not for mongo

src/machine_helpers.py
dmitry-ratushnyy: Let me actually simplify this

templates/logrotate.j2
dmitry-ratushnyy: Also : 

> When reporting activity data to a log file, by default, MongoDB only rotates logs in response to the [logRotate](https://www.mongodb.com/docs/manual/reference/command/logRotate/#mongodb-dbcommand-dbcmd.logRotate) command, or when the [mongod](https://www.mongodb.com/docs/manual/reference/program/mongod/#mongodb-binary-bin.mongod) or [mongos](https://www.mongodb.com/docs/manual/reference/program/mongos/#mongodb-binary-bin.mongos) process receives a SIGUSR1 signal from the operating system. Both server logs and [audit logs](https://www.mongodb.com/docs/manual/core/auditing/#std-label-auditing) may be rotated with the [logRotate](https://www.mongodb.com/docs/manual/reference/command/logRotate/#mongodb-dbcommand-dbcmd.logRotate) command, either together or independently.

src/machine_helpers.py
dmitry-ratushnyy: done

dmitry-ratushnyy:
> Looks good, thank you so much for your work here it is great and a big work on your end. I am very grateful for it.
> 
> There are a few comments, some nits and a bigger picture question (regarding sigkill). If we decide to address the bigger picture question later I won't have an issue merging.
> 
> But I do have a few other questions:
> 
> 1. does `mongos` write to its own log file? If so then we should do similar operations to what you have done here for `mongod` logs
> 2. Is it possible to add an integration test for this, (maybe with copying over large data). We could add  `@pytest.mark.skip("Test imports big files, do not run on CI")` to the top so it doesn't run on CI and take up big resources.



> Looks good, thank you so much for your work here it is great and a big work on your end. I am very grateful for it.
> 
> There are a few comments, some nits and a bigger picture question (regarding sigkill). If we decide to address the bigger picture question later I won't have an issue merging.
> 
> But I do have a few other questions:
> 
> 1. does `mongos` write to its own log file? If so then we should do similar operations to what you have done here for `mongod` logs

yes according to docs we need to sent same signal to `mongos`  but I think we can \ should do it in a separate PR

tests/integration/test_charm.py
```diff
@@ -391,3 +394,26 @@ async def test_audit_log(ops_test: OpsTest) -> None:
         item = json.loads(line)
         # basic sanity check
         assert audit_log_line_sanity_check(item), "Audit sanity log check failed for first line"
+
+
+@pytest.mark.group(1)
+async def test_log_roate(ops_test: OpsTest) -> None:
```
MiaAltieri:
```suggestion
async def test_log_rotate(ops_test: OpsTest) -> None:
```

tests/integration/test_charm.py
```diff
@@ -391,3 +394,26 @@ async def test_audit_log(ops_test: OpsTest) -> None:
...
+    """Test that log are being rotated."""
+    app_name = await get_app_name(ops_test)
+    await start_continous_writes(ops_test, 1)
+    # Note: this timeout out depends on max log size
+    # which is defined in "src/config.py::Config.MAX_LOG_SIZE"
+    time.sleep(60 * 10)
```
MiaAltieri: Can we either:
1. Add a constant `TIME_TO_WRITE_100MB`

or 

2. go into the charm and **temporarily** replace the `Config.MAX_LOG_SIZE` to `1MB` for the purpose of this test, so that we do not have to wait `60*10` for the test to run? I can imagine this will clog resources

tests/integration/test_charm.py
```diff
@@ -391,3 +394,26 @@ async def test_audit_log(ops_test: OpsTest) -> None:
...
+    await stop_continous_writes(ops_test, down_unit="mongodb/0", app_name=app_name)
+    # Just to make sure that logroate will run
+    time.sleep(60)
```
MiaAltieri: a constant to replace `60` with `TIME_FOR_LOG_ROTATE` would be great

tests/integration/test_charm.py
```diff
@@ -391,3 +394,26 @@ async def test_audit_log(ops_test: OpsTest) -> None:
...
+    await clear_db_writes(ops_test)
+    audit_log_snap_path = "/var/snap/charmed-mongodb/common/var/log/mongodb/"
+    log_files = check_output(
+        f"JUJU_MODEL={ops_test.model_full_name} juju ssh {app_name}/leader 'sudo ls {audit_log_snap_path}'",
+        stderr=subprocess.PIPE,
+        shell=True,
+        universal_newlines=True,
+    )
+    log_rotated = "audit.log.1.gz" in log_files
+    assert log_rotated, f"Could not find rotated log in {log_files}"
```
MiaAltieri: noob question: when log get rotated do they get added to a list of log files?

dmitry-ratushnyy: there will be created a new file named `audit.log.N.gz`,  where N is the number of rotation.

tests/integration/test_charm.py
dmitry-ratushnyy: Done

tests/integration/test_charm.py
dmitry-ratushnyy: Added constant

tests/integration/test_charm.py
dmitry-ratushnyy: fixed

templates/logrotate.j2
MiaAltieri: Ah thank you 🙇

tests/integration/test_charm.py
MiaAltieri: Also sorry, one more question, is 1minute gaurenteed to get us to the necessary number of writes? i.e. is it possible that the test could be flakey?

MiaAltieri: Also I think it would be best if we *eventually* did something like `go into the charm and temporarily replace the Config.MAX_LOG_SIZE to 1MB for the purpose of this test, so that we do not have to wait 60*10 for the test to run? I can imagine this will clog resources`

But no need to do immediately. Would you mind creating a JIRA ticket for it and if you have time before you leave, implementing it? Otherwise I will leave it on the backlog to be addressed later on :)

MiaAltieri (dismissed):

tests/integration/test_charm.py
dmitry-ratushnyy: 10 mins of writes using "continuous_writing" give more than 50MB of  audit.log,  another extra  minute is just for safety reason.

Adjusting size of the test will work can be done later, but it will work only for newly deployed clusters, not for existing ones.

tests/integration/ha_tests/helpers.py
```diff
@@ -612,10 +611,7 @@ async def update_service_logging(ops_test: OpsTest, unit, logging: bool):
             continue
         line = line.replace("\n", "")
 
-        if logging:
-            if LOGGING_OPTIONS not in line:
-                mongodb_service[index] = line + " " + LOGGING_OPTIONS + "\n"
-        else:
+        if not logging:
```
MiaAltieri: @dmitry-ratushnyy - I think it is enough to remove the function `update_service_logging ` all together.

IIRC - this function was made so that we could temporarily turn on logs for the tests. Now that we support logs, I think we no longer need to turn on / off logs for the tests.

Mind removing the usage of this function from the tests suite and confirming?

MiaAltieri (changes requested):

tests/integration/test_charm.py
```diff
@@ -319,6 +324,50 @@ async def test_no_password_change_on_invalid_password(ops_test: OpsTest) -> None
...
+    time.sleep(logrotate_timeout)  # Just to make sure that logroate will run
+    await clear_db_writes(ops_test)
+    audit_log_snap_path = "/var/snap/charmed-mongodb/common/var/log/mongodb/"
+    log_files = check_output(
+        f"JUJU_MODEL={ops_test.model_full_name} juju ssh {app_name}/leader 'sudo ls {audit_log_snap_path}'",
+        stderr=subprocess.PIPE,
+        shell=True,
+        universal_newlines=True,
+    )
+    log_rotated = "audit.log.1.gz" in log_files
```
Mehdi-Bendriss, line 381: Can we also add an assert with `audit.log in log_files`?

Mehdi-Bendriss (commented):
Thanks Dmitry! I have one minor comment

tests/integration/test_charm.py
```diff
@@ -319,6 +324,50 @@ async def test_no_password_change_on_invalid_password(ops_test: OpsTest) -> None
...
+
+@pytest.mark.group(1)
+async def test_log_rotate(ops_test: OpsTest) -> None:
+    """Test that log are being rotated."""
+    # Note: this timeout out depends on max log size
+    # which is defined in "src/config.py::Config.MAX_LOG_SIZE"
+    time_to_write_50m_of_data = 60 * 10
+    logrotate_timeout = 60
+    app_name = await get_app_name(ops_test)
+    await start_continous_writes(ops_test, 1)
```
Mehdi-Bendriss, line 368: can we add an assert that `"audit.log.1.gz" not in log_files` here?

dmitry-ratushnyy: done

tests/integration/test_charm.py
dmitry-ratushnyy: done

tests/integration/ha_tests/helpers.py
dmitry-ratushnyy: Done

Mehdi-Bendriss (dismissed):

MiaAltieri (dismissed):
excellent work! Thank you Dmtiry 🙇

MiaAltieri (approved):

Mehdi-Bendriss (approved):

PR #413, opened 2024-05-07
make necessary lib changes for k8s

MiaAltieri:
## Issue
MongoDB k8s charm does not support backups with current lib implementation

## Solution
1. add back in old version of libs 
2. make necessary changes to fix broken backups in k8s
3. bump libpatch so new version is published

## Easier diff
Diff of changes in last version of lib to new version can be [found here](https://github.com/canonical/mongodb-k8s-operator/pull/248/files)

## Why are you adding back in the old libapi?
The new LIBAPI introduces a breaking change by supporting sharding, old libs are used my the mongodb k8s charm. So in order to make those changes and publish them we must do so by adding them back to this repo

In the future (24.10) when we support sharding on MongoDB K8s charm we will remove these libs

### DIFF
--- a/lib/charms/mongodb/v0/helpers.py
+++ b/lib/charms/mongodb/v0/helpers.py
@@ -30,7 +30,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 8
+LIBPATCH = 9
 
 
 # path to store mongodb ketFile
--- a/lib/charms/mongodb/v0/helpers.py
+++ b/lib/charms/mongodb/v0/helpers.py
@@ -30,7 +30,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 9
+LIBPATCH = 8
 
 
 # path to store mongodb ketFile
--- a/lib/charms/mongodb/v0/helpers.py
+++ b/lib/charms/mongodb/v0/helpers.py
@@ -1,0 +1,274 @@
+"""Simple functions, which can be used in both K8s and VM charms."""
+
+# Copyright 2023 Canonical Ltd.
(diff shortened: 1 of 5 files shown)


lib/charms/mongodb/v0/helpers.py
```diff
@@ -0,0 +1,274 @@
...
+
+# The unique Charmhub library identifier, never change it
+LIBID = "b9a7fe0c38d8486a9d1ce94c27d4758e"
+
+# Increment this major API version when introducing breaking changes
+LIBAPI = 0
+
+# Increment this PATCH version before using `charmcraft publish-lib` or reset
+# to 0 if you are raising the major API version
+LIBPATCH = 9
```
dragomirp:
```
> charmcraft fetch-lib charms.mongodb.v0.helpers
Library charms.mongodb.v0.helpers version 0.7 downloaded.
```
I think the version here should be 8, I think charmhub will complain if you try to skip a version when publishing.

lib/charms/mongodb/v0/mongodb_backups.py
```diff
@@ -0,0 +1,677 @@
...
+
+# The unique Charmhub library identifier, never change it
+LIBID = "18c461132b824ace91af0d7abe85f40e"
+
+# Increment this major API version when introducing breaking changes
+LIBAPI = 0
+
+# Increment this PATCH version before using `charmcraft publish-lib` or reset
+# to 0 if you are raising the major API version
+LIBPATCH = 8
```
dragomirp:
```
> charmcraft fetch-lib charms.mongodb.v0.mongodb_backups
Library charms.mongodb.v0.mongodb_backups version 0.6 downloaded.
```
Like with the other lib, I think charm hub will complain for skipping a version. I guess we can publish manual versions to catch up if it does.

dragomirp (dismissed):

shayancanonical (dismissed):

shayancanonical (approved):

dragomirp (approved):

Mehdi-Bendriss (approved):

PR #415, opened 2024-05-27
[DPE-4434] fix lint in ci

MiaAltieri:
## Issue
linter is failing on all workflow runs. Linter complains about function names and string values (i.e. things it should not be) 

## Solution
Revert linter to an earlier revision (i.e. the one MySQLRouter uses)

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -42,10 +42,9 @@ deps =
     pyproject-flake8
     pep8-naming
     isort
-    codespell
+    codespell==2.2.5 # current version reports incorrectly reports issues in string names and function names
 commands =
-    codespell {[vars]all_path}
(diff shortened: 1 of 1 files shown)


tox.ini
```diff
@@ -42,10 +42,9 @@ deps =
     pyproject-flake8
     pep8-naming
     isort
-    codespell
+    codespell==2.2.5 # current version reports incorrectly reports issues in string names and function names
 commands =
-    codespell {[vars]all_path}
```
MiaAltieri, line 47: no need to run codespell twice

carlcsaposs-canonical (approved):

Mehdi-Bendriss (approved):

PR #416, opened 2024-05-27
[DPE-4157] Support juju refresh on replica set

MiaAltieri:
## Issue
Functionality for upgrades on Replica Set not implemented

## Solution
Add functionality for juju-refresh on replica set 

## Future work
Add in pre-upgrade check for replica set

## Integration Testing
Until a new revision of the charm is published that supports upgrades we cannot test this PR 

Charm was manually tested by building the charm with snap revision 117, then refreshing with a charm built with snap revision 118


src/upgrades/machine_upgrade.py
```diff
@@ -145,21 +145,47 @@ def authorized(self) -> bool:
...
+        # According to the MongoDB documentation, before upgrading the primary, we must ensure a
+        # safe primary re-election.
+        try:
+            if self._unit.name == charm.primary:
+                logger.debug("Stepping down current primary, before upgrading service...")
+                charm.upgrade.step_down_primary_and_wait_reelection()
+        except mongodb_upgrade.FailedToElectNewPrimaryError:
+            logger.error("Failed to reelect primary before upgrading service.")
+            # todo is it necessary to do something with status which forces the user to run
+            # "force-upgrade"
```
MiaAltieri: @carlcsaposs-canonical - question here 🙏

src/upgrades/machine_upgrade.py
```diff
@@ -145,21 +145,47 @@ def authorized(self) -> bool:
...
+            return
+
+        charm.install_snap_packages(packages=Config.SNAP_PACKAGES)
+
+        # todo question - do we set these on failed upgrades as well?
```
MiaAltieri: @carlcsaposs-canonical - question here 🙏

src/upgrades/machine_upgrade.py
```diff
@@ -145,21 +145,47 @@ def authorized(self) -> bool:
...
         logger.debug(f"Saved {_SNAP_REVISION} in unit databag after upgrade")
 
+        try:
+            logger.debug(
+                "Running post upgrade checks to verify cluster is not broken after upgrade"
+            )
+            charm.upgrade.post_upgrade_check()
+        except mongodb_upgrade.ClusterNotHealthyError:
+            # todo is it necessary to do something with status which forces the user to run
+            # "force-upgrade"
```
MiaAltieri: @carlcsaposs-canonical - question here 🙏

src/upgrades/machine_upgrade.py
carlcsaposs-canonical: To force user to run `force-upgrade`, all we need to do is not set unit state as "healthy"

FYI, if the snap revision is not updated, I believe upgrade_unit() will be called over and over again (since normally upgrade_unit would refresh the snap, then something would happen where we don't set healthy)

if that (upgrade_unit() getting called multiple times) is an issue, please lmk & can share how opensearch charm solved similar issue

src/upgrades/machine_upgrade.py
carlcsaposs-canonical: yes, as soon as the snap is refreshed we should set this

next unit won't proceed until we set unit state healthy

src/upgrades/machine_upgrade.py
carlcsaposs-canonical: these checks should run before we set unit state to healthy

also, if these checks should be re-tried after failure (opensearch does something like this, where it checks cluster health after start & will re-check it in subsequent events if it failed), then this needs to be moved

since after the snap is refreshed upgrade_unit() won't be called again

src/upgrades/machine_upgrade.py
MiaAltieri: Thank you so much! I really appreciate it! In this case, its actually benefical!

src/upgrades/machine_upgrade.py
MiaAltieri: thank you so much!

src/upgrades/machine_upgrade.py
MiaAltieri: Thank you!

lib/charms/mongodb/v0/mongodb_tls.py
```diff
@@ -81,6 +81,11 @@ def _on_set_tls_private_key(self, event: ActionEvent) -> None:
             event.fail("Mongos cannot set TLS keys until integrated to config-server.")
             return
 
+        if self.charm.upgrade_in_progress:
+            logger.warning("Setting TLS key during an ugprade is not supported.")
+            event.fail("Setting TLS key during an ugprade is not supported.")
```
carlcsaposs-canonical:
```suggestion
            logger.warning("Setting TLS key during an upgrade is not supported.")
            event.fail("Setting TLS key during an upgrade is not supported.")
```

lib/charms/mongodb/v1/mongodb_backups.py
```diff
@@ -157,6 +171,11 @@ def _on_s3_credential_changed(self, event: CredentialsChangedEvent):
 
     def _on_create_backup_action(self, event) -> None:
         action = "backup"
+        if self.charm.upgrade_in_progress:
+            logger.warning("Creating a backup is not supported during an upgrade.")
+            event.fail("Creating a backup is not supported during an upgrade.")
```
carlcsaposs-canonical: IMO, the log doesn't need to be warning level since the event failed & the isn't risk of the charm breaking

lib/charms/mongodb/v1/mongodb_backups.py
```diff
@@ -236,6 +255,11 @@ def _on_list_backups_action(self, event) -> None:
 
     def _on_restore_action(self, event) -> None:
         action = "restore"
+        if self.charm.upgrade_in_progress:
+            logger.warning("Restoring a backup is not supported during an upgrade.")
+            event.fail("Restoring a backup is not supported during an upgrade.")
```
carlcsaposs-canonical: same as last comment about log doesn't need to be warning level

lib/charms/mongodb/v0/config_server_interface.py
```diff
@@ -85,6 +85,13 @@ def pass_hook_checks(self, event: EventBase) -> bool:
         if not self.charm.unit.is_leader():
             return False
 
+        if self.charm.upgrade_in_progress:
+            logger.warning(
+                "Adding mongos applications is not supported during an upgrade. The charm may be in a broken, unrecoverable state."
+            )
+            event.defer()
```
carlcsaposs-canonical: question: will this only happen when a mongos application is added? is looks like this function is called in ClusterProvider._on_relation_broken

lib/charms/mongodb/v1/shards_interface.py
```diff
@@ -691,8 +711,36 @@ def _on_relation_changed(self, event):
...
+    def pass_hook_checks(self, event: EventBase):
         """Runs the pre-hooks checks for ConfigServerRequirer, returns True if all pass."""
+        proceed_complex_hook_checks = self.pass_sanity_hook_checks(event)
+        if not proceed_complex_hook_checks:
+            return False
+
+        # occasionally, broken events have no application, in these scenarios nothing should be
+        # processed.
+        if not event.relation.app and isinstance(event, RelationBrokenEvent):
+            return False
```
carlcsaposs-canonical, line 721: nit:
is this for deferred relationbroken events? if not, surprised to see this

> in these scenarios nothing should be processed

for your charm, is this acceptable? if you're getting relation broken, there's a relation with an application that's going away

during relation broken you can rely on your local app data being available, but not remote app data—since there's a race with the other app getting relation broken

but I'm imagining whatever cleanup you're running using the remote app maybe should still run if the remote app is not available?

not sure, just wanted to make you aware of this if you weren't already

lib/charms/mongodb/v1/shards_interface.py
```diff
@@ -691,8 +711,36 @@ def _on_relation_changed(self, event):
...
+
+        mongos_hosts = event.relation.data[event.relation.app].get(HOSTS_KEY, None)
```
carlcsaposs-canonical:
```suggestion
        mongos_hosts = event.relation.data[event.relation.app].get(HOSTS_KEY)
```
nit: .get default value defaults to `None`

lib/charms/mongodb/v1/shards_interface.py
```diff
@@ -691,8 +711,36 @@ def _on_relation_changed(self, event):
...
+            return False
+
+        if self.charm.upgrade_in_progress:
+            logger.warning(
+                "Adding/Removing shards is not supported during an upgrade. The charm may be in a broken, unrecoverable state"
+            )
+            if not isinstance(event, RelationBrokenEvent):
+                # upgrades should not block relation broken events
+                event.defer()
+            return False
```
carlcsaposs-canonical: question: how does the presence/lack of event.defer() after if the relation broken event is blocked? given that we're returing False in both cases

src/charm.py
```diff
@@ -855,6 +898,11 @@ def pass_pre_set_password_checks(self, event: ActionEvent) -> bool:
             event.fail("The action can be run only on leader unit.")
             return
 
+        if self.upgrade_in_progress:
+            logger.warning("Do not set the password while a backup/restore is in progress.")
+            event.fail("Cannot set passwords while an upgrade is in progress.")
```
carlcsaposs-canonical: same as earlier comment about warning log not necessary

src/upgrades/machine_upgrade.py
```diff
@@ -145,21 +145,36 @@ def authorized(self) -> bool:
...
         return False
 
-    def upgrade_unit(self) -> None:
+    def upgrade_unit(self, *, charm) -> None:
         """Runs the upgrade procedure.
 
         Only applies to machine charm.
         """
         logger.debug(f"Upgrading {self.authorized=}")
         self.unit_state = upgrade.UnitState.UPGRADING
```
carlcsaposs-canonical: nit: I would put this log & setter right before the snap is refreshed (for consistency with other charms)

may be worth adding another (different) debug log here in its place

src/upgrades/machine_upgrade.py
```diff
@@ -145,21 +145,36 @@ def authorized(self) -> bool:
...
+        # According to the MongoDB documentation, before upgrading the primary, we must ensure a
+        # safe primary re-election.
+        try:
+            if self._unit.name == charm.primary:
+                logger.debug("Stepping down current primary, before upgrading service...")
+                charm.upgrade.step_down_primary_and_wait_reelection()
+        except mongodb_upgrade.FailedToElectNewPrimaryError:
+            # by not setting the snap revision and immediately returning, this function will be
+            # called again, and an empty re-elect a primary will occur again.
+            logger.error("Failed to reelect primary before upgrading service.")
```
carlcsaposs-canonical: nit: "upgrading service" seems a bit inaccurate to me (when I think of "service", my thought is the snap service, but it's the snap/workload/mongodb that's being upgraded)

src/upgrades/mongodb_upgrade.py
```diff
@@ -112,10 +138,36 @@ def _on_force_upgrade_action(self, event: ActionEvent) -> None:
...
+        self._upgrade.upgrade_unit(charm=self.charm)
         event.set_results({"result": f"Forcefully upgraded {self.unit.name}"})
         logger.debug("Forced upgrade")
 
+    def post_upgrade_check(self, event: EventBase):
+        """Runs the post upgrade check to verify that the cluster is healthy.
+
+        By deferring before setting unit state to HEALTHY, the user will either:
+            1. have to wait for the unit to resolve itself.
+            2. have to run the force-upgrade action.
```
carlcsaposs-canonical:
```suggestion
            2. have to run the force-upgrade action (to upgrade the next unit).
```
nit: want to clarify that the user doesn't run force-upgrade on that unit not set to HEALTHY

src/upgrades/mongodb_upgrade.py
```diff
@@ -112,10 +138,36 @@ def _on_force_upgrade_action(self, event: ActionEvent) -> None:
...
+        """
+        logger.debug("Running post upgrade checks to verify cluster is not broken after upgrade")
+
+        if not self.is_cluster_healthy():
+            logger.error(
+                "Cluster is not healthy after upgrading unit %s, nodes are still syncing. Deferring post upgrade check.",
+                self.charm.unit.name,
+            )
```
carlcsaposs-canonical: consider logging rollback instructions

also consider setting unit status (for example, something similar to what we set if automatic pre-check fails
https://github.com/canonical/opensearch-operator/pull/262/files#diff-b0d4c16f2b702229d288067c4933254e190e3c0ac6696e337e604b33f970f47dR46-R48
https://github.com/canonical/opensearch-operator/pull/262/files#diff-b9ed39bbc9c0387bd3e07da31d13373745534a1cd723d3e292c73496b12e307cR106
)

src/upgrades/mongodb_upgrade.py
```diff
@@ -112,10 +138,36 @@ def _on_force_upgrade_action(self, event: ActionEvent) -> None:
...
+
+        By deferring before setting unit state to HEALTHY, the user will either:
+            1. have to wait for the unit to resolve itself.
+            2. have to run the force-upgrade action.
+        """
+        logger.debug("Running post upgrade checks to verify cluster is not broken after upgrade")
+
+        if not self.is_cluster_healthy():
+            logger.error(
+                "Cluster is not healthy after upgrading unit %s, nodes are still syncing. Deferring post upgrade check.",
```
carlcsaposs-canonical: nit: "deferring post upgrade check" may be unclear to user

something like "will retry next juju event" may be more clear

src/upgrades/mongodb_upgrade.py
```diff
@@ -131,6 +183,146 @@ def _set_upgrade_status(self):
         if isinstance(self.charm.unit.status, ActiveStatus):
             self.charm.unit.status = self._upgrade.get_unit_juju_status() or ActiveStatus()
 
+    def is_cluster_healthy(self) -> bool:
+        """Returns True if all nodes in the cluster/replcia set are healthy."""
+        if self.charm.is_role(Config.Role.SHARD):
+            logger.debug("Cannot run full cluster health check on shards")
+            return False
```
carlcsaposs-canonical: question: so are shards currently not upgradable by design? (i.e. will never set HEALTHY)

src/upgrades/mongodb_upgrade.py
```diff
@@ -131,6 +183,146 @@ def _set_upgrade_status(self):
...
+
+        return self.are_nodes_healthy()
+
+    def are_nodes_healthy(self) -> bool:
+        """Returns True if all nodes in the MongoDB deployment are healthy."""
+        try:
+            if self.charm.is_role(Config.Role.CONFIG_SERVER):
+                # TODO Future PR - implement node healthy check for sharded cluster
+                pass
```
carlcsaposs-canonical: nit: maybe return False so that function always returns bool instead of sometimes returning None

src/upgrades/mongodb_upgrade.py
```diff
@@ -131,6 +183,146 @@ def _set_upgrade_status(self):
...
+        secondary_config.hosts = {host}
+        with MongoDBConnection(secondary_config, direct=True) as direct_seconary:
+            db = direct_seconary.client[db_name]
+            test_collection = db[collection]
+            query = test_collection.find({}, {WRITE_KEY: 1})
+            if query[0][WRITE_KEY] != expected_write_value:
+                raise ClusterNotHealthyError
+
+    def get_random_write_and_collection(self) -> Tuple[str, str]:
+        """Returns a tutple for a random collection name and a unique write to add to it."""
```
carlcsaposs-canonical:
```suggestion
        """Returns a tuple for a random collection name and a unique write to add to it."""
```

src/upgrades/mongodb_upgrade.py
```diff
@@ -131,6 +183,146 @@ def _set_upgrade_status(self):
...
+
+        old_primary = self.charm.primary
+        with MongoDBConnection(self.charm.mongodb_config) as mongod:
+            mongod.step_down_primary()
+
+        for attempt in Retrying(stop=stop_after_attempt(30), wait=wait_fixed(1), reraise=True):
+            with attempt:
+                new_primary = self.charm.primary
+                if new_primary == old_primary:
+                    raise FailedToElectNewPrimaryError()
```
carlcsaposs-canonical: FYI, did not review this section since not too familiar with mongo

src/upgrades/upgrade.py
```diff
@@ -226,7 +226,7 @@ def authorized(self) -> bool:
         """
 
     @abc.abstractmethod
-    def upgrade_unit(self, *, snap) -> None:
+    def upgrade_unit(self, charm) -> None:
```
carlcsaposs-canonical:
```suggestion
    def upgrade_unit(self, *, charm) -> None:
```
nit: to keep some API consistency with other charms

carlcsaposs-canonical (approved):
fyi, did not review test code

src/config.py
```diff
@@ -12,6 +12,7 @@
 class Config:
     """Configuration for MongoDB Charm."""
 
+    HEALTHY = "healthy"
```
carlcsaposs-canonical:
```suggestion
```
where is this used? the upgrade shared code provides an enum already

carlcsaposs-canonical (changes requested):

lib/charms/mongodb/v0/config_server_interface.py
MiaAltieri: ah good point - let me update this to say "The processing of ..."

lib/charms/mongodb/v1/mongodb_backups.py
MiaAltieri: ah good info, I will change it to debug :)

src/config.py
MiaAltieri: you're right! can be removed :)

carlcsaposs-canonical (approved):

lib/charms/mongodb/v1/shards_interface.py
MiaAltieri: TBH it has been a long time since I have written this code and I am not too familiar with this use case, but I recall this check being necessary.

Thanks a ton for this context and info, it is super helpful

This code was just moved around in this PR. Perhaps we could discuss this at a later date to dig into why this issue was occurring, but not in this PR since it's related to sharding and not upgrades.

lib/charms/mongodb/v1/shards_interface.py
MiaAltieri: good point, relation broken should proceed and not return false, let me update this!

src/upgrades/mongodb_upgrade.py
MiaAltieri: They are, but the post upgrade check is more involved, let me update this log and add a TODO

src/upgrades/mongodb_upgrade.py
MiaAltieri: great point!

lib/charms/mongodb/v0/config_server_interface.py
```diff
@@ -130,6 +137,11 @@ def _on_relation_changed(self, event) -> None:
         self.database_provides.update_relation_data(event.relation.id, relation_data)
 
     def _on_relation_broken(self, event) -> None:
+        if self.charm.upgrade_in_progress:
+            logger.warning(
+                "disabling TLS is not supported during an upgrade. The charm may be in a broken, unrecoverable state."
```
Mehdi-Bendriss: Wrong message?

lib/charms/mongodb/v1/shards_interface.py
```diff
@@ -165,6 +165,14 @@ def pass_hook_checks(self, event: EventBase) -> bool:
         if not self.charm.unit.is_leader():
             return False
 
+        return True
+
+    def pass_hook_checks(self, event: EventBase) -> bool:
+        """Runs the pre-hooks checks for ShardingProvider, returns True if all pass."""
+        proceed_complex_hook_checks = self.pass_sanity_hook_checks(event)
+        if not proceed_complex_hook_checks:
```
Mehdi-Bendriss: nit: 
```suggestion
        if not self.pass_sanity_hook_checks(event):
```

lib/charms/mongodb/v1/mongodb_backups.py
```diff
@@ -138,6 +145,13 @@ def _on_s3_credential_changed(self, event: CredentialsChangedEvent):
         # handling PBM configurations requires that MongoDB is running and the pbm snap is
         # installed.
         action = "configure-pbm"
+        if self.charm.upgrade_in_progress:
+            logger.warning(
+                "Adding s3-relations is not supported during an upgrade. The charm may be in a broken, unrecoverable state."
```
Mehdi-Bendriss: Wrong message

lib/charms/mongodb/v1/shards_interface.py
```diff
@@ -691,8 +711,36 @@ def _on_relation_changed(self, event):
 
         self.charm.app_peer_data["mongos_hosts"] = json.dumps(self.get_mongos_hosts())
 
-    def pass_hook_checks(self, event):
+    def pass_hook_checks(self, event: EventBase):
         """Runs the pre-hooks checks for ConfigServerRequirer, returns True if all pass."""
+        proceed_complex_hook_checks = self.pass_sanity_hook_checks(event)
+        if not proceed_complex_hook_checks:
```
Mehdi-Bendriss: nit:
```suggestion
        if not self.pass_sanity_hook_checks(event):
```

lib/charms/mongodb/v1/shards_interface.py
```diff
@@ -691,8 +711,36 @@ def _on_relation_changed(self, event):
...
+
+        mongos_hosts = event.relation.data[event.relation.app].get(HOSTS_KEY)
+
+        if isinstance(event, RelationBrokenEvent) and not mongos_hosts:
+            logger.info("Config-server relation never set up, no need to process broken event.")
+            return False
+
+        if self.charm.upgrade_in_progress:
+            logger.warning(
+                "Adding/Removing shards is not supported during an upgrade. The charm may be in a broken, unrecoverable state"
```
Mehdi-Bendriss, line 731: I don't think this message is entirely accurate. This does not take into consideration updates to the `config-server` rel data (credentials, mongos hosts etc...)

lib/charms/mongodb/v1/shards_interface.py
```diff
@@ -182,6 +196,12 @@ def pass_hook_checks(self, event: EventBase) -> bool:
 
             if not self.charm.proceed_on_broken_event(event):
                 return False
+        elif self.charm.upgrade_in_progress:
+            logger.warning(
+                "Adding/Removing shards is not supported during an upgrade. The charm may be in a broken, unrecoverable state"
```
Mehdi-Bendriss: On the case of rel-changed event, Is this guaranteed to only happen during a shard add/removal? or even in the case of `shards` rel data update? in which case this message may need to be adjusted

src/charm.py
```diff
@@ -452,6 +474,13 @@ def _on_relation_handler(self, event: RelationEvent) -> None:
         Args:
             event: The triggering relation joined/changed event.
         """
+        if self.upgrade_in_progress:
+            logger.warning(
+                "Adding/Removing replicas during an upgrade is not supported. The charm may be in a broken, unrecoverable state"
```
Mehdi-Bendriss: May need to adapt the message for `peer-rel-changed` data update triggered event

src/charm.py
```diff
@@ -855,6 +898,11 @@ def pass_pre_set_password_checks(self, event: ActionEvent) -> bool:
             event.fail("The action can be run only on leader unit.")
             return
 
+        if self.upgrade_in_progress:
+            logger.debug("Do not set the password while a backup/restore is in progress.")
+            event.fail("Cannot set passwords while an upgrade is in progress.")
```
Mehdi-Bendriss, line 903: Missing `return False` maybe?

src/upgrades/mongodb_upgrade.py
```diff
@@ -112,10 +140,45 @@ def _on_force_upgrade_action(self, event: ActionEvent) -> None:
             return
         logger.debug("Forcing upgrade")
         event.log(f"Forcefully upgrading {self.unit.name}")
```
Mehdi-Bendriss:
```suggestion
        event.log(f"Forcefully upgrading {self.charm.unit.name}")
```

In addition to this one, a few lines above, `self.app.name` should be changed to `self.charm.app.name`

src/upgrades/mongodb_upgrade.py
```diff
@@ -112,10 +140,45 @@ def _on_force_upgrade_action(self, event: ActionEvent) -> None:
...
-        self._upgrade.upgrade_unit()
+        self._upgrade.upgrade_unit(charm=self.charm)
         event.set_results({"result": f"Forcefully upgraded {self.unit.name}"})
```
Mehdi-Bendriss:
```suggestion
        event.set_results({"result": f"Forcefully upgraded {self.charm.unit.name}"})
```

src/upgrades/mongodb_upgrade.py
```diff
@@ -131,6 +194,147 @@ def _set_upgrade_status(self):
...
+    def is_cluster_healthy(self) -> bool:
+        """Returns True if all nodes in the cluster/replcia set are healthy."""
+        if self.charm.is_role(Config.Role.SHARD):
+            logger.debug("Cannot run full cluster health check on shards")
+            # TODO Future PR - implement cgecj healthy check for single shard
+            return False
+
+        return self.are_nodes_healthy()
+
+    def are_nodes_healthy(self) -> bool:
```
Mehdi-Bendriss: This method doesn't seem needed, I think we can have all this logic within `is_cluster_healthy`

src/upgrades/mongodb_upgrade.py
```diff
@@ -131,6 +194,147 @@ def _set_upgrade_status(self):
...
+        """Returns True if all nodes in the MongoDB deployment are healthy."""
+        try:
+            if self.charm.is_role(Config.Role.CONFIG_SERVER):
+                # TODO Future PR - implement node healthy check for sharded cluster
+                return False
+            if self.charm.is_role(Config.Role.REPLICATION):
+                return self.are_replica_set_nodes_healthy(self.charm.mongodb_config)
+        except (PyMongoError, OperationFailure, ServerSelectionTimeoutError) as e:
```
Mehdi-Bendriss, line 210: I would prefer that this error handling moves to `are_replica_set_nodes_healthy` to simplify this method

src/upgrades/mongodb_upgrade.py
```diff
@@ -131,6 +194,147 @@ def _set_upgrade_status(self):
...
+        """Returns true if all nodes in the MongoDB replica set are healthy."""
+        with MongoDBConnection(mongodb_config) as mongod:
+            rs_status = mongod.get_replset_status()
+            rs_status = mongod.client.admin.command("replSetGetStatus")
+            return not mongod.is_any_sync(rs_status)
+
+    def is_cluster_able_to_read_write(self) -> bool:
+        """Returns True if read and write is feasible for cluster."""
+        if self.charm.is_role(Config.Role.SHARD):
+            logger.debug("Cannot run read/write check on shard, must run via config-server.")
```
Mehdi-Bendriss, line 226: How to know that a specific shard is able to accept read/writes?

src/upgrades/mongodb_upgrade.py
```diff
@@ -131,6 +194,147 @@ def _set_upgrade_status(self):
...
+            primary_ip = mongod.primary()
+
+        replica_ips = mongodb_config.hosts
+        secondary_ips = replica_ips - set(primary_ip)
+        for secondary_ip in secondary_ips:
+            try:
+                self.confirm_excepted_write_on_replica(
+                    secondary_ip, db_name, collection_name, expected_write_value, mongodb_config
+                )
+            except RetryError:
```
Mehdi-Bendriss: I believe this should be `except ClusterNotHealthyError` - given you're using `reraise=True`

Mehdi-Bendriss (commented):
Thanks Mia - I have a few comments / questions

lib/charms/mongodb/v0/config_server_interface.py
MiaAltieri: 🤦‍♀️ yes thanks for catching that

lib/charms/mongodb/v1/shards_interface.py
MiaAltieri: Ooh yeah - good point, let me update it to be more accurate!

src/upgrades/mongodb_upgrade.py
MiaAltieri: I would prefer to keep it here. Once we implement the code about for checking sharded cluster node health, we will need to do the very same error handling. I 

I would rather handle these errors once than duplicated in their respective functions

src/upgrades/mongodb_upgrade.py
MiaAltieri: Easy! We already [do something similar in our integration tests](https://github.com/canonical/mongodb-operator/blob/d67c8152a2f44f7f0e34759c9f47fa6dbcff804a/tests/integration/sharding_tests/test_sharding_backups.py#L41)

More specifically we would do:
1. create a DB 
2. set that DB to use a specific shard (mongoDB function called `movePrimary` - confusing name I know as it refers to Primary Shard **not** primary replica )
3. write to that DB
4. check write was successful on primary replica of that shard
5. check write was successful on secondary replicas of that shard

src/upgrades/mongodb_upgrade.py
MiaAltieri: good catch, thank you!

Mehdi-Bendriss (approved):

PR #417, opened 2024-06-04
[DPE-4156] pre upgrade hook pattern

MiaAltieri:


src/upgrades/upgrade.py
```diff
@@ -231,3 +249,46 @@ def upgrade_unit(self, *, charm) -> None:
...
+        On machines, this runs before any units are upgraded (after `juju refresh`)
+        On machines & Kubernetes, this also runs during pre-upgrade-check action
+        Can run on leader or non-leader unit
+        Raises:
+            PrecheckFailed: App is not ready to upgrade
+
+        TODO Kubernetes: Run (some) checks after `juju refresh` (in case user forgets to run
+        pre-upgrade-check action). Note: 1 unit will upgrade before we can run checks (checks may
+        need to be modified).
+        See https://chat.canonical.com/canonical/pl/cmf6uhm1rp8b7k8gkjkdsj4mya
```
carlcsaposs-canonical:
```suggestion
        Runs before any units are upgraded

        Does *not* run during rollback

        On machines, this runs before any units are upgraded (after `juju refresh`)
        On machines & Kubernetes, this also runs during pre-upgrade-check action

        Can run on leader or non-leader unit

        Raises:
            PrecheckFailed: App is not ready to upgrade

        TODO Kubernetes: Run (some) checks after `juju refresh` (in case user forgets to run
        pre-upgrade-check action). Note: 1 unit will upgrade before we can run checks (checks may
        need to be modified).
        See https://chat.canonical.com/canonical/pl/cmf6uhm1rp8b7k8gkjkdsj4mya
```
nit: sync with docstring in opensearch

src/upgrades/upgrade.py
```diff
@@ -231,3 +249,46 @@ def upgrade_unit(self, *, charm) -> None:
...
         Only applies to machine charm
         """
+
+    def pre_upgrade_check(self) -> None:
+        """Check if this app is ready to upgrade.
+
+        Runs before any units are upgraded
+        Does *not* run during rollback
+        On machines, this runs before any units are upgraded (after `juju refresh`)
+        On machines & Kubernetes, this also runs during pre-upgrade-check action
```
carlcsaposs-canonical, line 266: please lmk if you implement upgrades on k8s—we need to extend how the pre-upgrade-check works for k8s in the shared code

src/upgrades/upgrade.py
```diff
@@ -231,3 +249,46 @@ def upgrade_unit(self, *, charm) -> None:
...
+        pre-upgrade-check action). Note: 1 unit will upgrade before we can run checks (checks may
+        need to be modified).
+        See https://chat.canonical.com/canonical/pl/cmf6uhm1rp8b7k8gkjkdsj4mya
+        """
+        logger.debug("Running pre-upgrade checks")
+
+        if self._charm.is_role(SHARD):
+            raise PrecheckFailed(
+                "Cannot run pre-upgrade check on shards, run this action on the related config-server."
+            )
```
carlcsaposs-canonical: after `juju refresh`, pre-upgrade-check will automatically run on the first unit to upgrade

is this intended to fail in automatic pre-upgrade check?

see docstring—pre-upgrade-checks run both during the `pre-upgrade-check` action, and automatically after refresh (on VM)

src/upgrades/upgrade.py
```diff
@@ -231,3 +249,46 @@ def upgrade_unit(self, *, charm) -> None:
...
+
+        if not self._charm.upgrade.is_cluster_healthy():
+            raise PrecheckFailed(
+                "Cluster is not healthy, do not proceed with ugprade. Please check juju debug for information."
+            )
```
carlcsaposs-canonical: UX: please see the pre-check failed exception & how it uses the message here when constructing a Juju status (if pre-upgrade-check is run automatically after refresh) or when constructing logs in the action output (if pre-upgrade-check is run via action)

note that juju status messages are truncated after 120 characters, so please keep this message short (given that we're already adding several characters in the PrecheckFailed exception). Also, the call to action is already to rollback, would suggest not adding another call to action about checking the debug log

src/upgrades/upgrade.py
```diff
@@ -231,3 +249,46 @@ def upgrade_unit(self, *, charm) -> None:
...
+
+        # We do not get to decide the order of units to upgrade, so we move the primary to the
+        # last unit to upgrade. This prevents the primary from jumping around from unit to unit
+        # during the upgrade procedure.
+        try:
+            self._charm.upgrade.move_primary_to_last_upgrade_unit()
+        except FailedToMovePrimaryError:
+            raise PrecheckFailed(
+                "Cluster failed to move primary before re-election. do not proceed with ugprade."
+            )
```
carlcsaposs-canonical: same UX nit about call to action & length of message

src/upgrades/upgrade.py
```diff
@@ -231,3 +249,46 @@ def upgrade_unit(self, *, charm) -> None:
...
+
+        if not self._charm.upgrade.is_cluster_able_to_read_write():
+            raise PrecheckFailed(
+                "Cluster is not healthy cannot read/write to replicas, do not proceed with ugprade. Please check juju debug for information."
+            )
```
carlcsaposs-canonical: same UX nit about call to action & length of message

src/upgrades/mongodb_upgrade.py
```diff
@@ -182,6 +218,34 @@ def post_upgrade_check(self, event: EventBase):
...
+        """Returns the unit that will be the last to be upgraded (ie the one with the lowest id."""
+        unit_with_lowest_id = self.charm.unit
+        lowest_unit_id = unit_with_lowest_id.name.split("/")[1]
+        for unit in self.charm.peers.units:
+            unit_id = unit.name.split("/")[1]
+            if unit_id < lowest_unit_id:
+                unit_with_lowest_id = unit
+                lowest_unit_id = unit_id
+
+        return unit_with_lowest_id
```
carlcsaposs-canonical: nit: reuse https://github.com/canonical/mongodb-operator/blob/137878cfb7c93142f6cfc415714489a061e19ff8/src/upgrades/upgrade.py#L153 instead of re-implementing

carlcsaposs-canonical: e.g. `_sorted_units[-1]`

src/upgrades/mongodb_upgrade.py
```diff
@@ -191,14 +255,34 @@ def _set_upgrade_status(self):
...
+            if not direct_mongo.is_ready:
+                logger.debug("Cannot proceed with upgrade. Service mongod is not running")
+                return False
+
+        unit_state = self.charm.process_statuses()
+        if not isinstance(unit_state, ActiveStatus):
+            logger.debug(
+                "Cannot proceed with upgrade. Unit is not in Active state, in: %s.", unit_state
+            )
             return False
```
carlcsaposs-canonical: consider raising PrecheckFailed directly here instead of return False so that the PrecheckFailed error message can be more specific

src/upgrades/machine_upgrade.py
```diff
@@ -123,12 +123,28 @@ def authorized(self) -> bool:
...
             if unit.name == self._unit.name:
                 # Higher number units have already upgraded
-                if index == 1:
+                if index == 0:
+                    if (
+                        json.loads(self._app_databag["versions"])["charm"]
+                        == self._current_versions["charm"]
+                    ):
+                        # Assumes charm version uniquely identifies charm revision
+                        logger.debug("Rollback detected. Skipping pre-upgrade check")
```
carlcsaposs-canonical, line 141: is this assumption correct for mongo? it looks like you're missing the wrapper for charm version during builds that adds unique identifier

e.g.
https://github.com/canonical/mysql-router-k8s-operator/blob/583f9914f5df6e348479eccbb87c8498b4b3a987/tox.ini#L21-L44
https://github.com/canonical/opensearch-operator/blob/20816216f7fdb1b3ff8e706ac3dcdb84a0a2f133/tox.ini#L22-L45

src/charm.py
```diff
@@ -977,6 +977,12 @@ def _handle_reconfigure(self, event: UpdateStatusEvent):
         event.unit = self.unit
         self._on_relation_handler(event)
 
+        # make sure all nodes in the replica set have the same priority for re-election. This is
+        # necessary in the case that pre-upgrade hook fails to reset the priority of election for
+        # cluster nodes.
+        with MongoDBConnection(self.mongodb_config) as mongod:
+            mongod.set_replicaset_election_priority(priority=1)
```
carlcsaposs-canonical, line 986: can this function run while long-running upgrade in progress? if so, is that okay?

src/charm.py
```diff
@@ -903,6 +898,11 @@ def pass_pre_set_password_checks(self, event: ActionEvent) -> bool:
             event.fail("Cannot set passwords while an upgrade is in progress.")
             return False
 
+        if self.upgrade_in_progress:
+            logger.debug("Do not set the password while a backup/restore is in progress.")
+            event.fail("Cannot set passwords while an upgrade is in progress.")
+        return True
```
carlcsaposs-canonical: should this return False?

lib/charms/mongodb/v0/mongodb.py
```diff
@@ -290,6 +294,51 @@ def step_down_primary(self) -> None:
...
+                    self.step_down_primary()
+                    if self.primary() != new_primary:
+                        raise FailedToMovePrimaryError
+        except RetryError:
+            # catch all possible exceptions when failing to step down primary. We do this in order
+            # to ensure that we reset the replica set election priority.
+            is_move_successful = False
+
+        # reset all replicas to the same priority
+        self.set_replicaset_election_priority(priority=1)
```
carlcsaposs-canonical, line 323: I don't understand this code, so this is all speculation:

if you have network failure, could this function fail, raise an uncaught exception, causing a charm to re-execute the hook? if so, might be worth setting new_primary priority to 1 while setting priority to 0.5 for other members

tests/unit/test_upgrade.py
carlcsaposs-canonical, line 1: fyi, did not review this file

src/upgrades/upgrade.py
MiaAltieri: Will do. We will do this at some point during this 6mo cycle, but the goal is to use the upgrade implementation from the working group. :D

src/upgrades/mongodb_upgrade.py
MiaAltieri: oh heck yeah, thank u

src/charm.py
MiaAltieri: TL;DR Yes and Yes

Long Story:
The election priority is set + reset when the pre-upgrade-check is ran. It is used to "force" mongodb to move its primary to the lowest unit (Context: MongoDB does not let the user choose which unit the primary is on). If for some reason the charm fails to reset the election priority during the pre-upgrade-check hook and the user decides to move forward with `juju refresh` (despite the failing event) the charm will be configured to always have elections result in the lowest unit as a primary which does not follow HA-practices. It is necessary that we try to recorrect this and "self heal" this cluster config

lib/charms/mongodb/v0/mongodb.py
MiaAltieri: Sorry I don't follow 😅 

Would you mind explaining Why " if so, might be worth setting new_primary priority to 1 while setting priority to 0.5 for other members"? 

Maybe this information helps?
1. if the function fails and re-executes [the calling hook it won't trigger this call unless it is still needed to move the primary](https://github.com/canonical/mongodb-operator/blob/137878cfb7c93142f6cfc415714489a061e19ff8/src/upgrades/mongodb_upgrade.py#L240) 
2. It is dangerous to keep the priorities at 0.5, 0.5, 1 - since you're effectively "forcing" one unit to be primary and goes against HA-principles

src/upgrades/upgrade.py
MiaAltieri: Ah thank you. I have revisited it and added some new logic in.

src/upgrades/mongodb_upgrade.py
MiaAltieri: I could, but this function is called by the post upgrade check as well. I would rather keep this function multi-purpose if thats okay 🙏

src/upgrades/machine_upgrade.py
MiaAltieri: Ah great catch, thanks for that. I blindly copy pasta'd 🍝 It should always be ran even in a rollback. Would you mind checking that my fix here is correct?

src/charm.py
carlcsaposs-canonical: okay so, to confirm, if there's a primary switchover during an upgrade to a unit that's higher than the lowest unit number—that's acceptable (but you want to avoid it)?

as in, once we upgrade the primary, if we need to rollback that will be fine?

lib/charms/mongodb/v0/mongodb.py
carlcsaposs-canonical: > if the function fails and re-executes [the calling hook it won't trigger this call unless it is still needed to move the primary](https://github.com/canonical/mongodb-operator/blob/137878cfb7c93142f6cfc415714489a061e19ff8/src/upgrades/mongodb_upgrade.py#L240)

let's say that the primary still needs to be moved—then could you set a situation where all priorities are at 0.5? (because resetting them all back to 1 failed)

if not, please disregard—this is a very edge case (if it even exists) and it looks like you already have handling to reset to 1 in most cases if this fails

src/upgrades/mongodb_upgrade.py
carlcsaposs-canonical: you can catch precheckfailed during the post upgrade check

idk, was just a thought for more informative errors—but if that's not useful feel free to ignore

src/upgrades/machine_upgrade.py
carlcsaposs-canonical: the fix appears to be incorrect

this code should not be changed

the charm version is what should be changed

carlcsaposs-canonical: (we do not want to run pre-upgrade checks on rollback)

src/upgrades/upgrade.py
```diff
@@ -272,27 +272,22 @@ def pre_upgrade_check(self) -> None:
...
         # last unit to upgrade. This prevents the primary from jumping around from unit to unit
         # during the upgrade procedure.
         try:
             self._charm.upgrade.move_primary_to_last_upgrade_unit()
         except FailedToMovePrimaryError:
-            raise PrecheckFailed(
-                "Cluster failed to move primary before re-election. do not proceed with ugprade."
-            )
+            logger.error("Cluster failed to move primary before re-election.")
+            raise PrecheckFailed("Primary not on lowest unit")
```
carlcsaposs-canonical: nit: consider rewording this to something like "Primary switchover failed"—my concern is how it's currently worded might lead the user to think that they need to take action & switch the primary over themselves

lib/charms/mongodb/v0/mongodb.py
```diff
@@ -290,6 +294,51 @@ def step_down_primary(self) -> None:
         """Steps down the current primary, forcing a re-election."""
         self.client.admin.command("replSetStepDown", {"stepDownSecs": "60"})
 
+    def move_primary(self, new_primary: str) -> None:
```
Mehdi-Bendriss:
```suggestion
    def move_primary(self, new_primary_ip: str) -> None:
```

lib/charms/mongodb/v0/mongodb.py
```diff
@@ -290,6 +294,51 @@ def step_down_primary(self) -> None:
...
+        """Forcibly moves the primary to the new primary provided.
+
+        Args:
+            new_primary: str of the ip address which the new primary should belong.
```
Mehdi-Bendriss:
```suggestion
            new_primary: ip address of the unit chosen to be the new primary.
```

src/charm.py
MiaAltieri: > okay so, to confirm, if there's a primary switchover during an upgrade to a unit that's higher than the lowest unit number—that's acceptable (but you want to avoid it)?

Not quite, the goal of this piece of code is that we want to avoid that the `set_replicaset_election_priority ` is stuck in a way that "biases" one unit to become the primary. It is less related to upgrades and more just an HA/self-healing practice. This piece of code doesn't care about the success/failure of an upgrade. I hope that helps! :)

src/upgrades/machine_upgrade.py
MiaAltieri: Two things!
1.
I totally misread your first message about charm version , my apologies. Is it sufficient to update my tox file with that workflow you showed? Or is something else needed? 

2.
For MongoDB upgrades/downgrades we need to run the pre-upgrade check.

i.e. mongodb docs say that when downgrading there is a set of operations to follow (i.e. those in pre-upgrade check)  

Is there something can we do here? 😅

src/charm.py
carlcsaposs-canonical: so will things break if there's a primary switchover during an upgrade? (keeping in mind an upgrade can last for days or weeks)

src/upgrades/machine_upgrade.py
carlcsaposs-canonical: > Is it sufficient to update my tox file with that workflow you showed? Or is something else needed?

the shared code also needs to be reverted

> upgrades/downgrades we need to run the pre-upgrade check.

some definitions to explain the difference between rollback and downgrade:

> Workload: A software component that the charm operates (e.g. PostgreSQL)
> - Note: a charm can have 0, 1, or multiple workloads
> 
> Charm code: Contents of *.charm file or `charm` directory (e.g. `/var/lib/juju/agents/unit-postgresql-k8s-0/charm/`) on a unit. Contains charm source code and (specific versions of) Python dependencies
> 
> Outdated version: workload and/or charm code version on a unit that **does not** match the version of the application's current charm revision
> 
> Up-to-date version: workload and/or charm code version on a unit that **does** match the version of the application's current charm revision
> 
> Original version: workload and/or charm code version of all units immediately after the last completed upgrade—or, if no completed upgrades, immediately after `juju deploy` and (on machines) initial installation
> 
> ### For an application (or if not specified)
> Upgrade: `juju refresh` to a different workload and/or charm code version
> - Note: "rollback" and "downgrade" are specific types of "upgrade"
> 
> In-progress upgrade: 1+ units have an outdated workload and/or charm code version
> 
> Completed upgrade: All units have the up-to-date workload and charm code version
> 
> Rollback: While an upgrade is in-progress and 1+ units have the **original** workload and charm code version, `juju refresh` to the original workload and charm code version
>  - Note: If all units have already upgraded, then it would be a downgrade, not a rollback
>  - Note: If `juju refresh` is not to the original workload and charm code version, then it is not a rollback
> 
> Downgrade: Upgrade to older (lower) workload and/or charm code version

the pre-upgrade-check is skipped during rollback (because you're probably rolling back because something went wrong). It is not skipped for downgrades

src/charm.py
```diff
@@ -903,6 +898,11 @@ def pass_pre_set_password_checks(self, event: ActionEvent) -> bool:
             event.fail("Cannot set passwords while an upgrade is in progress.")
             return False
 
+        if self.upgrade_in_progress:
+            logger.debug("Do not set the password while a backup/restore is in progress.")
+            event.fail("Cannot set passwords while an upgrade is in progress.")
+            return False
```
Mehdi-Bendriss: this block seems to be a duplicate of the previous check

src/charm.py
```diff
@@ -958,12 +958,12 @@ def process_unremoved_units(self, event: LeaderElectedEvent) -> None:
...
 
-    def _handle_reconfigure(self, event: UpdateStatusEvent):
+    def perform_self_healing(self, event: UpdateStatusEvent):
         """Reconfigures the replica set if necessary.
 
-        Removes any mongod hosts that are no longer present in the replica set or adds hosts that
-        should exist in the replica set. This function is meant to be called periodically by the
-        leader in the update status hook to perform any necessary cluster healing.
+        Incidents such as network cuts can lead to new IP addresses and therefore will require a
+        reconfigure. Especially in the case that the leader a change in IP address it will not
```
Mehdi-Bendriss:
```suggestion
        reconfigure. Especially in the case that the leader's IP address changed, it will not
```

src/upgrades/mongodb_upgrade.py
```diff
@@ -109,6 +121,30 @@ def _on_upgrade_charm(self, _):
             # `upgrade_resumed`
             self._reconcile_upgrade()
 
+    def _on_pre_upgrade_check_action(self, event: ActionEvent) -> None:
+        if not self.charm.unit.is_leader():
+            message = f"Must run action on leader unit. (e.g. `juju run {self.app.name}/leader {upgrade.PRECHECK_ACTION_NAME}`)"
+            logger.debug(f"Pre-upgrade check event failed: {message}")
+            event.fail(message)
+            return
+        if not self._upgrade or self._upgrade.in_progress:
```
Mehdi-Bendriss, line 130: I think these 2 should have distinct messages. The first one effectively implies that the charm is not ready for an upgrade, correct?

src/upgrades/mongodb_upgrade.py
```diff
@@ -191,28 +245,60 @@ def _set_upgrade_status(self):
...
-            if self.charm.is_role(Config.Role.CONFIG_SERVER):
-                # TODO Future PR - implement node healthy check for sharded cluster
+            if self.charm.is_role(Config.Role.CONFIG_SERVER) or self.charm.is_role(
+                Config.Role.SHARD
+            ):
+                # TODO Future PR - implement node healthy check for entire cluster
                 return False
             if self.charm.is_role(Config.Role.REPLICATION):
-                return self.are_replica_set_nodes_healthy(self.charm.mongodb_config)
+                node_healthy = self.are_replica_set_nodes_healthy(self.charm.mongodb_config)
```
Mehdi-Bendriss: missing a return

src/upgrades/upgrade.py
```diff
@@ -231,3 +250,47 @@ def upgrade_unit(self, *, charm) -> None:
...
+        # TODO In future PR when we support upgrades on sharded clusters, have the shard verify
+        # that the config-server has already upgraded.
+
+        try:
+            self._charm.upgrade.wait_for_cluster_healthy()
+        except RetryError:
+            logger.error("Cluster is not healthy")
+            raise PrecheckFailed("Cluster is not healthy")
+
+        # We do not get to decide the order of units to upgrade, so we move the primary to the
```
Mehdi-Bendriss: I think we do get to decide the order of the units to upgrade, no?

Mehdi-Bendriss (commented):
Thanks Mia! I left a few comments / questions

src/charm.py
MiaAltieri: ah merge error, ty

src/upgrades/mongodb_upgrade.py
MiaAltieri: no, the return is performed below, but I think this can be cleaned up / reorganised. I'll do that

src/upgrades/mongodb_upgrade.py
carlcsaposs-canonical: the first case should never actually happen in the real world (it means the action was run while we're waiting for peer-relation-created), it's just to check that it's safe to check self._upgrade.in_progress (so anywhere we check `self._upgrade.in_progress`, we check for `self._upgrade` (not None) first)

src/upgrades/upgrade.py
MiaAltieri: No we dont. In V1 implementation we did, but in V2, it is highest to lowest

carlcsaposs-canonical: yes, as long as each unit can determine the order by themselves without using a databag and as long as the order does not change while the upgrade is in-progress

carlcsaposs-canonical: (on k8s, it's highest to lowest)

MiaAltieri: Wait I am confused. So @carlcsaposs-canonical we **can** choose the order?

MiaAltieri: If so, how can we do that. It would alleviate a lot of pain because then we could remove the "MovePrimary" logic

MiaAltieri: My apologise Carl, I had forgetten and of course you are correct. We **can** choose the order, but **only on VMs**.  I appreciate your patience

Back in April MongoDB team discussed the differences in K8s + VM implementations

And although we dislike the movePrimary functionality on VM (since we could choose the order of upgrades anyway) - we had decided that the benefits of keeping the charms insync out-weighed the benefits. So we decided that both charms should maintain the upgrade order of highest to lowest.

MiaAltieri: Let me update the comment to reflect this.

src/charm.py
MiaAltieri: Hmmm, definitely an interesting case you outline. The case being, before upgrading we move primary to the last unit to upgrade, but then the upgrade can last weeks and the primary moves to a different unit. TBH I am not sure here. 🤔 

On one hand: I believe that if a primary election happened, then there was a good reason for it. i.e. secondaries lost connection with the Primary. I don't think it makes sense for us to force a unit to be primary if internal mongodb "feels" otherwise. 

On the other hand: we want the primary to be the last upgrade, so we should try to guarantee that if possible. 

I am not sure if we can tell why MongoDB decided to elect a different replica as primary, which would be useful. I don't think we should "force" the charm to have a certain primary if the internal daemons think otherwise. The entire purpose of primary elections is to maintain HA, and if we override it we risk decreasing performance of the cluster. **But then again, we really want the primary to be the last unit to upgrade.** and maybe after enough time passes the cluster/replica set is "settled" and it can infact be moved back.

Personally, I think this is a complicated case to handle and I think it would be best to leave the code is. i.e. if the internal mongod daemon decides to move the primary to a different unit partway through the upgrade we should let it, since it probably has a good reason. 

But I would like to get someone else's opinion who might have more MongoDB knowledge (i.e. @delgod)

carlcsaposs-canonical: > we really want the primary to be the last unit to upgrade

want or need? wondering if it's a hard requirement or if the workload itself will be fine if primary is not last to upgrade

MiaAltieri: This is pending a technical conversation - we will revisit this topic later, if we decide to implement logic like this we will do so in a separate PR as it is regarding the upgrade procedure and not pre-upgrade-check

Thank you for the thought provoking topic!

src/upgrades/machine_upgrade.py
MiaAltieri: > the pre-upgrade-check is skipped during rollback (because you're probably rolling back because something went wrong). It is not skipped for downgrades

Ah okay, thank you. This makes me feel much better. Thank you Carl. Also thanks for this thorough description!

MiaAltieri: I believe I reverted the change. Is there something else that needs to be reverted?

carlcsaposs-canonical: think this diff still needs to be reverted https://github.com/canonical/mongodb-operator/pull/417/files/137878cfb7c93142f6cfc415714489a061e19ff8..55a5d25196c5fc12a31887f706613f6b9c16d6d1#diff-0817afc185cb1b0c81cf84d2af8c176418d24ef47bfb9609b5c7bd90bcd7c474

tox.ini
```diff
@@ -22,6 +22,29 @@ pass_env =
...
+    charmcraftcache
+    mv
+commands_pre =
+    # TODO charm versioning: Remove
+    # Workaround to add unique identifier (git hash) to charm version while specification
+    # DA053 - Charm versioning
+    # (https://docs.google.com/document/d/1Jv1jhWLl8ejK3iJn7Q3VbCIM9GIhp8926bgXpdtx-Sg/edit?pli=1)
+    # is pending review.
+    python -c 'import pathlib; import shutil; import subprocess; git_hash=subprocess.run(["git", "describe", "--always", "--dirty"], capture_output=True, check=True, encoding="utf-8").stdout; file = pathlib.Path("charm_version"); shutil.copy(file, pathlib.Path("charm_version.backup")); version = file.read_text().strip(); file.write_text(f"{version}+{git_hash}")'
+commands =
```
carlcsaposs-canonical: poetry export for requirements.txt is missing here—is this intentional?

if so, there's a mismatch with commands_post

MiaAltieri: This charm doesn't use poetry yet 😅 

Is this line needed even if it doesn't use poetry?

carlcsaposs-canonical: then `mv requirements.txt requirements-last-build.txt` should be removed or commented out
also, some things might break since this would be the first repo that has a pack wrapper without poetry

For example, the safeguards in `charmcraftcache` and that we include in charmcraft.yaml to ensure that pack doesn't accidentally happen via `charmcraft pack` instead of `tox run -e build-dev` or `tox run -e build-production` will break

would at least add a comment here in tox.ini, given that this is non-standard and non-supported use of the build wrapper

Mehdi-Bendriss (approved):

carlcsaposs-canonical (approved):

src/charm.py
delgod: > want or need? wondering if it's a hard requirement or if the workload itself will be fine if primary is not last to upgrade

I believe the next
e.g. upgrade from ver.X to ver.Y
1) We prefer, if possible, to have a minimal amount of switchovers during the upgrade.
2) It is expected if, because of failover, the primary ver.X is switched over to another ver.X member.
3) it is expected if, because of failover, the primary ver.X is switched over to ver.Y member **only when no ver.X members are left alive/available**.

In other words: HA should work.

MiaAltieri: According [to the mongodb docs](https://www.mongodb.com/docs/manual/core/replica-set-elections/), an election will occur for the following reasons:
- Adding a new node to the replica set **[We block this from happening during an upgrade]**
- [initiating a replica set,](https://www.mongodb.com/docs/manual/reference/method/rs.initiate/#mongodb-method-rs.initiate) **[This will not occur during an upgrade]**
- performing replica set maintenance using methods such as [rs.stepDown()](https://www.mongodb.com/docs/manual/reference/method/rs.stepDown/#mongodb-method-rs.stepDown) or [rs.reconfig()](https://www.mongodb.com/docs/manual/reference/method/rs.reconfig/#mongodb-method-rs.reconfig) - **[rs.stepDown() will not occur during an upgrade unless we willingly move primary + rs.reconfig() occurs only in the case of self-healing]**
- the [secondary](https://www.mongodb.com/docs/manual/reference/glossary/#std-term-secondary) members losing connectivity to the primary for more than the configured [timeout](https://www.mongodb.com/docs/manual/reference/replica-configuration/#mongodb-rsconf-rsconf.settings.electionTimeoutMillis) (10 seconds by default). **[this is a part of self healing and will occur during upgrades, but is expected for HA]**


So I believe it is fine to keep it as is.

carlcsaposs-canonical: @MiaAltieri does mongodb already take care of 2 (prefer primary switchover to units with older* version over units with newer version)?

*during downgrade, not sure if preference would be units on older version or on "non-upgraded" version

PR #418, opened 2024-06-11
[DPE-4490] New upgrade implementation

MiaAltieri:
## Issue
Current upgrade implementation should be replaced with v2

## Solution
Replace current upgrade implementation with v2


Mehdi-Bendriss (approved):
Thank you Mia - (Reviews done on merge to feature branch)

PR #419, opened 2024-06-11
[DPE-4494] Revision checking

MiaAltieri:
## Issue
Currently Charmed MongoDB supports creating clusters with different versions of charm components, i.e. a shard with revision 88 and a config server with revision 100. this can cause unexpected issues. 

## Solution
Prevent the charm from doing this by adding a revision check on relation joined and reporting inconsistencies in update status

## Future PRs
To keep the size of this PR down we will do integration and unit tests in the subsequent PRs 
If Pedro and other team member wish to use it we can create this as a lib in the data platform repo


src/charm.py
```diff
@@ -1535,12 +1551,48 @@ def is_relation_feasible(self, rel_interface) -> bool:
...
+        """Returns a Status if the cluster has mismatched revisions."""
+        this_charms_version = get_charm_revision(self.unit)
+        try:
+            # check for invalid versions in sharding integrations, i.e. a shard running on
+            # revision  88 and a config-server running on revision 110
+            if self.version_checker.are_related_apps_valid():
+                return
+        except NoVersionError as e:
+            # relations to shards/config-server are expected to provide a version number. If they
+            # do not, it is because they are from an earlier charm revision, i.e. pre-revison X.
```
phvalguima, line 1611: nit: "If they do not," can go in the same comment line

phvalguima (dismissed):
LGTM

There is one minor thing, but as we will touch this code later down the road, keep it as is

src/charm.py
MiaAltieri: they can't the line length goes to `105` which is greater than 99 :(

src/version_check.py
```diff
@@ -0,0 +1,231 @@
...
+            RelationInvalidError
+        """
+        if event.relation.name not in self.relations_to_check:
+            raise RelationInvalidError(
+                f"Provided relation: {event.relation.name} not in self.relations_to_check."
+            )
+
+        self.set_version_on_related_app(event.relation.name, event.app.name)
+
+    def is_local_charm(self, app_name: str) -> bool:
```
MiaAltieri, line 219: @Mehdi-Bendriss this function is not *yet* used in the charm, but we could use it. 

During a chat with Billy Olsen yesterday, he mentioned a potential scenario of the support team patching a hot fix to the charm. In this case the hotfix would result in the charm having a localhost revision. This would result in the relations showing either:

shard - Blocked(Charm revision 0 is not up-to date with config-server (Y))
config-server - Waiting(Waiting for all cluster components to upgrade to revision 0)

and the support teams fix not working. 

What we could do is if there is a revision mismatch and one of the charms is a local charm, we could not go into Blocked or Waiting, and instead log a `warning` with something like `WARNING: deploying a local charm, cannot check revision across components.` Should we do this?

Mehdi-Bendriss: I think that's a sensible approach

phvalguima: yes, seems reasonable

phvalguima (dismissed):

src/charm.py
```diff
@@ -136,6 +144,18 @@ def __init__(self, *args):
         self.legacy_client_relations = MongoDBLegacyProvider(self)
         self.tls = MongoDBTLS(self, Config.Relations.PEERS, substrate=Config.SUBSTRATE)
         self.backups = MongoDBBackups(self)
+
+        # TODO future PR - support pinning mongos version
+        self.version_checker = CrossAppVersionChecker(
```
Mehdi-Bendriss, line 149: Very nice - this should definitely make it into a lib and be adopted by other charms with cross-apps relations i.e OpenSearch and Postgres

src/charm.py
```diff
@@ -1478,6 +1498,9 @@ def get_invalid_integration_status(self) -> Optional[StatusBase]:
                 "Relation to s3-integrator is not supported, config role must be config-server"
             )
 
+        if self.get_cluster_mismatched_revision_status():
+            return self.get_cluster_mismatched_revision_status()
```
```suggestion
        return self.get_cluster_mismatched_revision_status()
```

src/charm.py
```diff
@@ -1558,12 +1581,56 @@ def is_relation_feasible(self, rel_interface) -> bool:
             )
             return False
```
Mehdi-Bendriss, line 1582: Side note: In general I would expect a method `is_relation_feasible` to only compute a boolean and not have a status setting as a side effect.

src/charm.py
```diff
@@ -1558,12 +1581,56 @@ def is_relation_feasible(self, rel_interface) -> bool:
...
 
+        if self.get_cluster_mismatched_revision_status():
+            self.unit.status = self.get_cluster_mismatched_revision_status()
+            return False
```
```suggestion
        if (revision_mismatch_status := self.get_cluster_mismatched_revision_status()):
            self.unit.status = revision_mismatch_status
            return False
```

src/charm.py
```diff
@@ -1558,12 +1581,56 @@ def is_relation_feasible(self, rel_interface) -> bool:
...
+            return
+
+        if self.version_checker.is_integrated_to_locally_built_charm():
+            logger.warning(INTEGRATED_TO_LOCALLY_BUIT_CHARM_WARNING)
+            return
+
+        try:
+            # check for invalid versions in sharding integrations, i.e. a shard running on
+            # revision  88 and a config-server running on revision 110
+            this_charms_version = get_charm_revision(self.unit)
```
Mehdi-Bendriss: I believe we can take it out of the try / except block + nit: renamed variable
```suggestion
        # check for invalid versions in sharding integrations, i.e. a shard running on
        # revision  88 and a config-server running on revision 110
        current_charm_revision = get_charm_revision(self.unit)
        try:
```

src/charm.py
```diff
@@ -1558,12 +1581,56 @@ def is_relation_feasible(self, rel_interface) -> bool:
...
+            if self.version_checker.are_related_apps_valid():
```
Mehdi-Bendriss, line 1607: I think here we should make a decision whether we return a boolean or raise an exception but not both. I think simply returning a boolean makes more sense here

src/version_check.py
```diff
@@ -0,0 +1,244 @@
...
+PREFIX_DIR = "/var/lib/juju/agents/"
+LOCAL_BUILT_CHARM_PREFIX = "local"
+
+
+def get_charm_revision(unit: Unit) -> int:
+    """Returns the charm revision.
+
+    TODO: Keep this until ops framework supports: https://github.com/canonical/operator/issues/1255
+    """
+    file_path = f"{PREFIX_DIR}unit-{unit.name.replace('/','-')}/charm/.juju-charm"
```
Mehdi-Bendriss, line 62: very nice find

src/version_check.py
```diff
@@ -0,0 +1,244 @@
...
+    revision = charm_path.split("-")[-1]
+    return int(revision)
+
+
+class CrossAppVersionChecker(Object):
+    """Verifies versions across multiple integrated applications."""
+
+    def __init__(
+        self,
+        charm,
```
Mehdi-Bendriss:
```suggestion
        charm: CharmBase,
```

src/version_check.py
```diff
@@ -0,0 +1,244 @@
...
+        version: int,
+        relations_to_check: List[str],
+        version_validity_range: Optional[Dict] = None,
+    ) -> None:
+        """Constructor for CrossAppVersionChecker object.
```
Mehdi-Bendriss:
```suggestion
        """Constructor for CrossAppVersionChecker.
```

src/version_check.py
```diff
@@ -0,0 +1,244 @@
...
+            charm: charm to inherit from
+            version: (int), the current version of the desired attribute of the charm
+            relations_to_check: (List), a list of relations who should have compatible versions
+                with the current charm
+            version_validity_range: (Optional Dict), a list of ranges for valid version ranges.
+                If not provided it is assumed that relations on the provided interface must have
+                the same version.
+        """
+        self.charm = charm
+        super().__init__(self.charm, None)
```
Mehdi-Bendriss: NIT: We can probably swap these, right? 
```suggestion
        super().__init__(charm, None)
        self.charm = charm
```

src/version_check.py
```diff
@@ -0,0 +1,244 @@
...
+        """
+        try:
+            invalid_relations = []
+            for relation_name in self.relations_to_check:
+                for relation in self.charm.model.relations[relation_name]:
+                    related_version = relation.data[relation.app][VERSION_CONST]
+                    if int(related_version) != self.version:
+                        invalid_relations.append((relation.app.name, related_version))
+        except KeyError:
+            raise NoVersionError(f"Expected {relation.app.name} to have version info.")
```
Mehdi-Bendriss, line 133: why not just add this relation to `invalid_relations` instead of throwing an Exception?

src/version_check.py
```diff
@@ -0,0 +1,244 @@
...
+            )
+
+        # this feature has yet to be implemented, MongoDB does not need it and it is unclear if
+        # this will be extended to other charms. If this code is extended to other charms and
+        # there is a valid usecase we will use the `version_validity_range` variable in the
+        # function `get_invalid_versions`
+        self.version_validity_range = version_validity_range
+
+    def get_invalid_versions(self) -> List[Tuple[str, int]]:
+        """Returns a tuple of (app name, version number), if the version number mismatches.
```
Mehdi-Bendriss:
```suggestion
        """Returns a list of (app name, version number) pairs, if the version number mismatches.
```

src/version_check.py
```diff
@@ -0,0 +1,244 @@
...
+        Mismatches are decided based on version_validity_range, if version_validity_range is not
+        provided, then the mismatch is expected to match this current app's version number.
+
+        Raises:
+            NoVersionError.
+        """
+        try:
+            invalid_relations = []
+            for relation_name in self.relations_to_check:
+                for relation in self.charm.model.relations[relation_name]:
```
Mehdi-Bendriss, line 128: Question: what would be the difference between iterating over the relations as done here, VS `self.charm.model.get_relation(relation_name)`?

src/version_check.py
```diff
@@ -0,0 +1,244 @@
...
+        """
+        try:
+            for relation_name in self.relations_to_check:
+                for rel in self.charm.model.relations[relation_name]:
+                    if rel.app.name == related_app_name:
+                        return int(rel.data[rel.app][VERSION_CONST])
+        except KeyError:
+            pass
+
+        raise NoVersionError(f"Expected {related_app_name} to have version info.")
```
Mehdi-Bendriss, line 151: This exception is not caught in `charm.py`. We could probably return None here instead of raising

src/version_check.py
```diff
@@ -0,0 +1,244 @@
...
+        with open(file_path) as f:
+            charm_path = f.read().rstrip()
+
+        return charm_path.split(":")[0]
+
+    def are_related_apps_valid(self) -> bool:
+        """Returns True if a related app has a version that's incompatible with the current app.
+
+        Raises:
+            NoVersionError.
```
Mehdi-Bendriss, line 169: See comment above about raising vs returning a bool

src/version_check.py
```diff
@@ -0,0 +1,244 @@
...
+                rel.data[self.charm.model.app][VERSION_CONST] = str(self.version)
+                rel.data[self.charm.model.app][DEPLOYMENT_TYPE] = str(self.get_deployment_prefix())
+
+    def set_version_on_related_app(self, relation_name: str, related_app_name: str) -> None:
+        """Sets the version number across for a specified relation on a specified app."""
+        if not self.charm.unit.is_leader():
+            return
+
+        relations = self.charm.model.relations[relation_name]
+        for rel in relations:
```
Mehdi-Bendriss: nit:
```suggestion
        for rel in self.charm.model.relations[relation_name]:
```

Mehdi-Bendriss (commented):
Thanks Mia! I have a few nits questions, mainly regarding the exceptions use / handling

src/charm.py
MiaAltieri: I'll be honest, I totally forgot one could do that. Thank you for the learnings!

src/version_check.py
MiaAltieri: Its all thanks to Pedro! <3

src/version_check.py
MiaAltieri: My reasoning is that relation X isn't necessarily invalid, it could be valid. But we don't know since we don't have any of its version info. I believe this is a separate case.

This can happen in the following scenarios:
1. Provider uses `self.version_checker = CrossAppVersionChecker( ... )` in src/charm.py, but Requirer does NOT. (Developer mistake)
2. The integration on is made to a much older version of a charm, in which the newer version *does* share the version info

src/version_check.py
MiaAltieri: ~~omg YES thank you Mehdi! `self.charm.model.get_relation(relation_name)` in an improvement~~
UPDATE this gives  
```
TypeError: 'Relation' object is not iterable
```

src/charm.py
MiaAltieri: I agree. This was something that was discussed previously (I think with Mykola?) when the PR was first made. Its certainly bad style. And also relevant to how our status handling could be improved. Lets please find time this 6mo cycle to update our status handling and resolve this 🙏

src/charm.py
MiaAltieri: Hmm, I actually disagree. IMO there are three scenarios:
1. relations are valid
2. relations are invalid
3. there is not enough information to determine validity of relations.

src/version_check.py
MiaAltieri: Two things:
> This exception is not caught in charm.py

This is on purpose. There is no need to check for this exception in `src/charm.py` at that point in time. This is because it would have caused an error on an earlier line. 

>  We could probably return None here instead of raising

I would prefer to keep it as is. I want to be consistent with the error of `NoVersionError` occurring when there is no version information present.

src/version_check.py
MiaAltieri: See response above

Mehdi-Bendriss (approved):

phvalguima (approved):

PR #421, opened 2024-06-17
[DPE-4494] add unit tests for revision checker

MiaAltieri:
# Issue
Code for version check has no unit tests. This is a requirement before publishing as a package

# Solution
Add unit tests for version check

### DIFF
--- a/tests/unit/test_version_check.py
+++ b/tests/unit/test_version_check.py
@@ -1,0 +1,93 @@
+# Copyright 2024 Canonical Ltd.
+# See LICENSE file for licensing details.
+import unittest
+from unittest.mock import patch
+
+from ops.testing import Harness
+
+from charm import MongodbOperatorCharm
+from version_check import DEPLOYMENT_TYPE, VERSION_CONST, NoVersionError
+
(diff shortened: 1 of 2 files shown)


Mehdi-Bendriss (approved):

tests/unit/test_version_check.py
```diff
@@ -0,0 +1,93 @@
...
+VALID_VERSION = CHARM_VERSION
+INVALID_VERSION = 456
+APP_0 = "APP_0"
+APP_1 = "APP_1"
+APP_2 = "APP_2"
+
+
+class TestCharm(unittest.TestCase):
+    @patch("charm.get_charm_revision")
+    @patch_network_get(private_address="1.1.1.1")
```
phvalguima, line 26: Do we need to patch network_get?

phvalguima (approved):

tests/unit/test_version_check.py
MiaAltieri: I'm not sure. In general most unit tests need it. But probably not needed here. I can remove it next time I revisit this file like we discussed

PR #422, opened 2024-06-18
[DPE-4665 ] Pre upgrade check sharding

MiaAltieri:
## Issue
Sharded deployments do not support preflight check

## Preflight checks added
**Both Config-Server and Shards check**
1. Ensure cluster is not running backup / tls actions [Config-Serve 
2. Health check - For shards and config servers, ensure that no node  is in the ROLLBACK or RECOVERING state.
3. Move Primary for Config Server to the unit with the lowest id 
4. make a dummy write to each shard (and read from each replica) to ensure the MongoDB cluster works.

**Config Server only:**
1. check feature compatibility version on all nodes
2. set Feature Compatibility Version for mongos 
3. Disable the balancer 
4. Set status waiting for refresh command 

## Preflight checks NOT added
1. Check juju unit state for all units via shared interface with the usage of “goal_state” 
2. Backup the config Database

1. Will be done in a future PR
2. Will not be performed, instead we will instruct users to perform a backup


src/config.py
```diff
@@ -107,3 +109,17 @@ class Secrets:
         SECRET_INTERNAL_LABEL = "internal-secret"
         SECRET_DELETED_LABEL = "None"
         MAX_PASSWORD_LENGTH = 4096
+
+    class Status:
+        """Status related constants.
+
+        TODO: move all status messages here.
```
MiaAltieri, line 114: Future work once we start to restructure how we handle statuses

lib/charms/mongodb/v1/mongos.py
```diff
@@ -358,12 +360,21 @@ def is_ready(self) -> bool:
 
         return True
 
+    def are_all_shards_aware(self) -> bool:
+        """Returns True if all shards are shard aware."""
+        sc_status = self.client.admin.command("listShards")
+        for shard in sc_status["shards"]:
+            if shard["state"] != SHARD_AWARE_STATE:
```
Mehdi-Bendriss, line 367: What does `state == 1` mean? The possible values of the state don't seem [documented](https://www.mongodb.com/docs/manual/reference/command/listShards/#output). 

> An internal field used during the [addShard](https://www.mongodb.com/docs/manual/reference/command/addShard/) sequence to ensure that all steps and configuration necessary to add the shard to the cluster are completed.

src/charm.py
```diff
@@ -646,6 +654,12 @@ def _on_update_status(self, event: UpdateStatusEvent):
                     self.unit.status = WaitingStatus("Waiting for MongoDB to start")
                     return
 
+        # pre-upgrade-check turns off the balancer and waits for cluster to complete upgrade
+        # before being re-enabled.
+        if self.upgrade.is_config_server_waiting_for_resfresh():
```
Mehdi-Bendriss:
```suggestion
        if self.upgrade.is_config_server_waiting_for_refresh():
```

src/upgrades/mongodb_upgrade.py
```diff
@@ -116,6 +122,14 @@ def _on_upgrade_charm(self, _):
         if self.charm.unit.is_leader():
             if not self._upgrade.in_progress:
                 logger.info("Charm upgraded. MongoDB version unchanged")
+
+            if self.charm.is_role(Config.Role.CONFIG_SERVER):
+                # The actions performed as a pre-upgrade check negatively impact cluster
+                # performance i.e. disabling the balancer. Users should NOT run the
+                # pre-upgrade-check action and never run the juju refresh command. Once the users
```
Mehdi-Bendriss, line 129: Being severely penalized because the user ran the pre-upgrade-check action but did not proceed with the upgrade sounds risky.
What is the argument against disabling the balancer on_upgrade if not already disabled?

src/upgrades/mongodb_upgrade.py
```diff
@@ -116,6 +122,14 @@ def _on_upgrade_charm(self, _):
...
+                # have ran the refresh we no longer have to worry about having to wait
+                del self.charm.app_peer_data[Config.Upgrade.WAITING_FOR_REFRESH_KEY]
```
Mehdi-Bendriss, line 131: What happens if the pre-upgrade-check hasn't been run by the user and `_reconcile_upgrade` hasn't run yet?

src/upgrades/mongodb_upgrade.py
```diff
@@ -288,19 +307,89 @@ def is_cluster_healthy(self) -> bool:
...
         except (PyMongoError, OperationFailure, ServerSelectionTimeoutError) as e:
             logger.error(
                 "Cannot proceed with upgrade. Failed to check cluster health, error: %s", e
             )
             return False
 
+    def are_cluster_nodes_healthy(self) -> bool:
+        """Returns true if all nodes in the MongoDB sharded cluster are healthy."""
+        if self.charm.is_role(Config.Role.REPLICATION):
+            return False
```
Mehdi-Bendriss: This may mask a bug along the lines if this is used by a replica set. I believe it's better to drop this check and let it crash in that case.

src/upgrades/mongodb_upgrade.py
```diff
@@ -419,6 +558,82 @@ def step_down_primary_and_wait_reelection(self) -> None:
...
+
+        if not self.is_feature_compatibility_version(Config.Upgrade.FEATURE_VERSION_6):
+            logger.debug(
+                "Not all replicas have the expected feature compatibility: %s",
+                Config.Upgrade.FEATURE_VERSION_6,
+            )
+            return False
+
+        self.set_mongos_feature_compatibilty_version(Config.Upgrade.FEATURE_VERSION_6)
```
Mehdi-Bendriss, line 566: Why is this tied to the upgrade flow? why not on start of the config server?

src/upgrades/mongodb_upgrade.py
```diff
@@ -419,6 +558,82 @@ def step_down_primary_and_wait_reelection(self) -> None:
...
+
+        try:
+            self.turn_off_and_wait_for_balancer()
```
Mehdi-Bendriss: Related to the previous comment - will this run for every unit upgrading?

Mehdi-Bendriss (commented):
Thanks Mia - I have a few questions

src/upgrades/mongodb_upgrade.py
delgod: +1 to Mehdi, some upgrades can be initiated by applying new terraform

src/upgrades/mongodb_upgrade.py
```diff
@@ -310,25 +399,69 @@ def are_replica_set_nodes_healthy(self, mongodb_config: MongoDBConfiguration) ->
...
-            # TODO Future PR - implement node healthy check for sharded cluster
-            pass
-        else:
+        if self.charm.is_role(Config.Role.REPLICATION):
             return self.is_replica_set_able_read_write()
+        else:
+            return self.is_sharded_cluster_able_to_read_write()
+
+    def is_sharded_cluster_able_to_read_write(self) -> bool:
+        """Returns True if possible to write all cluster shards and read from all replicas."""
```
delgod, line 401: I think we should 
1) via mongos, create a shared collection with equal distribution and write 100 values.
2) directly check that all values are distributed across all shards and all replicas correctly.

src/upgrades/mongodb_upgrade.py
```diff
@@ -419,6 +558,82 @@ def step_down_primary_and_wait_reelection(self) -> None:
...
+                        != expected_feature_version
+                    ):
+                        return False
+
+        return True
+
+    def set_mongos_feature_compatibilty_version(self, feature_version) -> None:
+        """Sets the mongos feature compatibility version."""
+        with MongosConnection(self.charm.mongos_config) as mongos:
+            mongos.client.admin.command("setFeatureCompatibilityVersion", "6.0")
```
delgod:
```suggestion
            mongos.client.admin.command("setFeatureCompatibilityVersion", feature_version)
```

src/upgrades/mongodb_upgrade.py
```diff
@@ -419,6 +558,82 @@ def step_down_primary_and_wait_reelection(self) -> None:
...
+
+        try:
+            self.turn_off_and_wait_for_balancer()
+        except BalancerStillRunningError:
+            logger.debug("Balancer is still running. Please try the pre-upgrade check later.")
+            return False
+
+        # The actions performed as a pre-upgrade check negatively impact cluster performance i.e.
+        # disabling the balancer. Users should NOT run the pre-upgrade-check action and never run
+        # the juju refresh command
```
delgod: 1) A disabled balancer can only improve performance 😂  
2) The balancer can be disabled because of maintenance or during backups, so maybe we should check it regularly on status upgrades without setting anything to peer relation. up to @Mehdi-Bendriss

delgod (changes requested):

lib/charms/mongodb/v1/mongos.py
MiaAltieri: As the quote you linked explains, it is to ensure that a shard has been added to the cluster. 

This check is to verify that the config-server added a shard. The integration between a config-server and a shard does not always mean that the shard was added.

src/charm.py
MiaAltieri: lol ty

src/upgrades/mongodb_upgrade.py
MiaAltieri: smart! 🧠 thank you

src/upgrades/mongodb_upgrade.py
MiaAltieri: Ah you are correct, I messed up the order of operations. Let me put this in the correct spot.

`upgrade_unit` calls `reconcile_upgrade` which updates the order of units to upgrade, which results in a relation changed event, which results in `reconcile_upgrade` being fired again, and the eventual call of `upgrade_unit`

src/upgrades/mongodb_upgrade.py
MiaAltieri: Smart - or even better merge `are_cluster_nodes_healthy` and `are_replica_set_nodes_healthy` into one function. Let me check this feasibility.

src/upgrades/mongodb_upgrade.py
MiaAltieri: We could do it there too! But I think it would be better to keep all the upgrade stuff together. Do you see a substantial benefit to moving it to the start of the config server?

The start hook handles a LOT of functionality already and I would prefer to keep that to the bare essentials at this point.

src/upgrades/mongodb_upgrade.py
MiaAltieri: No. the pre-upgrade check runs:
- once on the leader unit (from `juju run pre-upgrade-check`)
- once on the highest unit (from `juju refresh <charm> <options>`)

src/upgrades/mongodb_upgrade.py
MiaAltieri: 🤦‍♀️ - ah yes , thank you

src/upgrades/mongodb_upgrade.py
MiaAltieri: In theory I like this idea, but I think that the MongoDB balancer will not break up the collection across shards until the collection gets much bigger. (i.e. until there is an uneven chunk distribution across shards, where a chunk is 64megabytes by default). We cannot force the balancer to distribute equally across all shards and replicas until the collection is quite large. 

I believe that what we have here is sufficient.

src/upgrades/mongodb_upgrade.py
MiaAltieri: 1. But in the long term disabling the balancer can result in uneven shards and not the most optimised use of resources? But I get what youre saying, reads and wirtes are probably more efficent since data isn't always getting moved around

2. You mean "update status"? or "status upgrades"?

src/upgrades/mongodb_upgrade.py
MiaAltieri: @Mehdi-Bendriss I've modified it so that it only runs when the user has sent the `refresh` command

src/upgrades/mongodb_upgrade.py
MiaAltieri: Actually I have removed this all together given how I've changed the approach of disabling the balancer.

src/upgrades/mongodb_upgrade.py
delgod: up to @Mehdi-Bendriss

src/upgrades/mongodb_upgrade.py
delgod: up to @Mehdi-Bendriss

delgod (dismissed):

src/upgrades/mongodb_upgrade.py
Mehdi-Bendriss: I did some reading about this - I believe we can run this check without waiting for an automated rebalancing of the sharded collection (defined by the 64mb default).
The following steps should do it:
1. Get the count of the current shards
2. Write data in an equally distributed manner (with balanced `shardKey` values, meaning: out of a 100 documents collection and 4 shards, we should get 25 unique documents on each of the 4 shards. Something along the lines of:
```
shards = db.runCommand({listShards: 1})["shards"]
key_shards = [(key, shard["_id"]) for key, shard in zip(range(len(shards)), shards)]
key_shards = [ks for ks in key_shards for _ in range(25)]

sh.shardCollection("testDb.test", { shardKey: 1 })
documents = [{"_id": i, "shardKey": key, "data": shard} for key, shard in key_shards]
collection.insert_many(documents)
```
3. Verify the shard distribution - we should expect each shard to have 25 documents:
```
distribution = collection.getShardDistribution()
# check the count of docs for each shard - which should be equal to 25
```
4. Verify that each replica within the replicaset of each shard contains the 25 documents and that the values of `data` correspond to the name of the queried shard.

src/upgrades/mongodb_upgrade.py
Mehdi-Bendriss: This can be done in a next PR - I think this is a good check to add. 

Given that the balancer can be disabled by operations other than upgrades and/or by the user themselves. We must ensure that this doesn't go unnoticed since we don't want to keep it disabled unless we meant to (temporarily). 

So I think it does make sense to check for the status of the balancer `_on_update_status()` (regardless of an upgrade running or not) - and if the balancer is disabled we put the charm in a Blocked or WaitingStatus: `waiting for balancer to be re-enabled`.

src/upgrades/mongodb_upgrade.py
Mehdi-Bendriss: I don't have a strong opinion about it, but my idea was mainly that this doesn't need to be set multiple times or during upgrades specifically. You can ignore

src/upgrades/mongodb_upgrade.py
MiaAltieri: @Mehdi-Bendriss good point, and something of this sort will be handed in [DPE-4143] next 2week cycle. I promise :)

src/upgrades/mongodb_upgrade.py
MiaAltieri: I like this idea and I am grateful for your research here! It is certainly helpful 😄 , but we have limited time and I want to prioritise wrapping this up since I believe we will be finishing this 1w-ish later than anticipated (granted I've had a cumulative of 2W days off since the start of the 6mo cycle) I think that what we have here is sufficient for our needs , and the suggestion you both provided would only further improve the code ⭐ 

Can we save this as a "would be nice" for the end of the 24.10?

MiaAltieri: So I ran those commands (they needed some small tweaks, which I will list here so that we can remember them in the future)
```
use admin
shards = db.runCommand({listShards: 1})["shards"]

# use the result from above to make this
let shards = [
    {"_id": "shard0"},
    {"_id": "shard1"}
];

let key_shards = [];
for (let i = 0; i < shards.length; i++) {
    key_shards.push([i, shards[i]._id]);
}

let repeated_key_shards = [];
key_shards.forEach(ks => {
    for (let i = 0; i < 25; i++) {
        repeated_key_shards.push(ks);
    }
});

let documents = repeated_key_shards.map((ks, i) => {
    return { _id: i, shardKey: ks[0], data: ks[1] };
});


use testDb

sh.shardCollection("testDb.test", { shardKey: 1 })

let documents = repeated_key_shards.map((ks, i) => {
    return { _id: i, shardKey: ks[0], data: ks[1] };
});
db.test.insertMany(documents)

db.adminCommand( { flushRouterConfig: "test.myShardedCollection" } );
distribution = db.test.getShardDistribution()
```

and the result was:
```
{
  data: '2KiB',
  docs: 50,
  chunks: 1,
  'estimated data per chunk': '2KiB',
  'estimated docs per chunk': 50
}
---
Totals
{
  data: '2KiB',
  docs: 50,
  chunks: 1,
  'Shard shard1': [
    '100 % data',
    '100 % docs in cluster',
    '45B avg obj size on shard'
  ]
}
```

and according to the docs:
> <estDataPercent-x> is a calculated value that reflects, for each shard, the data size as the percentage of the collection's total data size, for example:

which means that even though we had the shard key specified all the data is still on Shard1 (which I assume to be the case due to the chunk size)

Mehdi-Bendriss (dismissed):

Mehdi-Bendriss (approved):

delgod (approved):

PR #424, opened 2024-06-25
Migrate to use version checker as a pip package

MiaAltieri:
## Issue
version checking should be made available to all charms

## Solution
migrate version checking to pip package


Mehdi-Bendriss (approved):
Nice

phvalguima (approved):

PR #425, opened 2024-06-26
[DPE-4667] check unit status before upgrade

MiaAltieri:
## Issue
Pre upgrade check doesn't check if all units in the deployment are active (replica set and sharded clusters)

## Solution
Add a check for individual replica set to see if all units are Active / Waiting for upgrade
Add a check for config-server to verify all units in the cluster are Active / Waiting for upgrade


lib/charms/mongodb/v0/set_status.py
```diff
@@ -0,0 +1,137 @@
...
+
+# Increment this major API version when introducing breaking changes
+LIBAPI = 0
+
+# Increment this PATCH version before using `charmcraft publish-lib` or reset
+# to 0 if you are raising the major API version
+LIBPATCH = 1
+
+
+class MongoDBStatusHanlder(Object):
```
phvalguima: Typo

src/upgrades/mongodb_upgrade.py
```diff
@@ -279,14 +281,19 @@ def is_cluster_healthy(self) -> bool:
...
             logger.error(
-                "Cannot proceed with upgrade. Unit is not in Active state, in: %s.", unit_state
+                "Cannot proceed with upgrade. Status of charm units do not show active / waiting for upgrade."
             )
             return False
 
+        if self.charm.is_role(Config.Role.CONFIG_SERVER):
+            if not self.charm.status.are_shards_status_active_or_waiting_for_upgrade():
+                logger.error(
+                    "Cannot proceed with upgrade. Status of shard units do not show active / waiting for upgrade."
```
phvalguima, line 308: So, you will fall in this "if" every time your cluster is all-green as well. If this is a check you do, say, every hook, then you will log this error every hook.

lib/charms/mongodb/v0/set_status.py
```diff
@@ -0,0 +1,137 @@
...
+
+        for sharding_relation in self.charm.config_server.get_all_sharding_relations():
+            for _, unit_data in sharding_relation.data.items():
+                status_type = unit_data.get(Config.Status.STATUS_TYPE_KEY, None)
+                status_messge = unit_data.get(Config.Status.STATUS_MESSAGE_KEY, None)
+                if status_type is None:
+                    return False
+                if "ActiveStatus" in status_type:
+                    continue
+                if "WaitingStatus" not in status_type:
```
phvalguima: Why BlockedStatus or ErrorStatus are not considered either?

src/charm.py
```diff
@@ -21,6 +21,7 @@
 )
 from charms.mongodb.v0.mongodb_secrets import SecretCache, generate_secret_label
 from charms.mongodb.v0.mongodb_tls import MongoDBTLS
+from charms.mongodb.v0.set_status import MongoDBStatusHanlder
```
phvalguima: Typo

src/charm.py
```diff
@@ -164,6 +165,7 @@ def __init__(self, *args):
         self.config_server = ShardingProvider(self)
         self.cluster = ClusterProvider(self)
         self.shard = ConfigServerRequirer(self)
+        self.status = MongoDBStatusHanlder(self)
```
phvalguima: Typo

lib/charms/mongodb/v0/set_status.py
phvalguima: If this class is only used to check if shards are active OR upgrading, why then not having two flags in the relation instead?

phvalguima (changes requested):

lib/charms/mongodb/v0/set_status.py
MiaAltieri: if it is in BlockedStatus or Error status, then we don't care about the message and we would immediately return False. Which is intended.

We reserve BlockedStatus and ErrorStatus for when the user needs to intervene and change something. If that is the case then that unit / app is certainly not ready for upgrade

src/charm.py
MiaAltieri: I am embarrased lmao

src/upgrades/mongodb_upgrade.py
MiaAltieri: if the cluster is all-green then this doesn't log? 

But also we only do this check for upgrade related hooks :)

lib/charms/mongodb/v0/set_status.py
MiaAltieri: I suppose we could! Thats a great point. Might help more with backwards compatibility in the future. @Mehdi-Bendriss what do you think?

lib/charms/mongodb/v0/set_status.py
phvalguima: Okay! I will leave this item open if anyone else reviewing has the same question.

lib/charms/mongodb/v0/set_status.py
```diff
@@ -0,0 +1,137 @@
...
+        """Constructor for CrossAppVersionChecker.
+
+        Args:
+            charm: charm to inherit from
+            version: (int), the current version of the desired attribute of the charm
+            relations_to_check: (List), a list of relations who should have compatible versions
+                with the current charm
+            version_validity_range: (Optional Dict), a list of ranges for valid version ranges.
+                If not provided it is assumed that relations on the provided interface must have
+                the same version.
```
Mehdi-Bendriss:
```suggestion
            charm: charm instantiating this class.
```

lib/charms/mongodb/v0/set_status.py
```diff
@@ -0,0 +1,137 @@
...
+
+        self.set_app_status()
+
+        if self.charm.is_role(Config.Role.SHARD):
+            self.share_status_to_config_server()
+
+    def set_app_status(self):
+        """TODO Future Feature/Epic: parse statuses and set a status for the entire app."""
+
+    def are_charm_units_all_active_or_waiting_for_upgrade(self) -> bool:
```
Mehdi-Bendriss: From the docstrings, we can perhaps adopt a shorter name. Perhaps: `are_all_units_ready_for_upgrade`

lib/charms/mongodb/v0/set_status.py
```diff
@@ -0,0 +1,137 @@
...
+        goal_state = self.charm.model._backend._run(
+            "goal-state", return_output=True, use_json=True
+        )
+        for _, unit_state in goal_state["units"].items():
+            if unit_state["status"] == "active":
+                continue
+            if unit_state["status"] != "waiting":
+                return False
+
+            if not self.charm.get_cluster_mismatched_revision_status():
```
Mehdi-Bendriss: This only runs for the current unit

lib/charms/mongodb/v0/set_status.py
```diff
@@ -0,0 +1,137 @@
...
+                return False
+
+        return True
+
+    def are_shards_status_active_or_waiting_for_upgrade(self) -> bool:
```
Mehdi-Bendriss: same as above

lib/charms/mongodb/v0/set_status.py
```diff
@@ -0,0 +1,137 @@
...
+        return True
+
+    def share_status_to_config_server(self):
+        """Shares this shards status info to the config server."""
+        if not self.charm.is_role(Config.Role.SHARD):
+            return
+
+        config_relation = self.charm.shard.get_config_server_relation()
+
+        if not config_relation:
```
Mehdi-Bendriss: nit: this would be a good usage of the walrus operator.
```suggestion
        if not (config_relation := self.charm.shard.get_config_server_relation()):
```

lib/charms/mongodb/v0/set_status.py
```diff
@@ -0,0 +1,137 @@
...
+            return
+
+        # prevent changing data and causing an unnecessary relation changed hook
```
Mehdi-Bendriss: does storing the same value as the one existing trigger a rel changed event?

Mehdi-Bendriss (commented):
Thanks Mia - I have a couple of questions

lib/charms/mongodb/v0/set_status.py
MiaAltieri: can you tell I am a fan of copy paste

lib/charms/mongodb/v0/set_status.py
MiaAltieri: TWIL this operator is called the walrus operator and its so cute I love it

lib/charms/mongodb/v0/set_status.py
phvalguima: No, it won't.

phvalguima (dismissed):
LGTM - I left small comment on one of Mehdi's point only

lib/charms/mongodb/v0/set_status.py
MiaAltieri: Ah - great to know, I will update this code here. Thank you so much Pedro :)

lib/charms/mongodb/v0/set_status.py
MiaAltieri: Yes, but it should have the same result across all units so it should be no big deal :)

Mehdi-Bendriss: But since the output is the same, it should be taken out of the loop :)

lib/charms/mongodb/v0/set_status.py
Mehdi-Bendriss: Maybe I missed something; but can units' relation data be accessed by external applications?

lib/charms/mongodb/v0/set_status.py
MiaAltieri: Do you mean take `self.charm.get_cluster_mismatched_revision_status()` out of the loop?

I could be wrong but I think the value of `self.charm.get_cluster_mismatched_revision_status()` will be cached and it shouldnt hurt the performance

Mehdi-Bendriss (dismissed):

Mehdi-Bendriss (approved):

PR #428, opened 2024-07-04
[DPE-4755] [DPE-4753] Implement upgrade proceedure and basic integration tests for upgrading a sharded cluster

MiaAltieri:
## Issue
Missing tests that verify:
1. success of pre-upgrade check
2. failure of pre-upgrade check
3. success of upgrade operations

## Solution
Add missing tests

## Notes:
The upgrade test is disabled until the upgrade work is published on charmhub, but the test suite was ran manually by building the charm with an earlier snap and upgrading the charm to a newer snap revision

## Future Work
Once the upgrade work is published on charm hub
1. Deploy charmcraft charm in the test deployment step
2. Re-enable upgrade test


lib/charms/mongodb/v0/set_status.py
```diff
@@ -105,8 +105,8 @@ def share_status_to_config_server(self):
 
     def is_unit_status_ready_for_upgrade(self) -> bool:
         """Returns True if the status of the current unit reflects that it is ready for upgrade."""
-        current_status = type(self.charm.unit.status)
-        status_message = self.charm.unit.status.message
+        current_status = self.charm.unit.status
```
MiaAltieri, line 108: this was causing a bug

lib/charms/mongodb/v0/set_status.py
```diff
@@ -105,20 +105,17 @@ def share_status_to_config_server(self):
...
 
         if not isinstance(current_status, WaitingStatus):
             return False
 
-        if (
-            status_message
-            and status_message != Config.Status.CONFIG_SERVER_WAITING_FOR_REFRESH.message
-        ):
-            return False
+        if status_message and "is not up-to date with config-server" in status_message:
```
MiaAltieri, line 116: looking for `CONFIG_SERVER_WAITING_FOR_REFRESH ` was incorrect, we should be checking if the shard has the wrong revision ...

src/config.py
```diff
@@ -116,10 +115,8 @@ class Status:
         """
 
         STATUS_READY_FOR_UPGRADE = "status-shows-ready-for-upgrade"
-        CONFIG_SERVER_WAITING_FOR_REFRESH = WaitingStatus("Waiting for refresh command.")
 
     class Upgrade:
         """Upgrade related constants."""
 
-        WAITING_FOR_REFRESH_KEY = "waiting_for_refresh"
```
MiaAltieri, line 124: This was an artefact from some old work. It can be removed as it is no longer used

Context: this was needed when we used to disable the balancer during the pre-upgrade check. But now that we disable it when the `refresh` command is given, we no longer need to "wait for refresh"

src/upgrades/mongodb_upgrade.py
```diff
@@ -123,13 +123,6 @@ def _on_upgrade_charm(self, _):
             if not self._upgrade.in_progress:
                 logger.info("Charm upgraded. MongoDB version unchanged")
 
-            if self.charm.is_role(Config.Role.CONFIG_SERVER):
```
MiaAltieri, line 126: no longer used

Context: this was needed when we used to disable the balancer during the pre-upgrade check. But now that we disable it when the `refresh` command is given, we no longer need to "wait for refresh"

src/upgrades/mongodb_upgrade.py
```diff
@@ -626,13 +619,6 @@ def turn_off_and_wait_for_balancer(self):
             if balancer_state["mode"] != "off":
                 raise BalancerStillRunningError("balancer is still Running.")
 
-    def is_config_server_waiting_for_refresh(self) -> bool:
```
MiaAltieri, line 629: no longer used

Context: this was needed when we used to disable the balancer during the pre-upgrade check. But now that we disable it when the `refresh` command is given, we no longer need to "wait for refresh"

tests/integration/sharding_tests/test_sharding_tls.py
```diff
@@ -214,44 +218,6 @@ async def check_cluster_tls_enabled(ops_test: OpsTest) -> None:
         ), f"Mongos TLS not enabled in unit {unit.name}"
 
 
-async def deploy_cluster_components(ops_test: OpsTest) -> None:
```
MiaAltieri, line 217: migrate these to sharding test helpers so they can be used by other tests ⭐

tests/integration/sharding_tests/helpers.py
```diff
@@ -107,3 +115,68 @@ def count_users(mongos_client: MongoClient) -> int:
     admin_db = mongos_client["admin"]
     users_collection = admin_db.system.users
     return users_collection.count_documents({})
+
+
+async def deploy_cluster_components(
```
MiaAltieri, line 120: These helpers were already made in the TLS tests for sharding, but are quite useful for other tests so I moved them into the sharding test helpers  ⭐

lib/charms/mongodb/v0/set_status.py
phvalguima: @MiaAltieri can we have it as a constant instead of just a string? Also update the place where the charm message is set to use this constant.

tests/integration/upgrade/test_sharding_upgrade.py
```diff
@@ -0,0 +1,160 @@
...
+    )
+    await integrate_cluster(ops_test)
+    await ops_test.model.wait_for_idle(
+        apps=CLUSTER_COMPONENTS, status="active", idle_period=20, timeout=TIMEOUT
+    )
+
+
+@pytest.mark.group(1)
+@pytest.mark.abort_on_fail
+async def test_upgrade(
```
phvalguima, line 59: We should also have a test to try the "rollback" in the middle of the upgrade.

tests/integration/ha_tests/continuous_writes.py
```diff
@@ -30,15 +38,11 @@ def continous_writes(connection_string: str, starting_number: int):
             ).update_one({"number": write_value}, {"$set": {"number": write_value}}, upsert=True)
 
             # update_one
-        except (NotPrimaryError, AutoReconnect):
-            # this means that the primary was not able to be found. An application should try to
-            # reconnect and re-write the previous value. Hence, we `continue` here, without
+        except PyMongoError:
+            # PyMongoErors should result in an attempt to retry a write. An application should
```
Mehdi-Bendriss, line 42: Nice

Mehdi-Bendriss (commented):
Thanks Mia - looks good, we also need testing of rollbacks as Pedro suggested

lib/charms/mongodb/v0/set_status.py
MiaAltieri: We could but the actual message is much longer - it includes the revision numbers of both the config-server and the shard, so doing
>  update the place where the charm message is set to use this constant.

would actually be more complicated than just that constant 

Which is also why I wouldn't want to add only a constant for this one case. If this is a deal breaker though let me know!

tests/integration/upgrade/test_sharding_upgrade.py
MiaAltieri: 100% agree. Unfortunately we cannot automate this test until the charm is published on charmhub. I wrote a version of the tests that we can update once this happens

Mehdi-Bendriss (dismissed):

phvalguima (dismissed):

Mehdi-Bendriss (approved):

Mehdi-Bendriss (approved):

phvalguima (approved):

PR #429, opened 2024-07-09
[DPE-4143] Add a post upgrade check for cluster upgrades

MiaAltieri:
## Issue
After the cluster upgrades the balancer stays disabled (a requirement for the upgrade procedure to happen)

## Solution
Wait for all cluster components to successfully upgrade and then verify cluster health. Once this is done we:
1. Re-enable the balancer
2. Set the new feature version across the cluster

## Other issues found along the way
The check for status across the cluster is faulty in these ways:
1. If the upgrade fails and sets the unit status to being Failed/Blocked, it prevents the post upgrade check from ever succeeding
2. It is not flexible in how it checks mismatched cluster statuses across the cluster.

Charm has no idea when a local cluster has completed an upgrade. (Solution was to add an internal charm revision)

## Charm internal version CI workflow
This workflow ensures that we bump the number on this file each time we make a PR.
We expect this workflow to fail right now since the file does not exist on main branch. For that reason we expect it to fail. Once the PR merges we will make a requirement of all future PRs


charm_internal_version
```diff
@@ -0,0 +1 @@
+1
```
MiaAltieri, line 1: @phvalguima - this is the internal revision I mentioned yesterday.

**Problem**
In local deployments we are unable to tell if the the shards/config-server are the same revision / source code. (A requirement of sharded deplyoments)

This is an additional problem when we upgrade a cluster. Without knowing if their revision / source code is the same we do not know at what point a cluster has all the same source code. 

**Solution**
Add an internal revision number so that we can know the version of source code someone is using (only used in local deployments) 

**TL;DR**
The purpose of this is to have a better idea of what version a local charm is. (it does not need to match what is on charm hub)

.github/workflows/internal_version_check.yaml
```diff
@@ -0,0 +1,33 @@
...
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
+      - name: Compare with internal version on main branch
+        run: |
+          NEW_VERSION="$(< charm_internal_version)"
+          mkdir repo; cd repo
+          git clone https://github.com/canonical/mongodb-operator.git
+          cd mongodb-operator
+          OLD_VERSION="$(< charm_internal_version)"
```
Mehdi-Bendriss: This will fail if the file doesn't exist already, we should perhaps set a default to `0` if it's not there.
```suggestion
          OLD_VERSION="$([ -f charm_internal_version ] && cat charm_internal_version || echo 0)"
```

.github/workflows/internal_version_check.yaml
```diff
@@ -0,0 +1,33 @@
...
+          [ "$NEW_VERSION" -eq "$EXPECTED_NEW_VERSION" ]
```
Mehdi-Bendriss: Leftover probably, safe to delete

.github/workflows/internal_version_check.yaml
```diff
@@ -0,0 +1,33 @@
...
+    name: Check Internal Version Incremented
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
+      - name: Compare with internal version on main branch
+        run: |
+          NEW_VERSION="$(< charm_internal_version)"
+          mkdir repo; cd repo
+          git clone https://github.com/canonical/mongodb-operator.git
+          cd mongodb-operator
```
Mehdi-Bendriss: nit:
```suggestion
          git clone https://github.com/canonical/mongodb-operator.git repo/mongodb-operator
          cd repo/mongodb-operator
```

lib/charms/mongodb/v0/set_status.py
```diff
@@ -54,19 +54,51 @@ def set_and_share_status(self, status: StatusBase):
...
+        function returning False g and preventing the status from being reset.
+        """
+        if isinstance(self.charm.unit.status, ActiveStatus):
+            return True
+
+        if ignore_unhealthy_upgrade and self.charm.unit.status == Config.Status.UNHEALTHY_UPGRADE:
+            return True
+
+        return self.is_status_related_to_mismatched_revision(
+            str(type(self.charm.unit.status)).lower()
```
Mehdi-Bendriss: This will return something along the lines of `"<class 'ops.model.ActiveStatus'>"`, may be safer to get the class name directly:
```suggestion
            type(self.charm.unit.status).__name__.lower()
```

lib/charms/mongodb/v0/set_status.py
```diff
@@ -54,19 +54,51 @@ def set_and_share_status(self, status: StatusBase):
...
+        )
+
+    def is_status_related_to_mismatched_revision(self, status_type: str) -> bool:
+        """Returns True if the current status is related to a mimsatch in revision.
+
+        Note: A few functions call this who receive states differently. One receives them by
```
Mehdi-Bendriss:
```suggestion
        Note: A few functions calling this method receive states differently. One receives them by
```

src/charm.py
```diff
@@ -363,6 +361,12 @@ def upgrade_in_progress(self):
             return False
         return self.upgrade._upgrade.in_progress
 
+    @property
+    def get_charm_internal_revision(self):
```
```suggestion
    def get_charm_internal_revision(self) -> int:
```

src/upgrades/machine_upgrade.py
```diff
@@ -169,7 +169,7 @@ def upgrade_unit(self, *, charm) -> None:
         # According to the MongoDB documentation, before upgrading the primary, we must ensure a
         # safe primary re-election.
         try:
-            if self._unit.name == charm.primary and len(self._sorted_units) > 1:
+            if self._unit.name == charm.primary:
```
Mehdi-Bendriss, line 172: Do we need to wait for the reelection to happen in the case where we only have a 1 node deployment?

src/upgrades/machine_upgrade.py
```diff
@@ -185,9 +185,12 @@ def upgrade_unit(self, *, charm) -> None:
         self._unit_workload_version = self._current_versions["workload"]
         logger.debug(f"Saved {_SNAP_REVISION} in unit databag after upgrade")
 
+        if charm.unit == self._sorted_units[-1]:
```
Mehdi-Bendriss, line 190: Could you explain this check?

src/upgrades/mongodb_upgrade.py
```diff
@@ -289,7 +340,11 @@ def is_cluster_healthy(self) -> bool:
                 logger.error("Cannot proceed with upgrade. Service mongod is not running")
                 return False
 
-        if not self.charm.status.are_all_units_ready_for_upgrade():
+        if not self.charm.status.is_current_unit_ready(
+            ignore_unhealthy_upgrade=True
```
Mehdi-Bendriss, line 346: Doesn't this wrong the output of `is_cluster_healthy`? given that not all the cluster is effectively healthy. And do we want to proceed with the upgrade in this case?

Mehdi-Bendriss (commented):
Thanks Mia! I left a few questions

src/upgrades/machine_upgrade.py
MiaAltieri: The re-election code checks in the case that there is a one node deployment

src/upgrades/machine_upgrade.py
MiaAltieri: sure - thing I will add a comment

src/upgrades/mongodb_upgrade.py
MiaAltieri: It's actually more nuanced than that - let me add a comment

src/upgrades/mongodb_upgrade.py
```diff
@@ -187,46 +194,90 @@ def _on_force_upgrade_action(self, event: ActionEvent) -> None:
...
+
+        # Leader of config-server must wait for all shards to be upgraded before finialising the
+        # upgrade.
+        if not self.charm.unit.is_leader() or not self.charm.is_role(Config.Role.CONFIG_SERVER):
+            return
+
+        self.charm.upgrade.post_cluster_upgrade_event.emit()
+
+    def run_post_cluster_upgrade_check(self, event: EventBase) -> None:
+        """Waits for entire cluster to be upgraded before enabling the balancer."""
```
phvalguima, line 218: We are starting things and waiting for them. This is not a "check", but a "post_upgrade_task". If that is correct, can we rename this method and their functions to "_task" instead of "_check"?

src/upgrades/mongodb_upgrade.py
```diff
@@ -187,46 +194,90 @@ def _on_force_upgrade_action(self, event: ActionEvent) -> None:
...
+            return
+
+        logger.debug(
+            "Entire cluster has been upgraded, checking health of the cluster and enabling balancer."
+        )
+        self.run_post_upgrade_checks(event, finished_whole_cluster=True)
 
+        try:
+            with MongosConnection(self.charm.mongos_config) as mongos:
+                mongos.start_and_wait_for_balancer()
```
phvalguima, line 236: See my comment above

lib/charms/mongodb/v0/set_status.py
```diff
@@ -54,19 +54,51 @@ def set_and_share_status(self, status: StatusBase):
     def set_app_status(self):
         """TODO Future Feature/Epic: parse statuses and set a status for the entire app."""
 
-    def are_all_units_ready_for_upgrade(self) -> bool:
+    def is_current_unit_ready(self, ignore_unhealthy_upgrade: bool = False) -> bool:
+        """Returns True if the current unit status shows that the unit is ready.
+
+        Note: we allow the use of ignore_unhealthy_upgrade, to avoid infinite loops due to this
+        function returning False g and preventing the status from being reset.
```
phvalguima:
```suggestion
        function returning False and preventing the status from being reset.
```

lib/charms/mongodb/v0/set_status.py
```diff
@@ -54,19 +54,51 @@ def set_and_share_status(self, status: StatusBase):
...
+        """Returns True if the current status is related to a mimsatch in revision.
+
+        Note: A few functions calling this method receive states differently. One receives them by
+        "goal state" which processes data differently and the other via the ".status" property.
+        Hence we have to be flexible to handle each.
+        """
+        if not self.charm.get_cluster_mismatched_revision_status():
+            return False
+
+        if "waiting" in status_type and self.charm.is_role(Config.Role.CONFIG_SERVER):
```
phvalguima, line 83: We only have "WaitingStatus" on config server if there were version mismatches?

phvalguima (commented):
Hi @MiaAltieri, my main question is how are we managing the status. Given you're using the status to define if we are ready or not to move on to the upgrade; how are we sure the "waiting" you see is indeed an upgrade-related "waiting" status?

lib/charms/mongodb/v0/set_status.py
MiaAltieri: Yes - see the section titled [statuses in the spec](https://docs.google.com/document/d/1_UntlfVjGwEvv5HgsaQo_o8XnJAR6Vm7_cuM9c2mPIk/edit?usp=sharing) :)

Mehdi-Bendriss (approved):

phvalguima (approved):

PR #430, opened 2024-07-12
[DPE-4757] - Update shared library code, so that mongos waits for upgrade

MiaAltieri:
## Context - What is this file?
This library code is used between `mongos`-charm and `mongodb`-charm (when Charmed MongoDB is running as a config-server), to manage connections between routers and config-servers. In this lib: 
- Charmed mongos requests the user and shares its user with its associated client application
- Charmed mongos manages its connection with the associated config-server
- Charmed mongos starts its `mongos` service (as it is required to have a config-server before starting)

This file is "owned" by Charmed MongoDB - so any direct changes to the file **must** be made in this repo, before copying them over to Charmed Mongos

## Issue
`mongos` does not wait for upgrades to finish before processing integrations to the `config-server`. To prevent any broken states, we should finish an upgrade before integrating to the config-server.

## Solution
Check if an upgrade is on-going. If it is prevent the relevant hook from executing.

## What is next
Update the `mongos` charm to have the newest version of this lib to ensure that `mongos` waits for upgrades to finish before


lib/charms/mongodb/v0/config_server_interface.py
```diff
@@ -286,6 +293,7 @@ def _on_relation_changed(self, event) -> None:
             return
 
         self.charm.status.set_and_share_status(ActiveStatus())
+        self.charm.mongos_intialised = True
```
MiaAltieri, line 297: This was an additional change made to the lib.

This will be used to verify if we have already configured + started the mongos process.

Mehdi-Bendriss (dismissed):

Gu1nness (dismissed):

Mehdi-Bendriss (dismissed):

Gu1nness (dismissed):

Gu1nness (dismissed):

Mehdi-Bendriss (dismissed):

Mehdi-Bendriss (approved):

Gu1nness (approved):

PR #433, opened 2024-07-15
[DPE-4663]: Fix flaky sharding + tls tests

Gu1nness:
## Issue
 
 * The Sharding + TLS test is flaky on `test_tls_inconsistent_rels`.
 * Sometimes it stalls and a unit stays in state `[active] executing` after we remove TLS
 * Sometimes, `juju status` returns an error because a volume is defined but not filesystem associated with it can be found.

## Solution
 * First problem : the `destroy_cluster` method was using `ops_test.fast_forward` in an unsafe way which globally lowered the `update-status-hook-interval` to 10s which is way too low and would stall a unit.
 * Second problem : Update `destroy_cluster` to destroy the storage safely when destroying the cluster so we are sure that the volumes are cleaned before doing anymore tests.
 * On top of that, two minor fixes, one making a call to `juju status` fail on errors, and an update in the charm to prevent `AttributeError`s when accessing `self.peers.units` in an unsafe way.


src/charm.py
```diff
@@ -314,6 +312,14 @@ def peers(self) -> Optional[Relation]:
         """
         return self.model.get_relation(Config.Relations.PEERS)
 
+    @property
+    def peers_units(self) -> list[Unit]:
+        """Peers units in a safe way."""
```
MiaAltieri: nit
```suggestion
        """Get peers units in a safe way."""
```

src/charm.py
```diff
@@ -314,6 +312,14 @@ def peers(self) -> Optional[Relation]:
         """
         return self.model.get_relation(Config.Relations.PEERS)
 
+    @property
+    def peers_units(self) -> list[Unit]:
```
MiaAltieri, line 316: BTW this was an excellent find and an excellent fix. I would be curious to see where else this sort of logic could be beneficial

tests/integration/helpers.py
```diff
@@ -362,21 +362,23 @@ async def get_unit_hostname(ops_test: OpsTest, unit_id: int, app: str) -> str:
     return hostname.strip()
 
 
-def get_raw_application(ops_test: OpsTest, app: str) -> Dict[str, Any]:
+async def get_raw_application(ops_test: OpsTest, app: str) -> Dict[str, Any]:
     """Get raw application details."""
-    return json.loads(
-        subprocess.check_output(
```
MiaAltieri, line 368: Nothing for you to change here, just some context

I believe the idea here of using `subprocess` instead of doing something like `ops.model.application.status` or something like that is that sometimes `ops_test` isn't the most reliable when it comes to statuses

here you use `ops_test.juju` to retrieve the status (doing essentially what we were doing before), which should be fine, even if it does rely on `ops_test`.

MiaAltieri (dismissed):
Great work thank you very much Neha!

src/charm.py
Gu1nness: Thanks a lot!
I'll take some time to see if there are similar cases where we should do the same and if that's the case, I'll add all those functions :)

tests/integration/helpers.py
Gu1nness: Ok!
I chose to go to `ops_test.juju` because it seemed more canonical to me to use it and this way I could get the `ret_code` and `stderr` (I could have using subprocess, it's more that it's not a library that I like using tbh)

MiaAltieri (approved):
Thank you!

Mehdi-Bendriss (approved):

PR #435, opened 2024-07-19
[DPE-4897]: Stabilize sharding race condition tests

Gu1nness:
## Issue

 * The sharding race condition test is flaky.
 * The test and the code seemed good on the Sharding side.
 * There was actually another race condition which was making this test flaky.

## Solution

 * The race condition that was happening is the following: the config server can run the `RelationJoined` hook before the shard has run `RelationCreated`. The `RelationJoined` hook relies on some information sent by the `RelationCreated` hook to succeed. The solution is hence to defer this event in the case we have some incompatible versions between shards and config server, or if we are missing those shards. This is done also for the shard because otherwise, the same issue could happen, we dropped the event and would never sync the credentials.
 * This patch reuses the already existing code as much at it can.
 * Adds some tests to ensure that this never happens againt

## Miscellaneous

 * Fix some small issues in the `test_config_server_lib.py` tests.


lib/charms/mongodb/v1/shards_interface.py
```diff
@@ -153,7 +153,13 @@ def pass_sanity_hook_checks(self, event: EventBase) -> bool:
...
 
         if not self.charm.is_relation_feasible(self.relation_name):
-            logger.info("Skipping event %s , relation not feasible.", str(type(event)))
+            if self.charm.status.is_status_related_to_mismatched_revision(
+                self.charm.unit.status.name
+            ):
+                logger.info("Deferring %s. Mismatched versions in the cluster.", str(type(event)))
+                event.defer()
+            else:
+                logger.info("Skipping event %s , relation not feasible.", str(type(event)))
```
MiaAltieri: Nit: Removing nested structure of if statement
```suggestion
      if self.charm.status.is_status_related_to_mismatched_revision(
           self.charm.unit.status.name
      ):
            logger.info("Deferring %s. Mismatched versions in the cluster.", str(type(event)))
            event.defer()
       elif not self.charm.is_relation_feasible(self.relation_name):
           logger.info("Skipping event %s , relation not feasible.", str(type(event)))
```

lib/charms/mongodb/v1/shards_interface.py
```diff
@@ -764,7 +770,13 @@ def pass_sanity_hook_checks(self, event: EventBase) -> bool:
...
```
MiaAltieri: Nit: Removing nested structure of if statement
```suggestion
      if self.charm.status.is_status_related_to_mismatched_revision(
           self.charm.unit.status.name
      ):
            logger.info("Deferring %s. Mismatched versions in the cluster.", str(type(event)))
            event.defer()
       elif not self.charm.is_relation_feasible(self.relation_name):
           logger.info("Skipping event %s , relation not feasible.", str(type(event)))
```

src/charm.py
```diff
@@ -1075,7 +1076,7 @@ def install_snap_packages(self, packages: List[Package]) -> None:
 
     def _instatiate_keyfile(self, event: StartEvent) -> None:
         # wait for keyFile to be created by leader unit
-        if not self.get_secret(APP_SCOPE, Config.Secrets.SECRET_KEYFILE_NAME):
+        if not (keyfile := self.get_secret(APP_SCOPE, Config.Secrets.SECRET_KEYFILE_NAME)):
```
MiaAltieri, line 1079: nice 😎

src/charm.py
```diff
@@ -1661,6 +1662,8 @@ def get_cluster_mismatched_revision_status(self) -> Optional[StatusBase]:
                 return BlockedStatus(
                     f"Charm revision ({current_charms_version}{local_identifier}) is not up-to date with config-server."
                 )
+            if self.is_role(Config.Role.CONFIG_SERVER):
+                return WaitingStatus("No information received yet from at least one shard.")
```
MiaAltieri: So this is a tricky one, we could have a `NoVersionError` due to:
1. race-condition, i.e. waiting for the information to be shared through `RelationCreatedEvent`
2. the shards are an older revision that do not share their status

In case 1. the situation should resolve itself fairly quickly, if there is a race condition causing this, it is because the users immediately related the applications and they are waiting to "settle" in this case we will receive a flood of juju events and it should resolve itself quite quickly. 

In case 2. the situation will not resolve itself without the user doing something on the shard side.

In general we want to avoid setting "unnecessary statuses", determining which statuses are unnecessary is an art and I am still by no means perfect at this. 

For these reasons I would suggest to not report a status for case 1. and for case 2. report the following.

```suggestion
            if self.is_role(Config.Role.CONFIG_SERVER):
                return WaitingStatus(
                f"Waiting for shards to upgrade/downgrade to revision {current_charms_version}{local_identifier}."
            )
```

If you **really wanted to** you could add some sort of sanity check 
```
            if self.is_role(Config.Role.CONFIG_SERVER) and <has shard already started>:
                return WaitingStatus( f"Waiting for shards to upgrade/downgrade to revision {current_charms_version}{local_identifier}."
```
but this might be a little overkill

MiaAltieri (changes requested):
Looks great! Thank you so much for your work here! Super appreciative!

lib/charms/mongodb/v1/shards_interface.py
Gu1nness: The suggestion doesn't work because the call to `self.charm.status.is_status_related_to_mismatched_revision` requires `self.charm.is_relation_feasible` to [have run](https://github.com/canonical/mongodb-operator/blob/6/edge/src/charm.py#L1621-L1623).
However I agree that this can be improved and I will come up with a not-nested version of this.

lib/charms/mongodb/v1/shards_interface.py
Gu1nness: (Same comment as the former one)

src/charm.py
Gu1nness: We can keep the code as it was then because the status will be handled by the code below this `try/catch`.
I'll remove this part!

lib/charms/mongodb/v1/shards_interface.py
```diff
@@ -764,8 +778,7 @@ def pass_sanity_hook_checks(self, event: EventBase) -> bool:
             return False
 
         if not self.charm.is_relation_feasible(self.relation_name):
-            logger.info("Skipping event %s , relation not feasible.", str(type(event)))
-            return False
+            self.__handle_relation_not_feasible(event)
```
MiaAltieri: move this check to after `if not self.charm.is_role(Config.Role.SHARD):`

reason: `__handle_relation_not_feasible ` defers, we don't want to end up in an infinite defer cycle if the component is not a shard. (same for above)

MiaAltieri (changes requested):

lib/charms/mongodb/v1/shards_interface.py
Gu1nness: Done!

MiaAltieri (dismissed):
Great work! thank you very much Neha!

lib/charms/mongodb/v1/shards_interface.py
Gu1nness: Actually, this has to be higher in the code as there are some checks that won't be run and some statuses won't be set (including the ones on the relation status and the current role) so I will be reverting this last commit and we will not end up in a infinite defer loop thanks to the checks in `is_relation_feasible`

lib/charms/mongodb/v1/shards_interface.py
```diff
@@ -145,6 +145,13 @@ def _on_relation_joined(self, event):
 
         self.database_provides.update_relation_data(event.relation.id, relation_data)
 
+    def __handle_relation_not_feasible(self, event: EventBase):
```
Mehdi-Bendriss: **nit:** though the `__` is commonly used for name mangling of "private" methods, it is not really necessary in the context of charms / charm libs, where inheritance is performed rather modestly, and where most of the client code is actual charm components aware of what (not) to use

lib/charms/mongodb/v1/shards_interface.py
```diff
@@ -145,6 +145,13 @@ def _on_relation_joined(self, event):
...
+        if self.charm.status.is_status_related_to_mismatched_revision(self.charm.unit.status.name):
```
Mehdi-Bendriss, line 149: Given this method only performs a check/deferral - would it make sense to directly put the logic in `pass_sanity_hook_checks`? Or are you expecting this method to perform additional logic in the future?

Mehdi-Bendriss (commented):
Thanks Neha! Just a minor question

lib/charms/mongodb/v1/shards_interface.py
Gu1nness: Sure, I'll rename it then :)

lib/charms/mongodb/v1/shards_interface.py
Gu1nness: I'm not expecting it to perform additional logic, I moved it from `pass_sanity_hook_checks` mainly to remove some nesting in the main function after [this](https://github.com/canonical/mongodb-operator/pull/435#discussion_r1687577879) comment from Mia.
I'm fine with both, I just find this code easier to read.

In the future, we should improve this `pass_sanity_hook_checks` and the way the `is_relation_feasible` works because `is_relation_feasible` (which looks like a `GET` function) does sets the status and it would be better that this function either returns nothing (AKA all is good) or a status that we can then handle and set. This helps decorrelating the functionality and makes a code easier to read and understand.

Mehdi-Bendriss (approved):

MiaAltieri (approved):

PR #437, opened 2024-07-24
[DPE-4912] Resolve library version issues

Gu1nness:
## Issue

 * Library `set_status` was never published
 * Library `mongodb_tls` was still on 0.X while 1.0 has been published on charmhub.

## Solution

 * Move `mongodb_tls` to v1.1
 * Reset `set_status` to v0.1 so it's published with the right version.

PR #438, opened 2024-07-24
[DPE-4906] - Make libs k8s friendly 

MiaAltieri:
## Issue
1. MongoDB K8s charm cannot use some of the shared code in the lib
2. MongoDB K8s charm is copying over an excessive amount of the status code 

## Solution
1. use a universal name to retrieve hosts across charms
2. move shared status code to `set_status` lib

## Lib versions
currently many of the library API + Patch are out of date, so we do not bump the versions here [this PR when it lands will resolve them](https://github.com/canonical/mongodb-operator/pull/437)


lib/charms/mongodb/v0/set_status.py
```diff
@@ -150,4 +161,98 @@ def is_unit_status_ready_for_upgrade(self) -> bool:
...
+            if replica_status == "PRIMARY":
+                return ActiveStatus("Primary")
+            elif replica_status == "SECONDARY":
+                return ActiveStatus("")
+            elif replica_status in ["STARTUP", "STARTUP2", "ROLLBACK", "RECOVERING"]:
+                return WaitingStatus("Member is syncing...")
+            elif replica_status == "REMOVED":
+                return WaitingStatus("Member is removing...")
+            else:
+                return BlockedStatus(replica_status)
```
Gu1nness: **Nit**: That's a very nice occasion of using a pattern matching here:
```suggestion
            match replica_status:
                case "PRIMARY":
                    return ActiveStatus("Primary")
                case "SECONDARY":
                    return ActiveStatus("")
                case  "STARTUP" | "STARTUP2" | "ROLLBACK" | "RECOVERING":
                    return WaitingStatus("Member is syncing...")
                case "REMOVED":
                    return WaitingStatus("Member is removing...")
                case _:
                    return BlockedStatus(replica_status)
```

tests/unit/test_charm.py
Gu1nness, line 1: Maybe we could also have a test here for the `unit_host` method (which is the former `unit_ip`  function)?

Gu1nness (dismissed):

lib/charms/mongodb/v0/set_status.py
```diff
@@ -150,4 +161,99 @@ def is_unit_status_ready_for_upgrade(self) -> bool:
...
+            statuses = self.get_statuses()
+        except OperationFailure as e:
+            if e.code in [UNAUTHORISED_CODE, AUTH_FAILED_CODE]:
+                waiting_status = f"Waiting to sync passwords across the {deployment_mode}"
+            elif e.code == TLS_CANNOT_FIND_PRIMARY:
+                waiting_status = (
+                    f"Waiting to sync internal membership across the {deployment_mode}"
+                )
+            else:
+                raise
```
Gu1nness, line 188: Another suggestion which will make this code easier to update in the future if we have to handle more status codes:
```suggestion
            match e.code:
                case UNAUTHORISED_CODE | AUTH_FAILED_CODE:
                    waiting_status = f"Waiting to sync passwords across the {deployment_mode}"
                case TLS_CANNOT_FIND_PRIMARY:
                    waiting_status = (
                        f"Waiting to sync internal membership across the {deployment_mode}"
                    )
                case _:
                    raise
```

Gu1nness (dismissed):

Gu1nness (dismissed):

lib/charms/mongodb/v0/set_status.py
```diff
@@ -150,4 +161,99 @@ def is_unit_status_ready_for_upgrade(self) -> bool:
...
+        except ServerSelectionTimeoutError:
+            waiting_status = f"Waiting to sync internal membership across the {deployment_mode}"
+
+        if waiting_status:
+            return WaitingStatus(waiting_status)
+
+        return self.prioritize_statuses(statuses)
+
+    def get_statuses(self) -> Tuple:
```
Mehdi-Bendriss, line 197: This is a good candidate for a `SimpleNamespace` so you don't have to consider the order for unpacking

Mehdi-Bendriss (approved):

Gu1nness (approved):

PR #440, opened 2024-07-25
Move to DP workflows and self-hosted runners

Mehdi-Bendriss:
## Issue
This PR: 
- moves our CI pipelines to using the DP workflows 
    - moves to poetry
    - enables allure reports 
- makes integ. tests run on self-hosted runners
- bumps juju version to `3.4.4` on CI
- updates the dp libs on client integration tests


tests/integration/upgrade/test_sharding_rollback.py
```diff
@@ -103,7 +104,7 @@ async def test_rollback_on_config_server(
     # TODO implement this check once we have implemented the post-cluster-upgrade code DPE-4143
 
 
-@pytest.mark.skip("re-enable these tests once upgrades are available on charmhub")
```
MiaAltieri, line 106: the tests in this suite and the tests in `tests/integration/upgrade/test_sharding_upgrade.py` still do not use the charm from charmhub and are effectively not testing the upgrade.

Now that CI is functioning and we have published a new revision to charmhub, we should update these tests to use the charm from charmhub

MiaAltieri: Since we rollback to the same charm (and a charm without any changes in the workload) there was no rolling upgrade and so `resume-upgrade` fails in the test since there is no upgrade to resume. Hence the error in the logs:
```
unit-config-server-0: 22:17:46 DEBUG unit.config-server/0.juju-log Emitting Juju event resume_upgrade_action.
unit-config-server-0: 22:17:46 DEBUG unit.config-server/0.juju-log self._app_workload_container_version='118' self._unit_workload_container_versions=***'config-server/2': '118', 'config-server/1': '118', 'config-server/0': '118'***
unit-config-server-0: 22:17:46 DEBUG unit.config-server/0.juju-log Resume upgrade event failed: No upgrade in progress
```

Please disable this test and make a separate GitHub Issue / JIRA issue and I can try to get to it this week

MiaAltieri (changes requested):
looks good added context

tests/integration/dummy_legacy_app/requirements.txt
Gu1nness, line 1: Why pinning this exact version?
Wouldn't be `ops~=2.15.0` be better so it accepts the patches updates?

Gu1nness (dismissed):

MiaAltieri (changes requested):
Mind updating the CONTRIBUTING file to reflect these changes :)

tests/integration/dummy_legacy_app/requirements.txt
Mehdi-Bendriss: good point, thanks!

Gu1nness (dismissed):

MiaAltieri (approved):

Gu1nness (approved):

PR #442, opened 2024-07-29
[DPE-4941]: Get rid of legacy relations

Gu1nness:
## Issue

 * We should remove the old legacy relations code

## Solution
 * Remove the old legacy code.


Gu1nness:
Should be merged after #440 so internal versions are good

delgod (dismissed):

src/charm.py
```diff
@@ -398,7 +391,7 @@ def _on_install(self, event: InstallEvent) -> None:
 
         # Construct the mongod startup commandline args for systemd and reload the daemon.
         update_mongod_service(
-            auth=auth,
+            auth=True,
```
MiaAltieri: IIRC the only scenario in which we started `mongod` service without auth, was when we were using legacy relations.

Would you mind updating this function to not using this parameter anymore?

tests/integration/backup_tests/test_backups.py
```diff
@@ -96,7 +96,7 @@ async def test_blocked_incorrect_conf(ops_test: OpsTest, github_secrets) -> None
     await helpers.set_credentials(ops_test, github_secrets, cloud="AWS")
 
     # wait for both applications to be idle with the correct statuses
-    ops_test.model.wait_for_idle(apps=[S3_APP_NAME], status="active")
+    await ops_test.model.wait_for_idle(apps=[S3_APP_NAME], status="active")
```
MiaAltieri: ugh ty

MiaAltieri (changes requested):
Great work here Neha! One suggestion in one of my comments and a few more here.

Looks like there are still a few remaining references to legacy/obsolete, mind taking care of those?
```
grep -rn legacy lib/charms/mongodb/v1/mongodb_provider.py
grep -rn legacy CONTRIBUTING.md
grep -rni obsolete src/config.py 
```

src/charm.py
Gu1nness: Done!

Mehdi-Bendriss (dismissed):

Mehdi-Bendriss (approved):

MiaAltieri (approved):

PR #443, opened 2024-07-31
[DPE-4997] fix backup tests

Gu1nness:
## Issue

Backups tests are flaky/failing.

## Solution
 * Better  `cluster_destroy`
 * Increase some dangerously low timeouts (just on `update-status` interval is too low)
 * Also remove deprecation warnings


tests/integration/backup_tests/test_backups.py
```diff
@@ -54,7 +59,7 @@ async def test_build_and_deploy(ops_test: OpsTest) -> None:
         await ops_test.model.deploy(db_charm, num_units=3)
 
     # deploy the s3 integrator charm
-    await ops_test.model.deploy(S3_APP_NAME, channel="edge", revision=17)
+    await ops_test.model.deploy(S3_APP_NAME, channel="edge", revision=24)
```
MiaAltieri: lets remove revision, and ensure it works on latest in our tests 🙏 

sorry, I know it wasn't you who pinned the revision in the first place, it was me 😬

tests/integration/helpers.py
```diff
@@ -75,6 +75,25 @@ def dump(self) -> Dict[str, Any]:
         return result
 
 
+async def destroy_cluster(ops_test: OpsTest, applications: list[str]) -> None:
+    """Destroy cluster in a forceful way."""
+    for app in applications:
+        await ops_test.model.applications[app].destroy(
+            destroy_storage=True, force=True, no_wait=False
```
MiaAltieri, line 82: Does it break with `no_wait=True`? I think it might be nice to add so that the process speeds up, especially since we don't care about it being done safely. [Docs link](https://pythonlibjuju.readthedocs.io/en/latest/api/juju.application.html#juju.application.Application.destroy)

tests/integration/helpers.py
```diff
@@ -75,6 +75,25 @@ def dump(self) -> Dict[str, Any]:
...
+
+    # destroy does not wait for applications to be removed, perform this check manually
+    for attempt in Retrying(stop=stop_after_attempt(100), wait=wait_fixed(10), reraise=True):
+        with attempt:
+            # pytest_operator has a bug where the number of applications does not get correctly
+            # updated. Wrapping the call with `fast_forward` resolves this
+            async with ops_test.fast_forward():
+                finished = all((item not in ops_test.model.applications for item in applications))
+            # This case we don't raise an error in the context manager which
+            # fails to restore the `update-status-hook-interval` value to it's former state.
```
MiaAltieri: nit: first line should go until char 99 and then wrap to the bottom

tests/integration/backup_tests/test_backups.py
```diff
@@ -84,7 +89,7 @@ async def test_blocked_incorrect_creds(ops_test: OpsTest) -> None:
     await ops_test.model.wait_for_idle(apps=[S3_APP_NAME], status="active")
 
     await wait_for_mongodb_units_blocked(
-        ops_test, db_app_name, status="s3 credentials are incorrect.", timeout=300
+        ops_test, db_app_name, status="s3 credentials are incorrect.", timeout=600
```
MiaAltieri: hmmm, I understand that a longer time out is needed, but also 10minutes is a long time and our tests are already taking a long time to run. 

Can we manually configure juju to have a shorter update status for these tests? This way they go into their expected states more quickly?

I am not sure `ops_test.fast_forward` will work since we are using subprocess to get the juju status. Maybe adding this in `wait_for_mongodb_units_blocked `:
```
await ops_test.model.set_config({"update-status-hook-interval": "15s"})
# check status
await ops_test.model.set_config({"update-status-hook-interval": "5m"})
```

tests/integration/sharding_tests/helpers.py
```diff
@@ -154,25 +153,6 @@ async def deploy_cluster_components(
     )
 
 
-async def destroy_cluster(ops_test):
```
MiaAltieri, line 157: thanks for cleaning this up

MiaAltieri (changes requested):
Thank you very much for your work here. 

We must be sure these are copied over to the K8s charm once they are approved. Right now I added these tests but they are on the previous version.

tests/integration/helpers.py
Gu1nness: I'll give it a try locally

tests/integration/backup_tests/test_backups.py
Gu1nness: Sure I can do that.
It's exactly what fast_forward does so I'll add itin the `wait_for_mongodb_units_blocked`

MiaAltieri (dismissed):

tests/integration/helpers.py
```diff
@@ -436,9 +455,15 @@ async def wait_for_mongodb_units_blocked(
 
     This is necessary because the MongoDB app can report a different status than the units.
     """
-    for attempt in Retrying(stop=stop_after_delay(timeout), wait=wait_fixed(1), reraise=True):
-        with attempt:
-            await check_all_units_blocked_with_status(ops_test, db_app_name, status)
+    hook_interval_key = "update-status-hook-interval"
+    try:
+        old_interval = (await ops_test.model.get_config())[hook_interval_key]
+        await ops_test.model.set_config({hook_interval_key: "1m"})
```
MiaAltieri, line 461: QQ I assume 1m is the shortest period of time. Is that correct?

MiaAltieri (commented):
Thanks for making all of these changes ! looks great, one question

tests/integration/helpers.py
Gu1nness: It could be lower (eg `15s`) but I noticed that it sometimes would leave units in state `executing` forever because it would get overwhelmed by `update status` events so I rather went for a safe in-between

tests/integration/helpers.py
```diff
@@ -75,6 +75,25 @@ def dump(self) -> Dict[str, Any]:
         return result
 
 
+async def destroy_cluster(ops_test: OpsTest, applications: list[str]) -> None:
+    """Destroy cluster in a forceful way."""
+    for app in applications:
+        await ops_test.model.applications[app].destroy(
```
Mehdi-Bendriss, line 81: The approach is varying from the previous one where units were deleted one by one, until finally deleting the app. What did we gain / lose by changing approaches?

If the flakiness of the test was partly attributed to the unit-by-unit removal, this may be a bug on scale-down we need to figure out.

tests/integration/helpers.py
Mehdi-Bendriss: hmm unless the old interval is lower than 1m, this looks like a hack 
@Gu1nness What causes the avalanche of update-status events? I would rather we investigate that and fix the root cause of wrongly triggered events

Mehdi-Bendriss (commented):
Thanks Neha - I have a couple of questions

tests/integration/helpers.py
Gu1nness: The flakiness doesn't come from this, I just unified the two different `destroy_cluster` that we had in the codebase to the one that is the fastest and keeps its safety. The version where we were removing units one by one was extra careful and the other solutions showed us it had no impact so I decided to go with this one after a chat with @miaaltieri

Gu1nness:
> hmm unless the old interval is lower than 1m, this looks like a hack
@Gu1nness What causes the avalanche of update-status events? I would rather we investigate that and fix the root cause of wrongly triggered events 

The interval is `5m` before (regular update status hook interval) and instead of increasing the timeout of the tests (which are already long tests) that are flaky because it would leave time for only one update event, I rather lower it down to `1m` locally and restore it afterwards. What I was explaining to Mia was that lowering it to 15s was sometimes creating an indefinite loop of update-status-hook somehow and wasn't good.

I'm just doing a safe version of `ops_test.fast_forward` here since the library version isn't safe and fails to restore the `update-status-hook-interval` if an error is raised which can be the case here with the assertion.

Mehdi-Bendriss (approved):

MiaAltieri (approved):

PR #450, opened 2024-08-05
[DPE-5003][DPE-4999] Upgrade tests use most recent charm

Gu1nness:
## Issue

 * Update the upgrade tests to start using  `6/edge`
 * Add resume-upgrade handling
 * Upgrade tests are flaky
 * Bug discovered : new version is not shared if there is no snap revision change

## Solution

 * Bug fix : always share the revision change in `_on_update_charm` handler
 * Add resume-upgrade handling in all tests
 * Fetch remote charms and upgrade to locally built charm

## Miscellaneous
 
 * Opened a bug in [libjuju](https://github.com/juju/python-libjuju/issues/1086) and add a workaround meanwhile


src/upgrades/mongodb_upgrade.py
```diff
@@ -189,6 +190,7 @@ def _on_force_upgrade_action(self, event: ActionEvent) -> None:
         logger.debug("Forcing upgrade")
         event.log(f"Forcefully upgrading {self.charm.unit.name}")
         self._upgrade.upgrade_unit(charm=self.charm)
+        self.charm.version_checker.set_version_across_all_relations()
```
MiaAltieri: QQ: Is this needed here?

I don't believe it is, but I could very well be wrong. 

Reasoning: `_on_pre_upgrade_check_action` actions in two cases:
1. before the user runs `juju refresh ...`
2. when the user runs `juju refresh ...`

In case 1, no version has changed so its not necessary share across all relations
In case 2, I believe your change on line `135` is sufficent

tests/integration/sharding_tests/helpers.py
```diff
@@ -13,9 +13,10 @@
     get_secret_data,
 )
 
-TIMEOUT = 10 * 60
+TIMEOUT = 15 * 60
```
MiaAltieri, line 16: QQ my memory fails me. Were you ever able to find the reason for the application taking awhile to deploy in the tests?

tests/integration/sharding_tests/helpers.py
```diff
@@ -117,7 +118,7 @@ def count_users(mongos_client: MongoClient) -> int:
 
 
 async def deploy_cluster_components(
-    ops_test: OpsTest, num_units_cluster_config: Dict = None
+    ops_test: OpsTest, num_units_cluster_config: dict | None = None, channel: str | None = None
```
MiaAltieri, line 121: nice 😎

tests/integration/upgrade/test_sharding_rollback.py
```diff
@@ -73,9 +71,11 @@ async def test_rollback_on_config_server(
     )
 
     # instead of resuming upgrade refresh with the old version
-    # TODO: instead of using new_charm - use the one deployed on charmhub - cannot do this until
-    # the newest revision is published
-    await ops_test.model.applications[CONFIG_SERVER_APP_NAME].refresh(path=new_charm)
+    # TODO: Use this when https://github.com/juju/python-libjuju/issues/1086 is fixed
```
MiaAltieri, line 74: thanks for the investigation here <3

tests/integration/upgrade/test_upgrade.py
```diff
@@ -55,6 +53,8 @@ async def test_upgrade(ops_test: OpsTest, continuous_writes) -> None:
     action = await leader_unit.run_action("pre-upgrade-check")
     await action.wait()
 
+    assert action.status == "completed", "pre-upgrade-check-failed, expected to succeed"
```
MiaAltieri, line 56: ty 🙇

src/upgrades/mongodb_upgrade.py
Gu1nness: I'm guessing you mean `_on_force_upgrade_action` but in any case.
I'm assuming we have to run `force-upgrade` if anything went wrong in the upgrade, including if it failed before running the `set_version_across_all_relations` in `_on_upgrade_charm`.
I might be wrong but it seems reasonable to me.
Maybe we should also have tests for force upgrade action?

tests/integration/sharding_tests/helpers.py
Gu1nness: From what I've seen, we were on the verge of timing out when deploying full cluster with local images.
I guess using remote images pushed us beyond timeout.
There is a small overhead using the self hosted runners it seems which comes on top of that.
I'm not happy at all about this increased timeout but there's nothing much we can do for now…

src/upgrades/mongodb_upgrade.py
MiaAltieri: > I'm guessing you mean _on_force_upgrade_action but in any case.

Oops, yes you are right, I was too speedy in my review thank you for catching this. 

Some context is that `_on_force_upgrade_action ` can be run on *any* unit (i.e. non-leader units) and `set_version_across_all_relations ` **must** be ran on the leader unit. So I don't think this is the right place for it.

What we can do instead, is change our `def _on_upgrade_charm(self, _):` function to switch up our order of operations. namely modify it to:
```
            self.charm.version_checker.set_version_across_all_relations()
            self._reconcile_upgrade()
```
this way we set the version before actually begging to reconcile. This way errors in `_reconcile_upgrade` will not block `set_version_across_all_relations`.

Some Juju knowledge for you:
If there is a python error during an event handler, that unit goes into the `error` state, and will keep calling that hook until it resolves itself. So we can trust once the charm goes out of `error` state that `set_version_across_all_relations` will eventually be called.

Now for your second point:
> Maybe we should also have tests for force upgrade action?

If you'd like to implement this, I would be happy to support you on it, but we would have to make a separate JIRA ticket. I imagine the timing of simulating a failure in a timely matter (such that pre-upgrade check succeeds, but post-upgrade check fails) would be challenging.

tests/integration/sharding_tests/helpers.py
MiaAltieri: QQ Did you ask Carl about it? They are an expert when it comes to caching images on CI

src/upgrades/mongodb_upgrade.py
Gu1nness: Ok perfect, I'll do this then.
For the additional test it needs to be an extra ticket but it would be nice if it helps having a framework to test for such errors.

tests/integration/sharding_tests/helpers.py
Gu1nness: We talked about it and as it seems there's no big difference between GH runners and self hosted runners, it might just be that we really take too much time because we spend already 5-6 minutes waiting for machines to be available when deploy the cluster

MiaAltieri: ack ty

MiaAltieri (approved):

Mehdi-Bendriss (approved):

PR #451, opened 2024-08-06
chore: fix juju version requirements

Gu1nness:
## Issue

* Wrong requirement specification for juju.

## Solution

Fix juju requirements in `pyproject.toml`.

More precisely for the future ([full spec here for poetry](https://python-poetry.org/docs/dependency-specification/)):
 * `juju = "^3.4.0"` means `">=3.4.0, <4.0.0"` which is: every new version until next major
 * `juju = "~3.4.0"` means `">=3.4.0, <3.5.0"` which is: every patch until next minor


MiaAltieri (approved):
Thank you Neha! TIL `~` thanks for knowledge sharing

Mehdi-Bendriss (approved):

PR #454, opened 2024-08-09
fix: relations tests

Gu1nness:
## Issue

* Some tests are still flaky
* Libs are not up to date

## Solution

* Bump cos-agent lib
* Fix flakiness on backup tests: reusing the same name + destroy/recreate is not really stable so use different names
* Fix flakiness on relations test: we sometimes call the `get_connection_string` before the secret is created, so retry here to get it as soon as it's ready


MiaAltieri (dismissed):
thank you Neha 🙇‍♀️

delgod (dismissed):

MiaAltieri (approved):

Mehdi-Bendriss (approved):

PR #456, opened 2024-08-13
[DPE-5040] - make necessary lib changes for mongos-k8s connections with config-server

MiaAltieri:
## Issue
The k8s router has a slightly different implementation from the VM router. This means that the current lib that supports the router interface does not work with K8s

Namely the following has a different implementation on the K8s router:
1. sharing user info to client
2. k8s router communicates via port
3. k8s router updates arguments when changing the pebble layer on restart

## Solution
To fix this some cases have been made in the charm lib
1. k8s charm does not provide its admin user/password to the client (instead it will handle this with the mongos_client interface in DPE-5090
2. communicate via port in K8s
3. dont make a call to update mongos args in K8s


src/config.py
```diff
@@ -15,7 +15,8 @@ class Config:
 
     MONGOS_PORT = 27018
     MONGODB_PORT = 27017
-    SUBSTRATE = "vm"
+    VM_SUBSTRATE = "vm"
+    K8S_SUBSTRATE = "k8s"
```
Gu1nness: I'm not a huge fan of bringing `K8S_SUBSTRATE` and `VM_SUBSTRATE` like that in the config as it brings another layer of "knowledge" about that in the charm code that could more easily bring havoc in the charm (eg: "What if I start checking about my substrate in my own charm").

Maybe adding those two possibilities in the library instead could help event it out: the library has the control and all the knowledge about the substrate given what the charm provides, which makes it more easily usable: Here, this library update brings a change in the charm config which will have to be replicated in the k8s charm and forever both charms would hold this information about the other substrate.

So my suggestion would be:
 * Keep `SUBSTRATE` here as it is
 * Do an enum, or a tiny class in the library for those two values like so:
```python3
class Substrate:
   VM = "vm"
   K8S = "k8s"
```
In a perfect world we would have StrEnum but we're not in python 3.11 yet so we'll have to do with it.

Mehdi-Bendriss: As a side note, we can have string enums in Python 3.10 by doing the following, but then it must always be cast to str:
```
class Substrate(str, Enum):
    VM = "vm"
    K8s = "k8s"

print(f"{Substrate.VM}")
```

Mehdi-Bendriss (dismissed):

Mehdi-Bendriss (approved):

Gu1nness (approved):

PR #459, opened 2024-08-16
[DPE-5090] merge parts of mongo / mongodb libs

MiaAltieri:
## Issue
For [DPE-5090](https://warthogs.atlassian.net/browse/DPE-5090) we need the ability for mongos to create users

copying over the create user code from mongodb.py to mongos.py is a bad scheme

## Solution
Create a shared class for shared code betweenmongodb.py to mongos.py 

[DPE-5090]: https://warthogs.atlassian.net/browse/DPE-5090?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lib/charms/mongodb/v0/mongo.py
```diff
@@ -0,0 +1,233 @@
...
+    when using MongoDBConnection:
+
+    with MongoMongos(MongoConfig) as mongo:
+        try:
+            mongo.<some operation from this class>
+        except ConfigurationError, OperationFailure:
+            <error handling as needed>
+    """
+
+    def __init__(self, config: MongoConfiguration, uri=None, direct=False):
```
MiaAltieri, line 156: the rest of this was pulled from `mongodb.py`

lib/charms/mongodb/v0/mongo.py
```diff
@@ -0,0 +1,238 @@
...
+                {"db": "admin", "role": "readWrite", "collection": ""},
+                {"db": "admin", "role": "backup"},
+                {"db": "admin", "role": "clusterMonitor"},
+                {"db": "admin", "role": "restore"},
+                {"db": "admin", "role": "pbmAnyAction"},
+            ],
+            "default": [
+                {"role": "readWrite", "db": config.database},
+            ],
+        }
```
Gu1nness: This dictionary can be moved out of the function so it's created once only (a bit faster, but also it would be accessible elsewhere if needed?

lib/charms/mongodb/v0/mongo.py
```diff
@@ -0,0 +1,238 @@
...
+            [
+                user_obj["user"]
+                for user_obj in users_info["users"]
+                if re.match(r"^relation-\d+$", user_obj["user"])
+            ]
+        )
+
+    def get_databases(self) -> Set[str]:
+        """Return list of all non-default databases."""
+        system_dbs = ("admin", "local", "config")
```
Gu1nness: As it's used twice (in this and the next function), this can also be made a constant so if ever this needs to be changed, it's changed only once and reduces the risk of error

lib/charms/mongodb/v1/mongodb.py
```diff
@@ -124,50 +103,9 @@ def __init__(self, config: MongoDBConfiguration, uri=None, direct=False):
             direct: force a direct connection to a specific host, avoiding
                     reading replica set configuration and reconnection.
         """
+        MongoConnection.__init__(self, config, uri, direct)
```
Gu1nness: Use `super()` here like so:
```python
super().__init__(self, config, uri, direct)
```
Also, considering that we save the config here, I'm wondering if there's a reason not to save it directly in the base class?

lib/charms/mongodb/v0/mongo.py
```diff
@@ -0,0 +1,238 @@
...
+    password: str
+    hosts: Set[str]
+    roles: Set[str]
+    tls_external: bool
+    tls_internal: bool
+
+    @property
+    @abc.abstractmethod
+    def uri(self):
+        """Return URI concatenated from fields."""
```
Gu1nness: Even though it's not mandatory, you can add a `raise NotImplentedError` here to ensure that it's implemented in the child classes.
This way you enforce the subclasses to implement this method.

lib/charms/mongodb/v0/mongo.py
```diff
@@ -0,0 +1,238 @@
...
+# Increment this PATCH version before using `charmcraft publish-lib` or reset
+# to 0 if you are raising the major API version
+LIBPATCH = 1
+
+# path to store mongodb ketFile
+logger = logging.getLogger(__name__)
+
+
+@dataclass
+class MongoConfiguration:
```
Gu1nness, line 67: To ensure that this class is not instantiated by itself, what you can do is the following:
```python3
@dataclass 
class AbstractDataclass(ABC): 
    def __new__(cls, *args, **kwargs): 
        if cls == AbstractDataclass or cls.__bases__[0] == AbstractDataclass: 
            raise TypeError("Cannot instantiate abstract class.") 
        return super().__new__(cls)

@dataclass 
class MongoConfiguration(AbstractDataclass):
	...
```

It's not really a must have here as we don't have many developers but it will prevent some errors.

Gu1nness (changes requested):
Overall that's a very good reusable implementation!
I added a few comments to make it even better, some are not necessarily mandatory (the `AbstractDataClass` one for example), and some are more important (regarding the use of super).

lib/charms/mongodb/v0/mongo.py
MiaAltieri: noob question, how would that work with the reference to the input` "db": config.database},`

lib/charms/mongodb/v0/mongo.py
MiaAltieri: 🧠

lib/charms/mongodb/v1/mongodb.py
MiaAltieri: 🧠

lib/charms/mongodb/v0/mongo.py
MiaAltieri: 🧠

lib/charms/mongodb/v0/mongo.py
MiaAltieri: :brain:

lib/charms/mongodb/v0/mongo.py
Gu1nness: Oh damn you're right, did not see this one.
Hum, let me think how we can do this right.
Maybe have the regular roles in the dict and complete it in this function?
Or provide a helper function to build the whole dict.
I would go with:
```python3
REGULAR_ROLES = {
    "admin": [
        {"role": "userAdminAnyDatabase", "db": "admin"},
        {"role": "readWriteAnyDatabase", "db": "admin"},
        {"role": "userAdmin", "db": "admin"},
    ],
    "monitor": [
        {"role": "explainRole", "db": "admin"},
        {"role": "clusterMonitor", "db": "admin"},
        {"role": "read", "db": "local"},
    ],
    "backup": [
        {"db": "admin", "role": "readWrite", "collection": ""},
        {"db": "admin", "role": "backup"},
        {"db": "admin", "role": "clusterMonitor"},
        {"db": "admin", "role": "restore"},
        {"db": "admin", "role": "pbmAnyAction"},
    ],
}


def supported_roles(config: MongoConfig):
    return REGULAR_ROLES | {"default": [{"db": config.database, "role": "readWrite"}]}
```

And then 
```python3
from itertools import chain

@staticmethod
def _get_roles(config: MongoConfiguration) -> List[dict]:
    all_roles = supported_roles(config)
    return list(chain.from_iterable(all_roles[role] for role in config.roles))
```
The use of itertools allows efficient iterations and chaining, this method uses minimal copies and object creations.

lib/charms/mongodb/v0/mongo.py
```diff
@@ -94,6 +108,8 @@ def __init__(self, config: MongoConfiguration, uri=None, direct=False):
             direct: force a direct connection to a specific host, avoiding
                     reading replica set configuration and reconnection.
         """
+        self.config = config
```
Gu1nness, line 165: Now, considering this instantiation, I guess we need to go through the two other libraries to see if everytime we provide a config to a method, it's the same config as this one or not.
If yes, we can remove this parameter and use this class attribute for better perfs, more object stability and smaller memory footprint.

lib/charms/mongodb/v0/mongo.py
```diff
@@ -0,0 +1,252 @@
...
+    - tls_internal: indicator for use of external TLS connection.
+    """
+
+    database: Optional[str]
+    username: str
+    password: str
+    hosts: Set[str]
+    roles: Set[str]
+    tls_external: bool
+    tls_internal: bool
```
Gu1nness: The documentation doesn't match up the defined attributes:
 * `port` is added, maybe this should be defined once and for all here, and overridden in the actual classes?
 * There are two kinds of dashes here and I'm not sure if there's a meaning for that or not, but it's disturbing

lib/charms/mongodb/v1/mongos.py
```diff
@@ -125,26 +109,7 @@ def __init__(self, config: MongosConfiguration, uri=None, direct=False):
...
-
-    def __enter__(self):
-        """Return a reference to the new connection."""
-        return self
-
-    def __exit__(self, object_type, value, traceback):
-        """Disconnect from MongoDB client."""
-        self.client.close()
-        self.client = None
+        super.__init__(config, uri, direct)
```
Gu1nness:
```suggestion
        super().__init__(config, uri, direct)
```

lib/charms/mongodb/v0/mongo.py
MiaAltieri: port actually isn't needed in mongo lib, PyMongo MongoClient always uses 27017 be default. Let me remove and update dashes

lib/charms/mongodb/v0/mongo.py
Gu1nness: We don't always (eg we create a `monitor` user with the regular config) so there's no need for that.

lib/charms/mongodb/v0/mongo.py
MiaAltieri: thank you very much!

Gu1nness (dismissed):

lib/charms/mongodb/v0/mongo.py
```diff
@@ -0,0 +1,284 @@
...
+    port: Optional[str] = Config.MONGODB_PORT
+    replset: Optional[str] = None
+    standalone: bool = False
+
+    @property
+    def uri(self):
```
MiaAltieri, line 91: @Gu1nness I forgot that a big point of this PR was to have a flexible `MongoConfiguration ` that can be used by `mongodb_provider` lib for clients. (A PR using this will be made after)

coverage.xml
Gu1nness, line 1: This should be removed, did the `.gitignore` got updated?

lib/charms/mongodb/v0/mongo.py
Gu1nness: Right :)

.gitignore
```diff
@@ -3,7 +3,7 @@
 
 *.idea
 .vscode/
-.coverage
+*coverage
```
Gu1nness:
```suggestion
.coverage
coverage*
```

lib/charms/mongodb/v0/mongo.py
```diff
@@ -0,0 +1,284 @@
...
+    — replset: name of replica set
+    — database: database name.
+    — username: username.
+    — password: password.
+    — hosts: full list of hosts to connect to, needed for the URI.
+    — tls_external: indicator for use of internal TLS connection.
+    — tls_internal: indicator for use of external TLS connection.
+    """
+
+    database: Optional[str]
```
Gu1nness: nit
```suggestion
    database: str | None
```

lib/charms/mongodb/v0/mongo.py
```diff
@@ -0,0 +1,284 @@
...
+    username: str
+    password: str
+    hosts: Set[str]
+    roles: Set[str]
+    tls_external: bool
+    tls_internal: bool
+    port: Optional[str] = Config.MONGODB_PORT
+    replset: Optional[str] = None
```
Gu1nness: Beware : port is an int not a str + it's always defined because of the default value
```suggestion
    port: int = Config.MONGODB_PORT
    replset: str | None = None
```

lib/charms/mongodb/v0/mongo.py
```diff
@@ -0,0 +1,284 @@
...
+                user_obj["user"]
+                for user_obj in users_info["users"]
+                if re.match(r"^relation-\d+$", user_obj["user"])
+            ]
+        )
+
+    def get_databases(self) -> Set[str]:
+        """Return list of all non-default databases."""
+        databases = self.client.list_database_names()
+        return set([db for db in databases if db not in SYSTEM_DBS])
```
Gu1nness: This is actually way more efficient in two ways:
  * Don't create a temp list (you can avoid this by directly providing the iterator to `set()`, but we're going one step further after that).
  * Calling `set(…)` instead of `{…}` is extremely inefficient (You can have a [read here](https://madebyme.today/blog/python-dict-vs-curly-brackets/) if you want to know more).
```suggestion
        return {db for db in databases if db not in SYSTEM_DBS}
```

Gu1nness (dismissed):
Some very last changes to do (but approved on the idea)

lib/charms/mongodb/v0/mongo.py
```diff
@@ -0,0 +1,254 @@
...
+    @property
+    def is_ready(self) -> bool:
+        """Is the MongoDB server ready for services requests.
+
+        Returns:
+            True if services is ready False otherwise. Retries over a period of 60 seconds times to
+            allow server time to start up.
+
+        Raises:
+            ConfigurationError, ConfigurationError, OperationFailure
```
Mehdi-Bendriss:
```suggestion
```

lib/charms/mongodb/v0/mongo.py
```diff
@@ -0,0 +1,284 @@
...
+        {"db": "admin", "role": "restore"},
+        {"db": "admin", "role": "pbmAnyAction"},
+    ],
+}
+
+# path to store mongodb ketFile
+logger = logging.getLogger(__name__)
+
+
+class AmbigiousConfigError(Exception):
```
Mehdi-Bendriss:
```suggestion
class AmbiguousConfigError(Exception):
```

lib/charms/mongodb/v0/mongo.py
```diff
@@ -0,0 +1,284 @@
...
+    username: str
+    password: str
+    hosts: Set[str]
+    roles: Set[str]
+    tls_external: bool
+    tls_internal: bool
+    port: Optional[str] = Config.MONGODB_PORT
+    replset: Optional[str] = None
+    standalone: bool = False
+
```
Mehdi-Bendriss, line 89: may be worth adding the `@override` decorator.

Mehdi-Bendriss (dismissed):

lib/charms/mongodb/v0/mongo.py
MiaAltieri: Lets see down the road if its needed. Right now the idea is to have them together and I think having `override` goes against that

Gu1nness (approved):

lib/charms/mongodb/v0/mongo.py
Mehdi-Bendriss: I am not entirely sure why has this comment landed here - it was meant to be on the child class.

Mehdi-Bendriss (approved):

PR #460, opened 2024-08-19
[DPE-5004] Fix internal version hassle by moving to hashes

Gu1nness:
## Issue

 * Updating Charm internal version on every new version is really painful and slows down the process

## Solution

 * Automatically generate a consistent hash that will be used as charm internal version
 * Update the data platform helpers to use that


scripts/gen_charm_internal_version.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/sh
```
MiaAltieri, line 1: missing copywrite and sentence describing file

MiaAltieri (commented):
Question:
1. Where is the charm_internal_version file? I thought it would have been PR, but it looks like it was removed
2. When does this file get generated?

src/charm.py
```diff
@@ -361,10 +361,10 @@ def upgrade_in_progress(self):
         return self.upgrade._upgrade.in_progress
 
     @property
-    def get_charm_internal_revision(self) -> int:
+    def get_charm_internal_revision(self) -> str:
         """Returns the contents of the get_charm_internal_revision file."""
         with open(Config.CHARM_INTERNAL_VERSION_FILE, "r") as f:
```
MiaAltieri, line 361: how does this work if the file was deleted

Gu1nness:
>     1. Where is the charm_internal_version file? I thought it would have been PR, but it looks like it was removed
>     2. When does this file get generated?

Answering both at the same time: It's not in the repo anymore, it's generated by the script which is called by the `tox` environments.
This way the file is generated when the charms are built (and when doing the unit tests), but it's not left in the repository (which was the pain of having to bump it everytime)

scripts/gen_charm_internal_version.sh
```diff
@@ -0,0 +1,23 @@
...
+
+if git --version > /dev/null; then
+    # Compute base files
+    BASE=$(git ls-files -s)
+    # Compute diff files
+    DIFF=$(git diff --raw)
+    # Compute staged files
+    STAGED=$(git diff --raw --staged)
+
+    HASH=$(echo $BASE $DIFF $STAGED | git hash-object --stdin | cut -c 1-8)
```
Mehdi-Bendriss, line 17: why is the `cut` necessary here? why not store the whole string?

Gu1nness: We discussed about it with @MiaAltieri and to avoid too long strings/messages in statuses, we decided to go to a shorter version.
It's not really important in itself, just makes things easier to read and understand
It still keeps something like 4B possibilities so there's a long way before we have a collision (and it's what git does too to generate the short id of a commit)

MiaAltieri (approved):
Thanks for the great work on this! You are excellent!

Mehdi-Bendriss (approved):
Thanks for taking initiative!

src/charm.py
Gu1nness: The workflow is the following:
```
/bin/sh gen_charm_internal_version.sh # Creates the file
tox -e build-productoin
rm charm_internal_version
```
so when we pack the charm, the file is there and ends up in the packed charm and then it's deleted to leave this repo clean.

PR #461, opened 2024-08-19
[DPE-5176] Substrate agnostic upgrade helpers in their library

Gu1nness:
* Create a substrate agnostic upgrade helpers library so it can be reused with mongodb-k8s when we implement upgrades there
* Will be updated after #459 is merged to resolve dependencies issues

### DIFF
--- a/lib/charms/mongodb/v0/upgrade_helpers.py
+++ b/lib/charms/mongodb/v0/upgrade_helpers.py
@@ -8,8 +8,9 @@ import secrets
 import string
 from typing import List, Tuple
 
-from charms.mongodb.v1.mongodb import MongoDBConfiguration, MongoDBConnection
-from charms.mongodb.v1.mongos import MongosConfiguration, MongosConnection
+from charms.mongodb.v0.mongo import MongoConfiguration
+from charms.mongodb.v1.mongodb import MongoDBConnection
+from charms.mongodb.v1.mongos import MongosConnection
 from ops.charm import CharmBase
 from ops.framework import Object
 from pymongo.errors import OperationFailure, PyMongoError, ServerSelectionTimeoutError
@@ -169,7 +170,7 @@ class GenericMongoDBUpgrade(Object, metaclass=abc.ABCMeta):
 
         return True
 
-    def are_replicas_in_sharded_cluster_healthy(self, mongos_config: MongosConfiguration) -> bool:
+    def are_replicas_in_sharded_cluster_healthy(self, mongos_config: MongoConfiguration) -> bool:
         """Returns True if all replicas in the sharded cluster are healthy."""
         # dictionary of all replica sets in the sharded cluster
         for mongodb_config in self.get_all_replica_set_configs_in_cluster():
@@ -179,7 +180,7 @@ class GenericMongoDBUpgrade(Object, metaclass=abc.ABCMeta):
 
         return True
 
-    def are_shards_healthy(self, mongos_config: MongosConfiguration) -> bool:
+    def are_shards_healthy(self, mongos_config: MongoConfiguration) -> bool:
         """Returns True if all shards in the cluster are healthy."""
         with MongosConnection(mongos_config) as mongos:
             if mongos.is_any_draining():
@@ -202,7 +203,7 @@ class GenericMongoDBUpgrade(Object, metaclass=abc.ABCMeta):
 
         return True
 
-    def get_all_replica_set_configs_in_cluster(self) -> List[MongoDBConfiguration]:
+    def get_all_replica_set_configs_in_cluster(self) -> List[MongoConfiguration]:
         """Returns a list of all the mongodb_configurations for each application in the cluster."""
         mongos_config = self.get_cluster_mongos()
         mongodb_configurations = []
@@ -224,7 +225,7 @@ class GenericMongoDBUpgrade(Object, metaclass=abc.ABCMeta):
 
         return mongodb_configurations
 
-    def are_replica_set_nodes_healthy(self, mongodb_config: MongoDBConfiguration) -> bool:
+    def are_replica_set_nodes_healthy(self, mongodb_config: MongoConfiguration) -> bool:
         """Returns true if all nodes in the MongoDB replica set are healthy."""
         with MongoDBConnection(mongodb_config) as mongod:
             rs_status = mongod.get_replset_status()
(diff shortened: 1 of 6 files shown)


lib/charms/mongodb/v0/upgrade_helpers.py
```diff
@@ -0,0 +1,473 @@
...
+
+# END: Exceptions
+
+
+class _PostUpgradeCheckMongoDB(EventBase):
+    """Run post upgrade check on MongoDB to verify that the cluster is healhty."""
+
+    def __init__(self, handle):
+        super().__init__(handle)
+
```
MiaAltieri: this might be specific to v2 implementation, not sure if it belongs here

lib/charms/mongodb/v0/upgrade_helpers.py
```diff
@@ -0,0 +1,473 @@
...
+
+class GenericMongoDBUpgrade(Object, metaclass=abc.ABCMeta):
+    """Substrate agnostif, abstract handler for upgrade events."""
+
+    post_app_upgrade_event = EventSource(_PostUpgradeCheckMongoDB)
+    post_cluster_upgrade_event = EventSource(_PostUpgradeCheckMongoDB)
```
MiaAltieri: not sure if this is needed or not.

Post upgrade events we had to make manually, we should see if they can be added in the v3 implementation as a first class citizen so that we don't have to add it here

lib/charms/mongodb/v0/upgrade_helpers.py
```diff
@@ -0,0 +1,473 @@
...
+
+    @abc.abstractmethod
+    def _observe_events(self, charm: CharmBase) -> None:
+        """Handler that should register all event observers."""
+        raise NotImplementedError()
+
+    @property
+    @abc.abstractmethod
+    def _upgrade(self) -> AbstractUpgrade | None:
+        raise NotImplementedError()
```
MiaAltieri, line 80: this should become a part of the v3 upgrade implementation

lib/charms/mongodb/v0/upgrade_helpers.py
```diff
@@ -0,0 +1,473 @@
...
+
+    # BEGIN: Helpers
+    def move_primary_to_last_upgrade_unit(self) -> None:
```
MiaAltieri, line 83: are these helpers copy paste?

pyproject.toml
```diff
@@ -88,6 +88,9 @@ minversion = "6.0"
 log_cli_level = "INFO"
 markers = ["unstable"]
 asyncio_mode = "auto"
+filterwarnings = [
```
MiaAltieri, line 92: noob q - what does this do?

MiaAltieri (commented):
Looks great! just a few questions, thanks Neha!

lib/charms/mongodb/v0/upgrade_helpers.py
Gu1nness: Ok, moving it

lib/charms/mongodb/v0/upgrade_helpers.py
Gu1nness: Ok, moving it as well

lib/charms/mongodb/v0/upgrade_helpers.py
Gu1nness: Yes all this was copy pasted

pyproject.toml
Gu1nness: This will transform any runtime warnings of the tests in errors.
`RuntimeWarning` are warnings about dubious runtime errors such as "coroutine was never awaited".
This will help cleaning the tests and avoid some errors we have were we forget an `await`

lib/charms/mongodb/v0/upgrade_helpers.py
Gu1nness: Don't know how this will be handled in the end but yeah.
The idea here is that each charm provides the implementation for this method returning the upgrade `Object`

charm_internal_version
```diff
@@ -1 +1 @@
-18
+19
```
Mehdi-Bendriss: no longer needed, right?

lib/charms/mongodb/v0/upgrade_helpers.py
```diff
@@ -0,0 +1,464 @@
...
+
+
+class BalancerStillRunningError(Exception):
+    """Raised when the balancer is still running after stopping it."""
+
+
+# END: Exceptions
+
+
+class GenericMongoDBUpgrade(Object, metaclass=abc.ABCMeta):
```
Mehdi-Bendriss: Any reason to not use `abc.ABC` instead? ABC implicitely defines the metaclass

src/upgrades/mongodb_upgrade.py
```diff
@@ -59,6 +38,8 @@ class MongoDBUpgrade(Object):
     def __init__(self, charm: CharmBase):
         self.charm = charm
         super().__init__(charm, upgrade.PEER_RELATION_ENDPOINT_NAME)
+
+    def _observe_events(self, charm: CharmBase) -> None:
```
Mehdi-Bendriss, line 44: Could you add the `@override` decorator in all overriden methods and properties?

lib/charms/mongodb/v0/upgrade_helpers.py
```diff
@@ -0,0 +1,464 @@
...
+
+logger = logging.getLogger(__name__)
+
+SHARD_NAME_INDEX = "_id"
+WRITE_KEY = "write_value"
+ROLLBACK_INSTRUCTIONS = "To rollback, `juju refresh` to the previous revision"
+
+
+# BEGIN: Useful classes
+class AbstractUpgrade(abc.ABC):
```
Mehdi-Bendriss, line 40: What would this base class contain?

src/upgrades/mongodb_upgrade.py
```diff
@@ -78,6 +59,16 @@ def __init__(self, charm: CharmBase):
         self.framework.observe(self.post_app_upgrade_event, self.run_post_app_upgrade_task)
         self.framework.observe(self.post_cluster_upgrade_event, self.run_post_cluster_upgrade_task)
 
+    # BEGIN: properties
+    @property
+    def _upgrade(self) -> Optional[machine_upgrade.Upgrade]:
+        try:
+            return machine_upgrade.Upgrade(self.charm)
+        except upgrade.PeerRelationNotReady:
+            pass
```
Mehdi-Bendriss: nit:
```suggestion
            return None
```

Mehdi-Bendriss (commented):
Thanks Neha - I have minor comments

charm_internal_version
Gu1nness: The file has been removed, maybe it was while you were doing your review, but I rebased on `6/edge` to fix that exact issue

lib/charms/mongodb/v0/upgrade_helpers.py
Gu1nness: I thought it would be an issue because this could require that `Object` was also an abstract class but after a test, it doesn't seem so, so let's go for it.
Updated in my next commit

src/upgrades/mongodb_upgrade.py
Gu1nness: Fixed

lib/charms/mongodb/v0/upgrade_helpers.py
Gu1nness: This class here is mainly for typing.
But it could also be used for the same use as `GenericMongoDBUpgrade` to embed the methods that we want to share between the VM and k8s charm.
For now it's an empty class for typing, later it can be extended when we have new upgrades and if there are some methods and patterns that can be shared.

lib/charms/mongodb/v0/upgrade_helpers.py
```diff
@@ -0,0 +1,464 @@
...
+        self._observe_events(charm)
+
+    @abc.abstractmethod
+    def _observe_events(self, charm: CharmBase) -> None:
+        """Handler that should register all event observers."""
+        raise NotImplementedError()
+
+    @property
+    @abc.abstractmethod
+    def _upgrade(self) -> AbstractUpgrade | None:
```
MiaAltieri, line 79: still not 100% convinced this will work with the new upgrade lib. But after our 1:1 call I know if it isn'y you're up for changing it / modifying it as needed

Thank you for the call + clarifications

MiaAltieri (approved):

Mehdi-Bendriss (approved):

PR #462, opened 2024-08-20
[DPE-5090] - Update mongodb provider to support K8s router

MiaAltieri:
## Issue
mongos k8s router cannot yet make use of the mongodb_provider lib due to the lib not be flexible to support mongos

## Solution
Update the lib so that it can support mongos


lib/charms/mongodb/v1/mongodb_provider.py
```diff
@@ -381,11 +395,16 @@ def _get_relations(self, rel=REL_NAME) -> List[Relation]:
         We create users for either direct relations to charm or for relations through the mongos
         charm.
         """
-        return (
-            self.model.relations[MONGOS_RELATIONS]
-            if self.charm.is_role(Config.Role.CONFIG_SERVER)
-            else self.model.relations[rel]
-        )
+        return self.model.relations[self.get_relation_name()]
```
Gu1nness, line 388: You're not using the `rel` argument anymore, so (because it's a private method) you can remove the argument `rel` from the function definition right?

src/charm.py
```diff
@@ -257,6 +257,11 @@ def remote_mongodb_config(self, hosts, replset=None, standalone=None) -> MongoCo
             OperatorUser, hosts, replset=replset, standalone=standalone
         )
 
+    @property
+    def mongo_config(self) -> MongoConfiguration:
+        """Returns a MongoConfiguration object for shared libs with agnoistic mongo commands."""
```
Gu1nness:
```suggestion
        """Returns a MongoConfiguration object for shared libs with agnostic mongo commands."""
```

lib/charms/mongodb/v1/mongodb_provider.py
Gu1nness, line 1: Don't forget to bump the libpatch here

Gu1nness (changes requested):

lib/charms/mongodb/v1/mongodb_provider.py
MiaAltieri: I'll do you even one better, lets remove this function and use `self.relation_name` properly

Gu1nness (dismissed):

Gu1nness (dismissed):

lib/charms/mongodb/v1/mongodb_provider.py
```diff
@@ -50,32 +51,27 @@
 class MongoDBProvider(Object):
     """In this class, we manage client database relations."""
 
-    def __init__(self, charm: CharmBase, substrate="k8s", relation_name: str = REL_NAME) -> None:
+    def __init__(self, charm: CharmBase, substrate="k8s", relation_name: str = "database") -> None:
```
Gu1nness, line 54: Is there a reason to introduce a magic string here instead of using the constant associated with it?

Gu1nness (approved):

lib/charms/mongodb/v1/mongodb_provider.py
MiaAltieri: good, eyes, I can do

Mehdi-Bendriss (approved):

lib/charms/mongodb/v1/mongodb_provider.py
MiaAltieri: [addressed in this pr](https://github.com/canonical/mongodb-operator/pull/466/files)

PR #463, opened 2024-08-21
chore: re-enable tests

Gu1nness:
## Issue

* Some upgrade tests were disabled in order to fix upgrade bugs + move to hash based charm internal version

## Solution

* re enable it


MiaAltieri (dismissed):

Mehdi-Bendriss (dismissed):

tests/integration/upgrade/test_sharding_rollback.py
```diff
@@ -102,36 +101,54 @@ async def test_rollback_on_config_server(
...
+
+    # Wait for statuses to settle down
+    asyncio.gather(
+        wait_for_mongodb_units_blocked(ops_test, SHARD_ONE_APP_NAME),
+        wait_for_mongodb_units_blocked(ops_test, SHARD_TWO_APP_NAME),
+        ops_test.model.wait_for_idle(
+            apps=[CONFIG_SERVER_APP_NAME],
+            timeout=1000,
+            idle_period=20,
+            status=f"Waiting for shards to upgrade/downgrade to revision {revision}-locally built.",
```
MiaAltieri, line 133: nice

tests/integration/upgrade/test_sharding_rollback.py
```diff
@@ -102,36 +101,54 @@ async def test_rollback_on_config_server(
...
     ops_test: OpsTest, continuous_writes_to_shard_one, continuous_writes_to_shard_two
 ) -> None:
     """Verify that a config-server and shard can safely rollback without losing writes."""
     new_charm = await ops_test.build_charm(".")
     await run_upgrade_sequence(ops_test, CONFIG_SERVER_APP_NAME, new_charm=new_charm)
 
-    shard_unit = await find_unit(ops_test, leader=True, app_name=SHARD_ONE_APP_NAME)
-    action = await shard_unit.run_action("pre-upgrade-check")
-    await action.wait()
-    assert action.status == "completed", "pre-upgrade-check failed, expected to succeed."
```
MiaAltieri, line 123: QQ why was this removed? Was it because it is checked as part of the function `run_upgrade_sequence`

tests/integration/upgrade/test_sharding_rollback.py
```diff
@@ -102,36 +101,54 @@ async def test_rollback_on_config_server(
...
+            timeout=1000,
+            idle_period=20,
+            status=f"Waiting for shards to upgrade/downgrade to revision {revision}-locally built.",
+        ),
+    )
 
-    # TODO: Use this when https://github.com/juju/python-libjuju/issues/1086 is fixed
-    # await ops_test.model.applications[SHARD_ONE_APP_NAME].refresh(
-    #     channel="6/edge", switch="ch:mongodb"
-    # )
```
MiaAltieri, line 128: nice 😎 so this is solved? 🤩

lib/charms/mongodb/v0/set_status.py
```diff
@@ -153,7 +153,7 @@ def is_unit_status_ready_for_upgrade(self) -> bool:
         if isinstance(current_status, ActiveStatus):
             return True
 
-        if not isinstance(current_status, WaitingStatus):
+        if not isinstance(current_status, BlockedStatus):
```
MiaAltieri, line 156: what about [this case](https://github.com/canonical/mongodb-operator/blob/74fd8c4c5c29e7e56bec9dc268d014f07773c712/src/charm.py#L1591)

Can this logic can be re worked to include both ?

FWIW I hate the checking of status for upgrades, I am going to investigate if it is truly needed. But the results of that we prob won't see for a few weeks / months, so best to patch this and move on until we know more

MiaAltieri (changes requested):

tests/integration/upgrade/test_sharding_rollback.py
Gu1nness: Yeah exactly.
This part was replicating the `run_upgrade_sequence` but without calling the function, so better do with the neat function instead of code duplication.

tests/integration/upgrade/test_sharding_rollback.py
Gu1nness: This was the upgrade which is now ran as `run_upgrade_sequence`, and on top of that, the refresh here was wrong since the deployed version at this moment is already on `6/edge` so we actually want to upgrade to local charm.

lib/charms/mongodb/v0/set_status.py
Gu1nness: This function is only called when it's a shard as you can see first by the `status_message` that is checked afterward. The case you are sending happens only when the charm is running as `config server`.
I think it might be valid here to check again that it's a `shard` to avoid future bugs if this function starts being used elsewhere on a config server for example.

MiaAltieri: > I think it might be valid here to check again that it's a shard to avoid future bugs if this function starts being used elsewhere on a config server for example.

plus one on this

MiaAltieri (approved):

Mehdi-Bendriss (approved):

PR #465, opened 2024-08-22
[DPE-4904] TLS doesn't restart charm if db isn't available

Gu1nness:
## Issue
 * In the ticket, we saw that sometimes mongodb tls library tries to restart the `mongod` service before the charm had ran the pebble ready event, messing with the log directory and data directory permissions.
 * TLS shouldn't restart the mongod service if the db isn't initialised already to avoid race conditions

## Solution

 * Add the checks on `db_initialised` in the library

## Follow Up
 * Update mongodb k8s charm with this library + a patch for another race condition in storage

### DIFF
--- a/tests/unit/test_tls_lib.py
+++ b/tests/unit/test_tls_lib.py
@@ -21,6 +21,7 @@ class TestMongoTLS(unittest.TestCase):
         self.harness = Harness(MongodbOperatorCharm)
         self.harness.begin()
         self.harness.add_relation("database-peers", "database-peers")
+        self.harness.charm.app_peer_data["db_initialised"] = "True"
         self.harness.set_leader(True)
         self.charm = self.harness.charm
         self.addCleanup(self.harness.cleanup)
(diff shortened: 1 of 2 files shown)


tests/unit/test_tls_lib.py
```diff
@@ -224,6 +225,40 @@ def test_unknown_certificate_available(self, restart_charm_services, _):
 
         restart_charm_services.assert_not_called()
 
+    @patch_network_get(private_address="1.1.1.1")
+    @patch("charm.MongodbOperatorCharm.push_tls_certificate_to_workload")
+    @patch("ops.framework.EventBase.defer")
+    def test_external_certificate_available_deferred(self, defer, _):
+        """Tests behavior when external certificate is made available."""
+        del self.harness.charm.app_peer_data["db_initialised"]
```
MiaAltieri, line 233: nice 😎

MiaAltieri (approved):

delgod (approved):

Mehdi-Bendriss (approved):

PR #468, opened 2024-08-28
[DPE-5227][DPE-5228]: Enable Storage reuse test

Gu1nness:
## Issue

 * This test is disabled and also needs some fixes
 * There is no test for storage reuse on a new cluster

## Solution

 * Enable the test and fix the small bugs
 * Develop the `Storage reuse on different cluster` test, and skip it as it's not a supported feature for now.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -9,7 +9,7 @@ import pwd
 import subprocess
 import time
 from pathlib import Path
-from typing import Dict, List, Optional, Set
+from typing import Dict, List, Mapping, Optional, Set
 
 from charms.grafana_agent.v0.cos_agent import COSAgentProvider
 from charms.mongodb.v0.config_server_interface import ClusterProvider
@@ -46,6 +46,7 @@ from data_platform_helpers.version_check import (
     NoVersionError,
     get_charm_revision,
 )
+from ops import StorageAttachedEvent
 from ops.charm import (
     ActionEvent,
     CharmBase,
@@ -108,6 +109,7 @@ class MongodbOperatorCharm(CharmBase):
         self.framework.observe(self.on.install, self._on_install)
         self.framework.observe(self.on.start, self._on_start)
         self.framework.observe(self.on.update_status, self._on_update_status)
+        self.framework.observe(self.on.mongodb_storage_attached, self._on_storage_attached)
         self.framework.observe(
             self.on[Config.Relations.PEERS].relation_joined, self._on_relation_joined
         )
@@ -148,7 +150,7 @@ class MongodbOperatorCharm(CharmBase):
             ],
         )
         self.upgrade = MongoDBUpgrade(self)
-        self.config_server = ShardingProvider(self, substrate="vm")
+        self.config_server = ShardingProvider(self, substrate=Config.SUBSTRATE)
         self.cluster = ClusterProvider(self)
         self.shard = ConfigServerRequirer(self)
         self.status = MongoDBStatusHandler(self)
@@ -587,6 +589,36 @@ class MongodbOperatorCharm(CharmBase):
             )
 
         self._update_hosts(event)
+
+    def exec(
+        self,
+        command: list[str] | str,
+        env: Mapping[str, str] | None = None,
+        working_dir: str | None = None,
+    ) -> str:
+        """Execs a command on the workload in a safe way."""
+        try:
+            output = subprocess.check_output(
+                command,
+                stderr=subprocess.PIPE,
+                universal_newlines=True,
+                shell=isinstance(command, str),
+                env=env,
+                cwd=working_dir,
+            )
+            logger.debug(f"{output=}")
+            return output
+        except subprocess.CalledProcessError as err:
+            logger.error(f"cmd failed - {err.cmd = }, {err.stdout = }, {err.stderr = }")
+            raise
+
+    def _on_storage_attached(self, event: StorageAttachedEvent) -> None:
+        """Handler for `storage_attached` event.
+
+        This should handle fixing the permissions for the data dir.
+        """
+        self.exec(f"chmod -R 770 {Config.MONGODB_COMMON_PATH}".split())
+        self.exec(f"chown -R {Config.SNAP_USER}:root  {Config.MONGODB_COMMON_PATH}".split())
 
     def _on_storage_detaching(self, event: StorageDetachingEvent) -> None:
         """Before storage detaches, allow removing unit to remove itself from the set.
@@ -1060,7 +1092,7 @@ class MongodbOperatorCharm(CharmBase):
                 logger.error(
                     "An exception occurred when installing %s. Reason: %s", snap_name, str(e)
                 )
-                raise
+            raise
 
     def _instatiate_keyfile(self, event: StartEvent) -> None:
         # wait for keyFile to be created by leader unit
--- a/src/charm.py
+++ b/src/charm.py
@@ -1092,7 +1092,7 @@ class MongodbOperatorCharm(CharmBase):
                 logger.error(
                     "An exception occurred when installing %s. Reason: %s", snap_name, str(e)
                 )
-            raise
+                raise
 
     def _instatiate_keyfile(self, event: StartEvent) -> None:
         # wait for keyFile to be created by leader unit
--- a/src/charm.py
+++ b/src/charm.py
@@ -9,7 +9,7 @@ import pwd
 import subprocess
 import time
 from pathlib import Path
-from typing import Dict, List, Mapping, Optional, Set
+from typing import Dict, List, Optional, Set
 
 from charms.grafana_agent.v0.cos_agent import COSAgentProvider
 from charms.mongodb.v0.config_server_interface import ClusterProvider
@@ -26,6 +26,7 @@ from charms.mongodb.v1.helpers import (
     generate_keyfile,
     generate_password,
     get_create_user_cmd,
+    safe_exec,
 )
 from charms.mongodb.v1.mongodb import MongoDBConnection, NotReadyError
 from charms.mongodb.v1.mongodb_backups import MongoDBBackups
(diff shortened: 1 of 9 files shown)


src/charm.py
```diff
@@ -588,6 +590,36 @@ def _on_relation_departed(self, event: RelationDepartedEvent) -> None:
...
+                universal_newlines=True,
+                shell=isinstance(command, str),
+                env=env,
+                cwd=working_dir,
+            )
+            logger.debug(f"{output=}")
+            return output
+        except subprocess.CalledProcessError as err:
+            logger.error(f"cmd failed - {err.cmd = }, {err.stdout = }, {err.stderr = }")
+            raise
```
MiaAltieri: Please move this to helpers :)

src/config.py
```diff
@@ -21,7 +22,13 @@ class Config:
     MONGOD_CONF_DIR = f"{MONGODB_SNAP_DATA_DIR}/etc/mongod"
     MONGOD_CONF_FILE_PATH = f"{MONGOD_CONF_DIR}/mongod.conf"
     CHARM_INTERNAL_VERSION_FILE = "charm_internal_version"
-    SNAP_PACKAGES = [("charmed-mongodb", "6/edge", 118)]
+    SNAP_PACKAGES = [("charmed-mongodb", "6/edge", 121)]
```
MiaAltieri, line 25: nice 🤩

tests/integration/ha_tests/helpers.py
```diff
@@ -450,11 +423,21 @@ async def reused_storage(ops_test: OpsTest, unit_name, removal_time) -> bool:
 
         item = json.loads(line)
 
+        # We need a msg
```
MiaAltieri: can you add why?

tests/integration/ha_tests/helpers.py
```diff
@@ -450,11 +423,21 @@ async def reused_storage(ops_test: OpsTest, unit_name, removal_time) -> bool:
...
         if "msg" not in item:
             continue
 
+        # We need the attrs
```
MiaAltieri: can you add why?

maybe something like:
`# "attr" is needed used to gather state information from mongodb`

tests/integration/ha_tests/helpers.py
```diff
@@ -641,10 +624,10 @@ async def verify_replica_set_configuration(ops_test: OpsTest, app_name=None) ->
 
 
 def convert_time(time_as_str: str) -> int:
-    """Converts a string time representation to an integer time representation."""
+    """Converts a string time representation to an integer time representation, in UTC."""
     # parse time representation, provided in this format: 'YYYY-MM-DDTHH:MM:SS.MMM+00:00'
     d = datetime.strptime(time_as_str, "%Y-%m-%dT%H:%M:%S.%f%z")
-    return time.mktime(d.timetuple())
+    return calendar.timegm(d.timetuple())
```
MiaAltieri, line 626: 🙏

tests/integration/ha_tests/test_ha.py
```diff
@@ -77,12 +77,16 @@ async def test_build_and_deploy(ops_test: OpsTest) -> None:
         return
 
     my_charm = await ops_test.build_charm(".")
-    await ops_test.model.deploy(my_charm, num_units=required_units)
+
+    storage = {"mongodb": {"pool": "lxd", "size": 2048}}
```
MiaAltieri, line 81: nice 😎

tests/integration/ha_tests/test_ha.py
```diff
@@ -122,6 +132,54 @@ async def test_storage_re_use(ops_test, continuous_writes):
     assert total_expected_writes["number"] == actual_writes
 
 
+@pytest.mark.runner(["self-hosted", "linux", "X64", "jammy", "large"])
+@pytest.mark.group(1)
+@pytest.mark.skip("This is currently unsupported on MongoDB charm.")
```
MiaAltieri, line 137: According to our chat in Mattermost I believe the next steps are to research this problem. 

I am okay with keeping this test here as skipped and doing the research separately (but best to remove the DPE-XXX from the title of this PR)

MiaAltieri (changes requested):

src/charm.py
Gu1nness: Sure!

src/config.py
Gu1nness: It made me find some bugs in the upgrade mechanism with revision upgrade ;)

tests/integration/ha_tests/helpers.py
Gu1nness: Thing is I don't know, we are not using the message, I was just trying to add a bit more documentation ^^"

tests/integration/ha_tests/helpers.py
Gu1nness: Sure

tests/integration/ha_tests/test_ha.py
Gu1nness: Hum.
The ticket was about adding the test, not the support of this feature.
I'd rather keep the ticket id here and create a new one for the research+optional implementation.

src/charm.py
```diff
@@ -588,6 +591,14 @@ def _on_relation_departed(self, event: RelationDepartedEvent) -> None:
 
         self._update_hosts(event)
 
+    def _on_storage_attached(self, event: StorageAttachedEvent) -> None:
+        """Handler for `storage_attached` event.
+
+        This should handle fixing the permissions for the data dir.
+        """
+        safe_exec(f"chmod -R 770 {Config.MONGODB_COMMON_PATH}".split())
+        safe_exec(f"chown -R {Config.SNAP_USER}:root  {Config.MONGODB_COMMON_PATH}".split())
```
Mehdi-Bendriss, line 600: Can you explain this? isn't the snap already setting the right permission / ownership?

tests/integration/ha_tests/helpers.py
Mehdi-Bendriss: I don't have context in this one, could you explain to me what `msg` represents in a log line?

tests/integration/ha_tests/test_ha.py
Mehdi-Bendriss: I agree that the test is not very useful of it skipped since we can't know if it works :D 
However, given that we discovered it wasn't supported out-of-the-box, we can create a new ticket for the implementation and start on it immediately.

Mehdi-Bendriss (commented):
Thanks Neha, I have a couple of questions

src/charm.py
Gu1nness: When we are attaching the storage, we need to fix the permissions on it.
On one side, we fixed the snap so it does not recreate and give permissions if the directories are already existing.
On this side, we grant the right permissions to the already existing directories so the snap will not break because of permissions error. If we don't fix those permissions, there will be an installation error when adding a unit with an already existing storage, as the snap will not be allowed to do anything with the existing files, not even reading the configuration file to update it as we do.
This is inspired by the work made on Kafka:
 * https://github.com/canonical/kafka-operator/pull/195
 * https://github.com/canonical/charmed-kafka-snap/pull/40

Gu1nness:
> I don't have context in this one, could you explain to me what msg represents in a log line?

As I said to @MiaAltieri I don't know and this part was already in the test. It seems that every log line should contain a `msg` field ([see here](https://www.mongodb.com/docs/manual/reference/log-messages/)).
I can remove it but I didn't want to change the test if unnecessary.
EDIT: I looked up the changes that added this line and it seems that it's not used anymore, it seems to come from an older version. I can try to remove it now

tests/integration/ha_tests/test_ha.py
Gu1nness: Yeah maybe this can be added to next pulse to have it working and I can start the investigation on what can be done to support that.

tests/integration/ha_tests/helpers.py
MiaAltieri: mind making this change?

tests/integration/ha_tests/test_ha.py
```diff
@@ -122,6 +132,54 @@ async def test_storage_re_use(ops_test, continuous_writes):
...
+@pytest.mark.abort_on_fail
+async def test_storage_re_use_different_cluster(ops_test: OpsTest, continuous_writes):
```
MiaAltieri, line 139: this test may need to change if we support it, but I am happy to keep for now

tests/integration/ha_tests/test_ha.py
```diff
@@ -122,6 +132,54 @@ async def test_storage_re_use(ops_test, continuous_writes):
...
+
+    writes_results = await stop_continous_writes(ops_test, app_name=app_name)
+    unit_ids = [unit.name for unit in ops_test.model.applications[app_name].units]
+    storage_ids = {}
+    for unit_id in unit_ids:
+        storage_ids[unit_id] = storage_id(ops_test, unit_id)
+        await ops_test.model.applications[app_name].destroy_unit(unit_id)
+        # Give some time to remove the unit. We don't use asyncio.sleep here to
+        # leave time for each unit to be removed before removing the next one.
+        time.sleep(60)
```
MiaAltieri: might be better to ensure waiting for units.

tests/integration/ha_tests/test_ha.py
```diff
@@ -122,6 +132,54 @@ async def test_storage_re_use(ops_test, continuous_writes):
...
+        pytest.skip(
+            "reuse of storage can only be used on deployments with persistent storage not on rootfs deployments"
+        )
+
+    writes_results = await stop_continous_writes(ops_test, app_name=app_name)
+    unit_ids = [unit.name for unit in ops_test.model.applications[app_name].units]
+    storage_ids = {}
+    for unit_id in unit_ids:
+        storage_ids[unit_id] = storage_id(ops_test, unit_id)
+        await ops_test.model.applications[app_name].destroy_unit(unit_id)
```
MiaAltieri, line 159: remove unit would be better/safer here .

tests/integration/ha_tests/test_ha.py
```diff
@@ -122,6 +132,54 @@ async def test_storage_re_use(ops_test, continuous_writes):
...
+        time.sleep(60)
+
+    # Wait until all apps are cleaned up
+    await ops_test.model.wait_for_idle(apps=[app_name], timeout=1000, wait_for_exact_units=0)
+
+    for unit_id in unit_ids:
+        await ops_test.model.applications[app_name].add_unit(
+            count=1, attach_storage=[tag.storage(storage_ids[unit_id])]
+        )
+        time.sleep(10)
```
MiaAltieri: same here

tests/integration/ha_tests/test_ha.py
Gu1nness: The method name to remove a unit is `destroy_unit` and that's what we use everywhere in our tests.

tests/integration/ha_tests/test_ha.py
Gu1nness: Sure, let me do that

tests/integration/ha_tests/helpers.py
Gu1nness: Oh my bad I thought I had done it

Gu1nness: It has been done

MiaAltieri (approved):

Mehdi-Bendriss (approved):

PR #472, opened 2024-08-30
[DPE-5215] feat(libs): Correct use of data_interfaces

Gu1nness:
## Issue
 * In the Mongos k8s charm, we saw that the data which is expected to be stored as secrets are shown in plain sight.
 * A thorough investigation and the help of @juditnovak lead us to see that we were not using the data_interfaces library the correct way, breaking the expected flow, which lead to this issue.

## Solution

This library was not using the data interface library the right way. The provider should not write in the databag before the `database_requested` event which is fired when the requirer asks for the DB and write its name + the `requested-secrets` list in the databag.

This commit hence:
 * Defines a handler for the `database_requested` event
 * Protects the call to `relation_changed` events by checking first if the `database` field is in the relation databag, which can happen only after the `database_requested` event has run.

### DIFF
--- a/lib/charms/mongodb/v0/config_server_interface.py
+++ b/lib/charms/mongodb/v0/config_server_interface.py
@@ -11,10 +11,11 @@ from typing import Optional
 
 from charms.data_platform_libs.v0.data_interfaces import (
     DatabaseProvides,
+    DatabaseRequestedEvent,
     DatabaseRequires,
 )
 from charms.mongodb.v1.mongos import MongosConnection
-from ops.charm import CharmBase, EventBase, RelationBrokenEvent
+from ops.charm import CharmBase, EventBase, RelationBrokenEvent, RelationChangedEvent
 from ops.framework import Object
 from ops.model import (
     ActiveStatus,
@@ -42,7 +43,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 11
+LIBPATCH = 12
 
 
 class ClusterProvider(Object):
@@ -57,6 +58,9 @@ class ClusterProvider(Object):
         self.database_provides = DatabaseProvides(self.charm, relation_name=self.relation_name)
 
         super().__init__(charm, self.relation_name)
+        self.framework.observe(
+            self.database_provides.on.database_requested, self._on_database_requested
+        )
         self.framework.observe(
             charm.on[self.relation_name].relation_changed, self._on_relation_changed
         )
@@ -105,8 +109,8 @@ class ClusterProvider(Object):
 
         return True
 
-    def _on_relation_changed(self, event) -> None:
-        """Handles providing mongos with KeyFile and hosts."""
+    def _on_database_requested(self, event: DatabaseRequestedEvent | RelationChangedEvent) -> None:
+        """Handles the database requested event, which is the first time we can write data."""
         if not self.pass_hook_checks(event):
             if not self.is_valid_mongos_integration():
                 self.charm.status.set_and_share_status(
@@ -116,12 +120,9 @@ class ClusterProvider(Object):
                 )
             logger.info("Skipping relation joined event: hook checks did not pass")
             return
-
         config_server_db = self.generate_config_server_db()
-
         # create user and set secrets for mongos relation
         self.charm.client_relations.oversee_users(None, None)
-
         relation_data = {
             KEYFILE_KEY: self.charm.get_secret(
                 Config.Relations.APP_SCOPE, Config.Secrets.SECRET_KEYFILE_NAME
@@ -135,8 +136,18 @@ class ClusterProvider(Object):
         )
         if int_tls_ca:
             relation_data[INT_TLS_CA_KEY] = int_tls_ca
-
         self.database_provides.update_relation_data(event.relation.id, relation_data)
+
+    def _on_relation_changed(self, event: RelationChangedEvent) -> None:
+        """Handles providing mongos with KeyFile and hosts."""
+        # First we need to ensure that the database requested event has run
+        # otherwise we would write secrets in plain sight.
+        if not self.database_provides.fetch_relation_field(event.relation.id, "database"):
+            logger.info("Database Requested has not run yet, skipping.")
+            event.defer()
+            return
+
+        self._on_database_requested(event)
 
     def _on_relation_broken(self, event) -> None:
         if self.charm.upgrade_in_progress:
--- a/lib/charms/mongodb/v0/config_server_interface.py
+++ b/lib/charms/mongodb/v0/config_server_interface.py
@@ -110,7 +110,10 @@ class ClusterProvider(Object):
         return True
 
     def _on_database_requested(self, event: DatabaseRequestedEvent | RelationChangedEvent) -> None:
-        """Handles the database requested event, which is the first time we can write data."""
+        """Handles the database requested event, the first time secrets are written to relations should be on this event.
+        
+        Note: If secrets are written for the first time on other events we risk the chance of writing secrets in plain sight
+        """
         if not self.pass_hook_checks(event):
             if not self.is_valid_mongos_integration():
                 self.charm.status.set_and_share_status(
--- a/lib/charms/mongodb/v0/config_server_interface.py
+++ b/lib/charms/mongodb/v0/config_server_interface.py
@@ -110,9 +110,12 @@ class ClusterProvider(Object):
         return True
 
     def _on_database_requested(self, event: DatabaseRequestedEvent | RelationChangedEvent) -> None:
-        """Handles the database requested event, the first time secrets are written to relations should be on this event.
-        
-        Note: If secrets are written for the first time on other events we risk the chance of writing secrets in plain sight
+        """Handles the database requested event.
+
+        The first time secrets are written to relations should be on this event.
+
+        Note: If secrets are written for the first time on other events we risk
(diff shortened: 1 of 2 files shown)


lib/charms/mongodb/v0/config_server_interface.py
```diff
@@ -116,12 +120,9 @@ def _on_relation_changed(self, event) -> None:
                 )
             logger.info("Skipping relation joined event: hook checks did not pass")
             return
-
         config_server_db = self.generate_config_server_db()
-
         # create user and set secrets for mongos relation
         self.charm.client_relations.oversee_users(None, None)
```
juditnovak, line 131: Is there a way we could split the code such as `set_credentials()` is extracted from `oversee_users()`?
The current soultion is not quite the right thing...

Basically the code structure should go as such:

```
def _on_database_requested(self, event: DatabaseRequestedEvent):
   """Distribute credentials to the Requirer within the scope of setting and initializing the connection.
   
   NOTE: These credentials can NOT be updated throughout the lifetime of the relation by Data Platform policy
   """
    self.provider.set_credentials(event.relation.id, username, password)
    # Optionally
    self.provider.update_relation_data(event.relation.id, <any additional data>)

def _on_relation_changed(self, event: RelationChangedEvent):
    """Any actions to be performed in case the relation is updated."""
   [..]
```

The current code doesn't provide the clear distinction between the strictly one-off `set_credential()` operation (ONLY to be performed on `database-requested` event!) other operations that are to be executed regularly. I rather suggest to introduce the correct usage of the libs (see references on usage in my Review Comment). 

Practically it means the few lines of code I've typed above :-)

lib/charms/mongodb/v0/config_server_interface.py
```diff
@@ -135,9 +136,19 @@ def _on_relation_changed(self, event) -> None:
...
+    def _on_relation_changed(self, event: RelationChangedEvent) -> None:
+        """Handles providing mongos with KeyFile and hosts."""
+        # First we need to ensure that the database requested event has run
+        # otherwise we would write secrets in plain sight.
+        if not self.database_provides.fetch_relation_field(event.relation.id, "database"):
+            logger.info("Database Requested has not run yet, skipping.")
+            event.defer()
+            return
+
+        self._on_database_requested(event)
```
juditnovak, line 157: If I get it right, this event handler is executed when the TLS credentials are updated. What may be the other reasons to trigger the event?

I understand that the `oversee_users()` (which is the main functionality that we expect from the `relation-changed` handler) does the following:
 1. create users
 2. update users
 3. cleanup users 
 4. cleanup databases
 
What I sugges is to split 1) (and somewhat 2)) into `database-requested`. Note that 2 is not supported for user credentials `username` and `password`. (If we want a new password, we need to break the relation and re-relate by current DP Policy. We are planning to change that in the future.)

Thus `oversee_users()` would do update and cleanup, while creation goes to the dedicated event handler where it belongs.

lib/charms/mongodb/v0/config_server_interface.py
```diff
@@ -135,9 +136,19 @@ def _on_relation_changed(self, event) -> None:
...
         self.database_provides.update_relation_data(event.relation.id, relation_data)
 
+    def _on_relation_changed(self, event: RelationChangedEvent) -> None:
+        """Handles providing mongos with KeyFile and hosts."""
+        # First we need to ensure that the database requested event has run
+        # otherwise we would write secrets in plain sight.
+        if not self.database_provides.fetch_relation_field(event.relation.id, "database"):
+            logger.info("Database Requested has not run yet, skipping.")
+            event.defer()
+            return
```
juditnovak, line 154: The above should be removed as soon as moving on to the new version of the libs that I hope to merge next week :-)

This will be addressed on the lib level as highlighted on the [corresponding issue](https://github.com/canonical/data-platform-libs/issues/193)

lib/charms/mongodb/v0/config_server_interface.py
```diff
@@ -105,8 +109,8 @@ def is_valid_mongos_integration(self) -> bool:
 
         return True
 
-    def _on_relation_changed(self, event) -> None:
-        """Handles providing mongos with KeyFile and hosts."""
+    def _on_database_requested(self, event: DatabaseRequestedEvent | RelationChangedEvent) -> None:
+        """Handles the database requested event, which is the first time we can write data."""
```
juditnovak: We shouldn't do this. We should handle `DatabaseRequestedEvent` separate from others, performing its dedicated action (user creationg and credentials distribution)

juditnovak (commented):
Guys this technically may do the trick (as of today), but it's not the exactly right thing...
(As of the libs protocol documentation and [spec](https://docs.google.com/document/d/1UpIZbBz-af2-uNLVTijdbwuW9ASy4IOE5tU9dVEY1Ho/edit?usp=sharing) (*))

(*) Note: the spec is informally pre-approved, I hope for official approvals in the near future.

lib/charms/mongodb/v0/config_server_interface.py
Gu1nness: The thing is that `oversee_users` is a reconcile pattern that handles user creation, user deletion, and many other things.
I'm not sure we can easily remove the `set_credentials` from this function (it's already embedded in another function doing many things).
I need to talk with @MiaAltieri about that to decide what we could do because it might imply bigger changes, and I'm not sure if we have situations were users are created in the `relation_joined` event (it would make it way easier if that never happens.
Also I know the oversee users function is also used elsewhere for another reason I forgot.

Gu1nness:
I have logged [a ticket](https://warthogs.atlassian.net/browse/DPE-5313) to do the implementation you are suggesting @juditnovak but I think this is out of scope for this ticket and I don't have some additional bandwidth to handle this during this pulse. This will be reviewed, scoped, and tackled as soon as we have some time.

lib/charms/mongodb/v0/config_server_interface.py
```diff
@@ -57,6 +58,9 @@ def __init__(
         self.database_provides = DatabaseProvides(self.charm, relation_name=self.relation_name)
 
         super().__init__(charm, self.relation_name)
+        self.framework.observe(
```
MiaAltieri, line 61: 🎉

lib/charms/mongodb/v0/config_server_interface.py
MiaAltieri: Great suggestion @juditnovak ! Certainly appreciate your keen eyes on this PR 🙇 

Lets definitely do this, I want to be sure that we use the lib correctly. I think it is out of scope for this PR but I believe there is a ticket that @Gu1nness created that should address it. Maybe it can even been addressed in pairing session in the future pulse.

lib/charms/mongodb/v0/config_server_interface.py
```diff
@@ -135,9 +136,20 @@ def _on_relation_changed(self, event) -> None:
         )
         if int_tls_ca:
             relation_data[INT_TLS_CA_KEY] = int_tls_ca
-
         self.database_provides.update_relation_data(event.relation.id, relation_data)
 
+    def _on_relation_changed(self, event: RelationChangedEvent) -> None:
+        """Handles providing mongos with KeyFile and hosts."""
+        # First we need to ensure that the database requested event has run
+        # otherwise we would write secrets in plain sight.
```
MiaAltieri:
```suggestion
        # First we need to ensure that the database requested event has run
        # otherwise we risk the chance of writing secrets in plain sight.
```

lib/charms/mongodb/v0/config_server_interface.py
MiaAltieri:
```suggestion
        """Handles the database requested event, the first time secrets are written to relations should be on this event.
        
        Note: If secrets are written for the first time on other events we risk the chance of writing secrets in plain site
        """
```

lib/charms/mongodb/v0/config_server_interface.py
juditnovak: No worries, I'm happy to help :-)

However, I'm struggling to approve the current solution. (I find it a particularly important matter that security-related workflows are followed strictly... Compared to which the current solution feels like a hack... :-( )

Is it OK if I just participated this PR as an "advisor", and approve on the one that's implementing the right thing?
(It should not block you from merging, you Guys can provide enough approvals w/o me.)

lib/charms/mongodb/v0/config_server_interface.py
Gu1nness:
```suggestion
        """Handles the database requested event, the first time secrets are written to relations should be on this event.
        
        Note: If secrets are written for the first time on other events we risk the chance of writing secrets in plain sight
        """
```

lib/charms/mongodb/v0/config_server_interface.py
Gu1nness: It is completely OK to me.
As we talked about, I know this first solution is not sufficient to have a good and clear path, but at least it works.
I have already created a new ticket in order to have a better solution, and I hope to bring it in next pulse if I can, and I would love your advice on the next PR as well since you know the library and its ways, way better than I do :)

lib/charms/mongodb/v0/config_server_interface.py
MiaAltieri: It seems that more and more charms are following reconcile patterns (our oversee users function follows that pattern)

I wonder how other reconcile charms handle this.

MiaAltieri (approved):

Mehdi-Bendriss (approved):

PR #473, opened 2024-08-30
Sharding relations test are flaky

Gu1nness:
## Issue

 * Test fails 90% of the time.
 * The real issue here is an unprotected call to `charm.client_provider.update_app_relation_data` which would create the password only for an invalid relation (integrating a shard as database).
 * This call is run in every status check.

## Solution

 * Protect the call by running some sanity checks.
 * Update all tests that started failing due to this (needed to add some monkey patching for all methods used in this checks.

## Bug fixes

 * The `charm.db_initialised` function was buggy: setting `charm.db_initialised = False` would still return `charm.db_initialised == True`. This is now fixed in a retro compatible way to ensure that nothing breaks with upgrades.
 * This fix needs to be ported to mongodb k8s as well.

### DIFF
--- a/lib/charms/mongodb/v1/mongodb_provider.py
+++ b/lib/charms/mongodb/v1/mongodb_provider.py
@@ -77,7 +77,7 @@ class MongoDBProvider(Object):
             self.database_provides.on.database_requested, self._on_relation_event
         )
 
-    def sanity_hook_cheks(self) -> bool:
+    def pass_sanity_hook_checks(self) -> bool:
         """Runs reusable and event agnostic checks."""
         # We shouldn't try to create or update users if the database is not
         # initialised. We will create users as part of initialisation.
@@ -97,7 +97,7 @@ class MongoDBProvider(Object):
 
     def pass_hook_checks(self, event: EventBase) -> bool:
         """Runs the pre-hooks checks for MongoDBProvider, returns True if all pass."""
-        if not self.sanity_hook_cheks():
+        if not self.pass_sanity_hook_checks():
             return False
 
         if self.charm.upgrade_in_progress:
@@ -248,7 +248,7 @@ class MongoDBProvider(Object):
 
     def update_app_relation_data(self) -> None:
         """Helper function to update application relation data."""
-        if not self.sanity_hook_cheks():
+        if not self.pass_sanity_hook_checks():
             return
 
         database_users = set()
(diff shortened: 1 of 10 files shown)


MiaAltieri:
Note for readers test failing will be addressed #473

lib/charms/mongodb/v1/mongodb_provider.py
```diff
@@ -76,8 +76,8 @@ def __init__(self, charm: CharmBase, substrate="k8s", relation_name: str = REL_N
             self.database_provides.on.database_requested, self._on_relation_event
         )
 
-    def pass_hook_checks(self, event: EventBase) -> bool:
-        """Runs the pre-hooks checks for MongoDBProvider, returns True if all pass."""
+    def sanity_hook_cheks(self) -> bool:
```
Mehdi-Bendriss: 1. Could you rename this method with the `is_` prefix? it is unclear from the name what this check does. 
2. Why don't we extend the existing `pass_hook_checks` and call it everywhere instead of a second method?

Mehdi-Bendriss (commented):
Thanks Neha - left a clarification point.

lib/charms/mongodb/v1/mongodb_provider.py
Gu1nness: The `pass_hook_checks` has two parts:
 * The first one which I extracted to the `sanity_hook_checks` which is event independent
 * The second part is event dependent and can defer the event.
 
 In this case, we don't want to defer the `update_status` event, we just want to protect the call.
 That's why I decided to split the hook checks between the sanity ones which always have to be run, and the ones that can defer the event.
The existing `pass_hook_checks` hasn't changed it what it does, I just went for reusability :)

Gu1nness: > 1. Could you rename this method with the `is_` prefix? it is unclear from the name what this check does.

I think I'll go to `pass_sanity_hook_checks` which is a method name that we already have in other libraries, this way we'll keep some uniformity here.

tests/integration/sharding_tests/test_sharding_relations.py
```diff
@@ -55,7 +55,6 @@ async def test_build_and_deploy(
     await ops_test.model.deploy(
         MONGOS_APP_NAME,
         channel="6/edge",
-        revision=3,
```
MiaAltieri, line 58: ty

MiaAltieri (dismissed):

MiaAltieri (approved):

Mehdi-Bendriss (approved):

PR #476, opened 2024-09-04
revert recent lib changes 

MiaAltieri:
## Issue
Previous changes made it so that on Mongos-K8s would return different endpoints to inside K8s apps and outside k8s apps if expose-external enabled.

However it was decided recently that this approach is incorrect.

## Solution

Revert this change and bump libpatch

### DIFF
--- a/lib/charms/mongodb/v1/mongodb_provider.py
+++ b/lib/charms/mongodb/v1/mongodb_provider.py
@@ -31,7 +31,7 @@ LIBAPI = 1
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 12
+LIBPATCH = 13
 
 logger = logging.getLogger(__name__)
 REL_NAME = "database"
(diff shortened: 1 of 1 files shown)


lib/charms/mongodb/v1/mongodb_provider.py
```diff
@@ -303,9 +303,7 @@ def _get_config(self, username: str, password: Optional[str]) -> MongoConfigurat
         if self.charm.is_role(Config.Role.MONGOS):
             mongo_args["port"] = Config.MONGOS_PORT
             if self.substrate == Config.Substrate.K8S:
-                external = self.is_external_client(relation.id)
-                mongo_args["port"] = self.charm.get_mongos_port(external)
-                mongo_args["hosts"] = self.charm.get_mongos_hosts(external)
+                mongo_args["port"] = self.charm.get_mongos_port()
```
MiaAltieri, line 306: note - now the retrieval of hosts + ips is done in charm side on mongos k8s and is **only based on whether or not expose-external is toggled on** 



https://github.com/canonical/mongos-k8s-operator/blob/25f2b05145c2ea4b74b92e0be0e4f7c37cc3b147/src/charm.py#L410-L436

reneradoi (approved):

phvalguima (approved):

PR #478, opened 2024-09-06
[DPE-5236] update TLS lib for K8s mongos external clients

MiaAltieri:
## Issue
K8s mongos charm needs to be able to compare the current sans with the expected sans in order to update TLS certs when the k8s ip changes 

## Solution
Update lib to support the needed functionality

### DIFF
--- a/lib/charms/mongodb/v1/helpers.py
+++ b/lib/charms/mongodb/v1/helpers.py
@@ -340,5 +340,5 @@ def safe_exec(
         logger.debug(f"{output=}")
         return output
     except subprocess.CalledProcessError as err:
-        logger.error(f"cmd failed - {err.cmd = }, {err.stdout = }, {err.stderr = }")
+        logger.error(f"cmd failed - {err.cmd}, {err.stdout}, {err.stderr}")
         raise
--- a/lib/charms/mongodb/v1/helpers.py
+++ b/lib/charms/mongodb/v1/helpers.py
@@ -23,7 +23,7 @@ LIBAPI = 1
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 9
+LIBPATCH = 10
 
 # path to store mongodb ketFile
 KEY_FILE = "keyFile"
--- a/lib/charms/mongodb/v1/helpers.py
+++ b/lib/charms/mongodb/v1/helpers.py
@@ -23,7 +23,7 @@ LIBAPI = 1
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 10
+LIBPATCH = 9
 
 # path to store mongodb ketFile
 KEY_FILE = "keyFile"
@@ -340,5 +340,5 @@ def safe_exec(
         logger.debug(f"{output=}")
         return output
     except subprocess.CalledProcessError as err:
-        logger.error(f"cmd failed - {err.cmd}, {err.stdout}, {err.stderr}")
+        logger.error(f"cmd failed - {err.cmd = }, {err.stdout = }, {err.stderr = }")
         raise
--- a/lib/charms/mongodb/v1/mongodb_tls.py
+++ b/lib/charms/mongodb/v1/mongodb_tls.py
@@ -38,7 +38,7 @@ SANS_IPS_KEY = "sans_ips"
 LIBID = "e02a50f0795e4dd292f58e93b4f493dd"
 
 # Increment this major API version when introducing breaking changes
-LIBAPI = 1
+LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
@@ -105,7 +105,7 @@ class MongoDBTLS(Object):
         internal: bool,
     ):
         """Request TLS certificate."""
-        if self.charm.model.get_relation(Config.TLS.TLS_PEER_RELATION):
+        if not self.charm.model.get_relation(Config.TLS.TLS_PEER_RELATION):
             return
 
         if param is None:
@@ -128,7 +128,6 @@ class MongoDBTLS(Object):
         label = "int" if internal else "ext"
         self.charm.unit_peer_data[f"{label}_certs_subject"] = self._get_subject_name()
         self.charm.unit_peer_data[f"{label}_certs_subject"] = self._get_subject_name()
-
         self.certs.request_certificate_creation(certificate_signing_request=csr)
         self.set_waiting_for_cert_to_update(internal=internal, waiting=True)
 
@@ -229,12 +228,12 @@ class MongoDBTLS(Object):
         )
         self.set_tls_secret(internal, Config.TLS.SECRET_CERT_LABEL, event.certificate)
         self.set_tls_secret(internal, Config.TLS.SECRET_CA_LABEL, event.ca)
+        self.set_waiting_for_cert_to_update(waiting=False, internal=internal)
 
         if self.charm.is_role(Config.Role.CONFIG_SERVER) and internal:
             self.charm.cluster.update_ca_secret(new_ca=event.ca)
             self.charm.config_server.update_ca_secret(new_ca=event.ca)
 
-        self.set_waiting_for_cert_to_update(waiting=False, internal=internal)
         if self.waiting_for_both_certs():
             logger.debug(
                 "Defer till both internal and external TLS certificates available to avoid second restart."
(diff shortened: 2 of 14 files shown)


lib/charms/mongodb/v1/helpers.py
```diff
@@ -340,5 +340,5 @@ def safe_exec(
         logger.debug(f"{output=}")
         return output
     except subprocess.CalledProcessError as err:
-        logger.error(f"cmd failed - {err.cmd = }, {err.stdout = }, {err.stderr = }")
+        logger.error(f"cmd failed - {err.cmd}, {err.stdout}, {err.stderr}")
```
Gu1nness: If possible, I'd rather keep the formatting which shows what is what in the error message.
Does this break the linting or something?

lib/charms/mongodb/v1/mongodb_tls.py
```diff
@@ -302,21 +315,93 @@ def _on_certificate_expiring(self, event: CertificateExpiringEvent) -> None:
...
         ]
 
+        if self.charm.is_role(Config.Role.MONGOS) and self.charm.is_external_client:
+            sans[SANS_IPS_KEY].append(
+                self.charm.get_ext_mongos_host(self.charm.unit, incl_port=False)
+            )
+
+        return sans
+
+    def get_current_sans(self, internal: bool) -> List | None:
```
Gu1nness:
```suggestion
    def get_current_sans(self, internal: bool) -> List[str] | None:
```

lib/charms/mongodb/v1/mongodb_tls.py
```diff
@@ -302,21 +315,93 @@ def _on_certificate_expiring(self, event: CertificateExpiringEvent) -> None:
...
+        """Gets the current SANs for the unit cert."""
+        # if unit has no certificates do not proceed.
+        if not self.is_tls_enabled(internal=internal):
+            return
+
+        try:
+            sans_lines = self.get_sans_from_host(internal)
+        except (subprocess.CalledProcessError, ExecError) as e:
+            logger.error(e.stdout)
```
Gu1nness: I would log as well `stderr` since sometimes `stdout` will be almost empty while `stderr` will contain the actual issue.

lib/charms/mongodb/v1/mongodb_tls.py
```diff
@@ -302,21 +315,93 @@ def _on_certificate_expiring(self, event: CertificateExpiringEvent) -> None:
...
+        return {SANS_IPS_KEY: sorted(sans_ip), SANS_DNS_KEY: sorted(sans_dns)}
+
+    def get_sans_from_host(self, internal) -> List[str]:
+        """Returns the sans lines from the host.
+
+        Raises: subprocess.CalledProcessError, ExecError
+        """
+        pem_file = Config.TLS.INT_PEM_FILE if internal else Config.TLS.EXT_PEM_FILE
+
+        command = [
```
Gu1nness: Do we have to ensure that `openssl` is in the machine/container or will it be by default?
I would rather use python `cryptography` to get the DNS/IP instead of doing some manual parsing, what do you think?
It's not mandatory but a bit more clean.

lib/charms/mongodb/v1/mongodb_tls.py
```diff
@@ -365,3 +450,27 @@ def _get_subject_name(self) -> str:
...
+        internal: bool,
+    ):
+        """Sets a boolean indicator, for whether or not we are waiting for a cert to update."""
+        if internal and not self.charm.unit.is_leader():
+            return
+
+        if internal:
+            self.charm.app_peer_data[WAIT_CERT_UPDATE] = json.dumps(waiting)
+        else:
+            self.charm.unit_peer_data[WAIT_CERT_UPDATE] = json.dumps(waiting)
```
Gu1nness:
```suggestion
        if internal and self.charm.unit.is_leader():
            self.charm.app_peer_data[WAIT_CERT_UPDATE] = json.dumps(waiting)
        elif not internal:
            self.charm.unit_peer_data[WAIT_CERT_UPDATE] = json.dumps(waiting)
```

src/config.py
```diff
@@ -130,8 +130,8 @@ class Status:
     class Substrate:
         """Substrate related constants."""
 
-        VM = "vm"
         K8S = "k8s"
+        VM = "vm"
```
Gu1nness: This was done by the linter?

lib/charms/mongodb/v1/mongodb_tls.py
Gu1nness: Coming with an alternate and safe version if you want

Gu1nness: I would go with something along the lines of that:
```python3
from cryptography import x509
from cryptography.hazmat.backends import default_backend

def get_current_sans(self, internal: bool) -> dict[str, str]:
    [Some code to get the certificate data in bytes]
    try:
        cert = cert = x509.load_pem_x509_certificate(data, default_backend())
        sans = cert.extensions.get_extenstion_for_class(x509.SubjectAlternativeName).value
        sans_ip = [str(san) for san in sans.get_values_for_type(x509.IPAddress)]
        sans_dns = [str(san) for san in sans.get_values_for_type(x509.DNSName)]
    except x509.ExtensionNotFound:
        sans_ip = []
        sans_dns = []
    return {SANS_IPS_KEY: sorted(sans_ip), SANS_DNS_KEY: sorted(sans_dns)}
```
Just need to provide the code that gets the certs data (so either from the workload with a cat or get from secrets) and this will be easier, cleaner, no risk of text interface change, and no need to ensure that openssl binary is installed on the workload

lib/charms/mongodb/v1/helpers.py
MiaAltieri: I thought it broke linting, but perhaps its something in my local env - let me see

src/config.py
MiaAltieri: I believe I did it while playing with some things.

MiaAltieri: I'll revert it

lib/charms/mongodb/v1/mongodb_tls.py
MiaAltieri: maybe its that its 17h00 and my coffee is wearing off, but I find this suggestion a little hard to read (even though it is correct)

I am going to close it for now. If you come back to this and feel strongly let me know

lib/charms/mongodb/v1/mongodb_tls.py
MiaAltieri: 🤦‍♀️ - lol yes thank you that was a mistake on my end

lib/charms/mongodb/v1/mongodb_tls.py
MiaAltieri: If this works out of the box then lets go with it - otherwise lets keep it as is (we use the same function here as Kafka - so there is value in that)

lib/charms/mongodb/v1/helpers.py
MiaAltieri: breaks locally but not here. reverted

lib/charms/mongodb/v1/mongodb_tls.py
```diff
@@ -302,21 +315,93 @@ def _on_certificate_expiring(self, event: CertificateExpiringEvent) -> None:
...
+            raise e
+
+        line = ""
+        for line in sans_lines:
+            if "DNS" in line or "IP" in line:
+                break
+
+        sans_ip = []
+        sans_dns = []
+        for item in line.split(", "):
```
Mehdi-Bendriss: should this block be in the above `for` loop? otherwise, when is `line` set in the previous loop?

lib/charms/mongodb/v1/mongodb_tls.py
```diff
@@ -302,21 +315,93 @@ def _on_certificate_expiring(self, event: CertificateExpiringEvent) -> None:
...
+
+        command = [
+            "openssl",
+            "x509",
+            "-noout",
+            "-ext",
+            "subjectAltName",
+            "-in",
+            pem_file,
+        ]
```
Mehdi-Bendriss: We can enrich this command to only have DNS / IP Adrresses in the output, so we only get what we're looking for: 
```
openssl x509 -in newfile.crt.pem -noout -text | grep -oP 'DNS:[^,]+|IP Address:[^,]+' | sort -uV
```

Here is the output:
```
DNS:ec2-123-70-16-178.eu-west-1.compute.amazonaws.com
DNS:juju-b28639-3
DNS:juju-b28639-3.lxd
DNS:main-2.dca
IP Address:10.122.32.68
IP Address:123.70.16.178
```

lib/charms/mongodb/v1/mongodb_tls.py
```diff
@@ -302,21 +315,93 @@ def _on_certificate_expiring(self, event: CertificateExpiringEvent) -> None:
...
+        try:
+            sans_lines = self.get_sans_from_host(internal)
+        except (subprocess.CalledProcessError, ExecError) as e:
+            logger.error("failed to get sans from host, error: %s", e.stderr)
+            raise e
+
+        line = ""
+        for line in sans_lines:
+            if "DNS" in line or "IP" in line:
+                break
```
Mehdi-Bendriss: `line` is never set

Mehdi-Bendriss (commented):
Thanks Mia - I have a couple of comments

lib/charms/mongodb/v1/mongodb_tls.py
MiaAltieri: python works such that when this loop finishes, `line` is evaluated to be the last value that it was in the for loop

if you're not sure run
```
for i in range(0,10)
     if i == 3:
         break

print(i)
```

lib/charms/mongodb/v1/mongodb_tls.py
MiaAltieri: since there is a lot of contention with the processing here I will try to get Neha's solution working so we can perform automatic processing :)

lib/charms/mongodb/v1/mongodb_tls.py
MiaAltieri: this was an excellent suggestion - thank you Neha ⭐

Gu1nness (dismissed):

lib/charms/mongodb/v1/mongodb_tls.py
```diff
@@ -302,21 +314,52 @@ def _on_certificate_expiring(self, event: CertificateExpiringEvent) -> None:
...
         """Create a list of DNS names for a MongoDB unit.
 
         Returns:
             A list representing the hostnames of the MongoDB unit.
         """
+        sans = {}
+
         unit_id = self.charm.unit.name.split("/")[1]
-        return [
+        sans[SANS_DNS_KEY] = [
```
Mehdi-Bendriss: nit: you could build the dictionary directly with both `SANS_DNS_KEY` and `SANS_IPS_KEY` instead of diving them into 3 steps

Mehdi-Bendriss (dismissed):

Gu1nness:
```
  File "/var/lib/juju/agents/unit-shard-one-5/charm/lib/charms/mongodb/v1/shards_interface.py", line 714, in _on_relation_changed
    if tls_enabled and self.charm.tls.waiting_for_certs():
AttributeError: 'MongoDBTLS' object has no attribute 'waiting_for_certs'
```
There's a small miss there it seems :)

Mehdi-Bendriss (approved):

Gu1nness (approved):

PR #481, opened 2024-09-10
[DPE-5374] Move agnostic upgrade to upgrade_helpers

Gu1nness:
## Issue

* The actual `Upgrade` class is substrate agnostic and should be reused.
* This will allow testing the pre-upgrade checks on mongodb-k8s charm.

## Solution

* Extract in the upgrade_helpers.


MiaAltieri (commented):
I like this and I acknowledge that the upgrade class is substrate agnostic. 

But is it implementation agnostic. i.e. if we go with v3 on MongoDB K8s but v2 on MongoDB VM with this change still stand?

Gu1nness:
> But is it implementation agnostic. i.e. if we go with v3 on MongoDB K8s but v2 on MongoDB VM with this change still stand?

This code does not care about the actual upgrade but only the checks.
If V3 needs some update, we can always overload the functions that would need a fix.
This only ports the minimal amount of code to make the pre upgrade hooks work :)

Mehdi-Bendriss (approved):

MiaAltieri (approved):

MiaAltieri:
My concern is that functions in `AbstractUpgrade ` won't be used in K8s , since ideally those upgrade specific functions will be in a lib for v3. But I trust you!

PR #483, opened 2024-09-11
[DPE-5235] add changes necessary for mongos-k8s ext connections

MiaAltieri:
## Issue
mongos k8s charm can have errors while retrieving hosts, due to race conditions in units 

## Solution
library should be flexible to handle this scenario

### DIFF
--- a/src/exceptions.py
+++ b/src/exceptions.py
@@ -34,7 +34,3 @@ class SecretAlreadyExistsError(MongoSecretError):
 
 class NotConfigServerError(Exception):
     """Raised when an operation is performed on a component that is not a config server."""
-
-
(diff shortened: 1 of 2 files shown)


src/exceptions.py
Gu1nness, line 1: Maybe this exception should live in mongodb provider to avoid useless code in MongoDB VM charm??

Gu1nness (commented):
See comment,  I think the exception should be defined in the library directly :)

src/exceptions.py
MiaAltieri: good shout!

Mehdi-Bendriss (approved):

Gu1nness (approved):

PR #488, opened 2024-09-17
[DPE-5371]: Storage reuse on a different cluster

Gu1nness:
## Issue
 * Storage reuse doesn't work if we use a new application

## Solution

There are multiple problems to face when reusing storage in a new application:
* We need to be able to detect that we are using a new application and not adding back to the same replica set.
* We cannot suppose we have kept all users passwords so we need to change it.
* We might have to change the replica set name.
* We need in any case to reconfigure the replicaset.

The solution chosen here is the following:
 * We store on the storage volume a file containing a random string
 * This string is also stored as a secret (because we have access to app secrets in `install` event, which is not the case for the app peer databag)
 * If we detect that, after installing charmed mongodb, we'll start mongodb in a degraded mode: no replicaset, no auth validation. This allows us to patch the deployment
 * Then we init the replica set, but we need to use authentication here in order to ensure that it works, because we already have the users in this case.
 * We add one more optional reconfigure : for at least two situations, it can happen that all IP changes. This leads to the replica set being fully broken, no host being able to find the other members of the replica set. In order to fix this, a new method to forcefully reconfigure the replica set is added in a specific case : None of the IPs in the mongoDB replica set config is matching the IPs in the replica set config in the Databag. This is achieved by opening a standalone connection to the node (which doesn't require any server selection in the replicaset) and getting the `config` (not the `status` which does requires to be connected to other nodes).

The (not so big) drawback:
 * Some of the most recent data might be lost (sic). WiredTiger writes the snapshots to disk every 60 seconds, which means some of the last data might be lost.
 * I tried to go to different solutions (update users and rename replica set only) in order to avoid that, which ended up in: Core dump of mongodb (yes…), MongoDB restarting in loop because of some config collections being broken. I haven't been able to figure out what the issue was, so I decided to go with this in-between solution.
 * I find this is an acceptable solution granted that:
  * If we're reusing storage in a new application willingly, we can leave time for snapshot to be written on disk
  * If we reusing storage after a crash, it's an acceptable loss to lose the very latest data.

## Implements
 * Storage reuse on the same cluster (scale down, scale to zero).
 * Storage reuse on a different cluster with same app name.
 * Storage reuse on a different cluster with different app name.
 * Restore after a full cluster crash.

## Suggestions for the future

* A big part of the issues we have come from the fact that we use IPs and no DNS inside the cluster.
Having a DNS in the cluster would help as network crash or machines restarting wouldn't change the name of each unit, and the cluster would be able to survive IP changes.
* An integration to external secrets could help make this easier: deploying a new application connected to external secrets would help not losing the secrets in the first application and just restarting as before, with "only" big reconfigurations of hosts.

### DIFF
--- a/lib/charms/mongodb/v0/mongo.py
+++ b/lib/charms/mongodb/v0/mongo.py
@@ -31,7 +31,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 1
+LIBPATCH = 2
 
 ADMIN_AUTH_SOURCE = "authSource=admin"
 SYSTEM_DBS = ("admin", "local", "config")
@@ -270,6 +270,11 @@ class MongoConnection:
             ]
         )
 
+    def get_all_users(self) -> Set[str]:
+        """Get all users, including the three charmed managed users."""
+        users_info = self.client.admin.command("usersInfo")
+        return {user_obj["user"] for user_obj in users_info["users"]}
+
     def get_databases(self) -> Set[str]:
         """Return list of all non-default databases."""
         databases = self.client.list_database_names()
@@ -280,3 +285,7 @@ class MongoConnection:
         if database in SYSTEM_DBS:
             return
         self.client.drop_database(database)
+
+    def drop_local_database(self):
+        """DANGEROUS: Drops the local database."""
+        self.client.drop_database("local")
--- a/lib/charms/mongodb/v1/helpers.py
+++ b/lib/charms/mongodb/v1/helpers.py
@@ -23,7 +23,7 @@ LIBAPI = 1
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 10
+LIBPATCH = 11
 
 # path to store mongodb ketFile
 KEY_FILE = "keyFile"
@@ -176,6 +176,8 @@ def get_mongod_args(
     auth: bool = True,
     snap_install: bool = False,
     role: str = "replication",
+    machine_ip: str = "127.0.0.1",
+    degraded: bool = False,
 ) -> str:
     """Construct the MongoDB startup command line.
 
@@ -186,6 +188,24 @@ def get_mongod_args(
     full_conf_dir = f"{MONGODB_SNAP_DATA_DIR}{CONF_DIR}" if snap_install else CONF_DIR
     logging_options = _get_logging_options(snap_install)
     audit_log_settings = _get_audit_log_settings(snap_install)
+    if degraded:
+        cmd = [
+            # Only bind to local IP
+            f"--bind_ip {machine_ip}",
+            # db must be located within the snap common directory since the
+            # snap is strictly confined
+            f"--dbpath={full_data_dir}",
+            # for simplicity we run the mongod daemon on shards, configsvrs,
+            # and replicas on the same port
+            f"--port={Config.MONGODB_PORT}",
+            "--setParameter processUmask=037",  # required for log files perminission (g+r)
+            "--logRotate reopen",
+            "--logappend",
+            logging_options,
+            "--setParameter enableLocalhostAuthBypass=0",
+            "\n",
+        ]
+        return " ".join(cmd)
     cmd = [
         # bind to localhost and external interfaces
         "--bind_ip_all",
@@ -264,6 +284,15 @@ def generate_keyfile() -> str:
     """
     choices = string.ascii_letters + string.digits
     return "".join([secrets.choice(choices) for _ in range(1024)])
+
+
+def generate_lock_hash() -> str:
+    """Lock hash used to check if we are reusing storage in a different context or not.
+
+    Returns:
+       An 8 character random hexadecimal string.
+    """
+    return secrets.token_hex(8)
 
 
 def copy_licenses_to_unit():
--- a/lib/charms/mongodb/v1/helpers.py
+++ b/lib/charms/mongodb/v1/helpers.py
@@ -41,6 +41,9 @@ DATA_DIR = "/var/lib/mongodb"
 LOG_DIR = "/var/log/mongodb"
 CONF_DIR = "/etc/mongod"
 MONGODB_LOG_FILENAME = "mongodb.log"
+
+LOCALHOST = "127.0.0.1"
+
 logger = logging.getLogger(__name__)
 
 
@@ -171,13 +174,39 @@ def get_mongos_args(
     return " ".join(cmd)
 
 
+def get_degraded_mongod_args(snap_install: bool = False):
+    """Builds a degraded MongoDB startup command line.
+
+    This degraded command line starts MongoDB with minimal configuration,
+    only binds to localhost, without replica set configuration and has no
+    auth validation in order to be able to update the users and local database.
+    """
+    full_data_dir = f"{MONGODB_COMMON_DIR}{DATA_DIR}" if snap_install else DATA_DIR
+    logging_options = _get_logging_options(snap_install)
+    cmd = [
+        # Only bind to local IP
+        f"--bind_ip {LOCALHOST}",
+        # db must be located within the snap common directory since the
+        # snap is strictly confined
+        f"--dbpath={full_data_dir}",
+        # for simplicity we run the mongod daemon on shards, configsvrs,
+        # and replicas on the same port
+        f"--port={Config.MONGODB_PORT}",
+        "--setParameter processUmask=037",  # required for log files perminission (g+r)
+        "--logRotate reopen",
+        "--logappend",
+        logging_options,
+        "--setParameter enableLocalhostAuthBypass=0",
+        "\n",
+    ]
+    return " ".join(cmd)
+
+
 def get_mongod_args(
     config: MongoConfiguration,
     auth: bool = True,
     snap_install: bool = False,
     role: str = "replication",
-    machine_ip: str = "127.0.0.1",
-    degraded: bool = False,
 ) -> str:
     """Construct the MongoDB startup command line.
 
@@ -188,24 +217,6 @@ def get_mongod_args(
     full_conf_dir = f"{MONGODB_SNAP_DATA_DIR}{CONF_DIR}" if snap_install else CONF_DIR
     logging_options = _get_logging_options(snap_install)
     audit_log_settings = _get_audit_log_settings(snap_install)
-    if degraded:
-        cmd = [
-            # Only bind to local IP
-            f"--bind_ip {machine_ip}",
-            # db must be located within the snap common directory since the
-            # snap is strictly confined
-            f"--dbpath={full_data_dir}",
-            # for simplicity we run the mongod daemon on shards, configsvrs,
-            # and replicas on the same port
-            f"--port={Config.MONGODB_PORT}",
-            "--setParameter processUmask=037",  # required for log files perminission (g+r)
-            "--logRotate reopen",
-            "--logappend",
-            logging_options,
-            "--setParameter enableLocalhostAuthBypass=0",
-            "\n",
-        ]
-        return " ".join(cmd)
     cmd = [
         # bind to localhost and external interfaces
         "--bind_ip_all",
@@ -288,6 +299,12 @@ def generate_keyfile() -> str:
 
 def generate_lock_hash() -> str:
     """Lock hash used to check if we are reusing storage in a different context or not.
+
+    This string is written in the storage of all members of the replica set in
+    a file, and also as a secret.
+    Upon starting a new unit, it will check in the storage and in the secret to
+    compare and decide if we're reusing storage and if yes, in which case we
+    are.
 
     Returns:
        An 8 character random hexadecimal string.
--- a/src/charm.py
+++ b/src/charm.py
@@ -788,6 +788,12 @@ class MongodbOperatorCharm(CharmBase):
                     return
 
         try:
+            self._reconfigure_replica_set()
+        except PyMongoError as e:
+            logger.info(f"Could not reconfigure replica set due to {e}")
+            return
+
+        try:
             self.perform_self_healing(event)
         except ServerSelectionTimeoutError:
             # health checks that are performed too early will fail if the hasn't elected a primary
--- a/src/charm.py
+++ b/src/charm.py
@@ -18,10 +18,12 @@ from charms.mongodb.v0.mongodb_secrets import SecretCache, generate_secret_label
 from charms.mongodb.v0.set_status import MongoDBStatusHandler
 from charms.mongodb.v1.helpers import (
     KEY_FILE,
+    LOCALHOST,
     TLS_EXT_CA_FILE,
     TLS_EXT_PEM_FILE,
     TLS_INT_CA_FILE,
     TLS_INT_PEM_FILE,
+    add_args_to_env,
     copy_licenses_to_unit,
     generate_keyfile,
     generate_lock_hash,
@@ -264,7 +266,7 @@ class MongodbOperatorCharm(CharmBase):
 
     @property
     def _is_storage_from_different_cluster(self) -> bool:
-        """Checks if we are reusing storage from a different cluster."""
+        """Returns True if we are reusing storage from a different cluster."""
         lock_hash = self.lock_hash  # Avoid duplicate read of the file
         return lock_hash != UNDEFINED and lock_hash != self.databag_lock_hash
 
@@ -275,7 +277,11 @@ class MongodbOperatorCharm(CharmBase):
 
     @property
     def standalone_config(self) -> MongoConfiguration:
-        """Generates a MongoConfiguration object for local authenticated standalone connection."""
+        """Generates a MongoConfiguration object for local authenticated standalone connection.
+
+        This allows to connect to a mongodb node which is in a replicaset but
+        without going through replicaset checks.
+        """
         return self._get_mongodb_config_for_user(
             OperatorUser, {self.unit_host(self.unit)}, standalone=True
         )
@@ -284,13 +290,6 @@ class MongodbOperatorCharm(CharmBase):
     def mongodb_config(self) -> MongoConfiguration:
         """Generates a MongoConfiguration object for this deployment of MongoDB."""
         return self._get_mongodb_config_for_user(OperatorUser, set(self.app_hosts))
-
-    @property
-    def single_host_config(self) -> MongoConfiguration:
-        """Generates a MongoConfiguration object for operator user on a single host."""
-        return self._get_mongodb_config_for_user(
-            OperatorUser, standalone=True, hosts={self.unit_host(self.unit)}
-        )
 
     @property
     def monitor_config(self) -> MongoConfiguration:
@@ -418,6 +417,13 @@ class MongodbOperatorCharm(CharmBase):
         # If we are a replica and the lock hash doesn't match
         if self.is_role(Config.Role.REPLICATION) and self._is_storage_from_different_cluster:
             self._fix_mongodb_for_reuse()
+        elif self.is_sharding_component() and self._is_storage_from_different_cluster:
+            self.status.set_and_share_status(
+                BlockedStatus(
+                    f"Reusing storage from a different ReplicaSet is not allowed on a {self.role}."
+                )
+            )
+            return
 
         # Construct the mongod startup commandline args for systemd and reload the daemon.
         update_mongod_service(
@@ -454,9 +460,15 @@ class MongodbOperatorCharm(CharmBase):
             )
 
     def __fix_users_for_reuse(self, direct_mongo: MongoDBConnection) -> None:
-        """Fix the users in the DB for storage reuse."""
+        """Fix the users in the DB for storage reuse.
+
+        Context: If we are reusing storage from a different Juju application,
+        the MongoDB Users need to be updated with new passwords because we
+        cannot assume that the juju operator has kept the credentials.
+        This method will update the credentials of the existing users that
+        should be patched in this case.
+        """
         users = direct_mongo.get_all_users()
-        breakpoint()
         # Update operator password since operator is present.
         if self.unit.is_leader() and OperatorUser.get_username() in users:
             logger.info("[Recovery] Update operator password (leader).")
@@ -477,7 +489,6 @@ class MongodbOperatorCharm(CharmBase):
             direct_mongo.drop_user(BackupUser.get_username())
         if MonitorUser.get_username() in users:
             direct_mongo.drop_user(MonitorUser.get_username())
-        pass
 
     def _fix_mongodb_for_reuse(self) -> None:
         """If we are reusing storage, we need to fix multiple things in the mongodb data.
@@ -491,21 +502,25 @@ class MongodbOperatorCharm(CharmBase):
          * Delete the local database (contains the replica set local information) on all units.
         """
         logger.debug("Fixing MongoDB for reuse.")
-        self.status.set_and_share_status(MaintenanceStatus("Updating MongoDB for storage reuse."))
+        self.status.set_and_share_status(MaintenanceStatus("Reusing provided storage."))
         try:
             # Update with degraded configuration.
             # Dangerous, this is only allowed here because we need to fix the DB configuration.
+            logger.info(
+                "[Recovering] Updating MongoDB to degraded configuration to update users passwords."
+            )
             update_mongod_service(
-                machine_ip="127.0.0.1",
+                machine_ip=LOCALHOST,
                 config=self.mongodb_config,
                 role=self.role,
                 degraded=True,
             )
             # Start the charm services in degraded mode.
+            logger.info("[Recovering] Starting the charm services in degraded mode.")
             self.start_charm_services()
             # Perform data modification
             if self.unit.is_leader():
-                # We can create only if we are the leader.
+                # We can only create operator user password if we are the leader.
                 self._check_or_set_user_password(OperatorUser)
             with MongoDBConnection(
                 self.mongodb_config, uri="localhost", direct=True
@@ -521,14 +536,18 @@ class MongodbOperatorCharm(CharmBase):
                             raise NotReadyError
                 self.__fix_users_for_reuse(direct_mongo)
                 # Drop local database
-                logger.info("Dropping local database.")
+                logger.info("[Recovering] Dropping local database.")
                 direct_mongo.drop_local_database()
         except Exception as err:
             logger.error(f"Error encountered while updating database config: {err}")
             raise
         finally:
-            # Whatever happens here, we stop the charm services.
+            # It is important that the charm services are stopped here to
+            # prevent the application running in degraded mode.
+            logging.info("[Recovering] Stopping degraded charm services")
             self.stop_charm_services()
+            # Reset start args to empty string
+            add_args_to_env("MONGOD_ARGS", "")
 
     def _on_start(self, event: StartEvent) -> None:
         """Enables MongoDB service and initialises replica set.
@@ -625,13 +644,8 @@ class MongodbOperatorCharm(CharmBase):
         self._connect_mongodb_exporter()
         self._connect_pbm_agent()
 
-        if (
-            self.db_initialised
-            and self.is_role(Config.Role.REPLICATION)
-            and self.lock_hash != self.databag_lock_hash
-        ):
-            # We write the lock file if it wasn't written yet.
-            self.lock_hash = self.databag_lock_hash
+        # Update the lock hash on the file system if needed.
+        self._update_lock_hash()
 
         # only leader should configure replica set and app-changed-events can trigger the relation
         # changed hook resulting in no JUJU_REMOTE_UNIT if this is the case we should return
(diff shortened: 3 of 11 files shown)


lib/charms/mongodb/v0/mongo.py
```diff
@@ -270,6 +270,11 @@ def get_users(self) -> Set[str]:
             ]
         )
 
+    def get_all_users(self) -> Set[str]:
+        """Get all users, including the three charmed managed users."""
+        users_info = self.client.admin.command("usersInfo")
+        return {user_obj["user"] for user_obj in users_info["users"]}
```
MiaAltieri, line 276: keep in mind: relation users can create other users

Context: When we remove  MongoDB Application it should result in a relation-broken and remove all client-users
Question: When we scale down MongoDB Application to 0, will it result in a relation-broken event and remove client-users

lib/charms/mongodb/v0/mongo.py
```diff
@@ -280,3 +285,7 @@ def drop_database(self, database: str):
         if database in SYSTEM_DBS:
             return
         self.client.drop_database(database)
+
+    def drop_local_database(self):
+        """DANGEROUS: Drops the local database."""
+        self.client.drop_database("local")
```
MiaAltieri, line 291: Is there something we can calculate / store in app databag to verify whether it is safe to drop the local db? then we could add a guardrail by raising a custom exception

lib/charms/mongodb/v1/helpers.py
```diff
@@ -176,6 +176,8 @@ def get_mongod_args(
     auth: bool = True,
     snap_install: bool = False,
     role: str = "replication",
+    machine_ip: str = "127.0.0.1",
```
MiaAltieri: nit `"127.0.0.1"` as a const LOCAL_HOST

lib/charms/mongodb/v1/helpers.py
```diff
@@ -186,6 +188,24 @@ def get_mongod_args(
     full_conf_dir = f"{MONGODB_SNAP_DATA_DIR}{CONF_DIR}" if snap_install else CONF_DIR
     logging_options = _get_logging_options(snap_install)
     audit_log_settings = _get_audit_log_settings(snap_install)
+    if degraded:
+        cmd = [
+            # Only bind to local IP
+            f"--bind_ip {machine_ip}",
```
MiaAltieri: Woukd bw good to enforce this with a constant

lib/charms/mongodb/v1/helpers.py
```diff
@@ -176,6 +176,8 @@ def get_mongod_args(
     auth: bool = True,
     snap_install: bool = False,
     role: str = "replication",
+    machine_ip: str = "127.0.0.1",
+    degraded: bool = False,
```
MiaAltieri: Up to you:

but what do you think of having two functions? one for degraded args and one for normal? up to you.

reason: 
i am worried this function is doing too much

lib/charms/mongodb/v1/helpers.py
```diff
@@ -186,6 +188,24 @@ def get_mongod_args(
     full_conf_dir = f"{MONGODB_SNAP_DATA_DIR}{CONF_DIR}" if snap_install else CONF_DIR
     logging_options = _get_logging_options(snap_install)
     audit_log_settings = _get_audit_log_settings(snap_install)
+    if degraded:
```
MiaAltieri: if you decide to keep this as one function then can you please add a docstring with your implemenation specific knowledge i.e. reasoning + MongoDB specifics? might be easy enough to copy+paste from your PR message

lib/charms/mongodb/v1/helpers.py
```diff
@@ -186,6 +188,24 @@ def get_mongod_args(
...
+        cmd = [
+            # Only bind to local IP
+            f"--bind_ip {machine_ip}",
+            # db must be located within the snap common directory since the
+            # snap is strictly confined
+            f"--dbpath={full_data_dir}",
+            # for simplicity we run the mongod daemon on shards, configsvrs,
+            # and replicas on the same port
+            f"--port={Config.MONGODB_PORT}",
```
MiaAltieri: QQ: this PR *only* works for bare replica sets correct?

lib/charms/mongodb/v1/helpers.py
```diff
@@ -186,6 +188,24 @@ def get_mongod_args(
...
+            "--setParameter processUmask=037",  # required for log files perminission (g+r)
+            "--logRotate reopen",
+            "--logappend",
+            logging_options,
+            "--setParameter enableLocalhostAuthBypass=0",
+            "\n",
```
MiaAltieri: seems that several of these are re-used by the non-degraded command - mind merging to remove duplication? (that is if you keep these as one function)

lib/charms/mongodb/v1/helpers.py
```diff
@@ -266,6 +286,15 @@ def generate_keyfile() -> str:
     return "".join([secrets.choice(choices) for _ in range(1024)])
 
 
+def generate_lock_hash() -> str:
+    """Lock hash used to check if we are reusing storage in a different context or not.
+
```
MiaAltieri, line 302: mind adding some context of how this is used? i.e. the file that is written/read at install

src/charm.py
```diff
@@ -99,6 +101,9 @@
 class MongodbOperatorCharm(CharmBase):
     """Charm the service."""
 
+    # Lock Hash descriptor
```
MiaAltieri, line 106: is this comment necessary? If it is can you expand on what purpose `lock_hash = LockHashHandler()` serves?

src/charm.py
```diff
@@ -257,16 +262,36 @@ def remote_mongodb_config(self, hosts, replset=None, standalone=None) -> MongoCo
             OperatorUser, hosts, replset=replset, standalone=standalone
         )
 
+    @property
+    def _is_storage_from_different_cluster(self) -> bool:
+        """Checks if we are reusing storage from a different cluster."""
```
MiaAltieri:
```suggestion
        """Returns True if we are reusing storage from a different cluster."""
```

src/charm.py
```diff
@@ -257,16 +262,36 @@ def remote_mongodb_config(self, hosts, replset=None, standalone=None) -> MongoCo
...
+        return lock_hash != UNDEFINED and lock_hash != self.databag_lock_hash
+
     @property
     def mongo_config(self) -> MongoConfiguration:
         """Returns a MongoConfiguration object for shared libs with agnostic mongo commands."""
         return self.mongodb_config
 
+    @property
+    def standalone_config(self) -> MongoConfiguration:
+        """Generates a MongoConfiguration object for local authenticated standalone connection."""
```
MiaAltieri: can you add to the docstring when this is useful (in a MongoDB context not charming)

src/charm.py
```diff
@@ -257,16 +262,36 @@ def remote_mongodb_config(self, hosts, replset=None, standalone=None) -> MongoCo
...
+        )
+
     @property
     def mongodb_config(self) -> MongoConfiguration:
         """Generates a MongoConfiguration object for this deployment of MongoDB."""
         return self._get_mongodb_config_for_user(OperatorUser, set(self.app_hosts))
 
+    @property
+    def single_host_config(self) -> MongoConfiguration:
+        """Generates a MongoConfiguration object for operator user on a single host."""
```
MiaAltieri: same here for docstring and clarity based on MongoDB context

src/charm.py
```diff
@@ -257,16 +262,36 @@ def remote_mongodb_config(self, hosts, replset=None, standalone=None) -> MongoCo
...
+            OperatorUser, {self.unit_host(self.unit)}, standalone=True
+        )
+
     @property
     def mongodb_config(self) -> MongoConfiguration:
         """Generates a MongoConfiguration object for this deployment of MongoDB."""
         return self._get_mongodb_config_for_user(OperatorUser, set(self.app_hosts))
 
+    @property
+    def single_host_config(self) -> MongoConfiguration:
```
MiaAltieri:
```suggestion
    def unit_host_config(self) -> MongoConfiguration:
```

not sure on the naming of this, but should be updated to reflect that it using the current unit's host

src/charm.py
```diff
@@ -390,6 +415,10 @@ def _on_install(self, event: InstallEvent) -> None:
             self.status.set_and_share_status(BlockedStatus("Could not install MongoDB"))
             return
 
+        # If we are a replica and the lock hash doesn't match
+        if self.is_role(Config.Role.REPLICATION) and self._is_storage_from_different_cluster:
+            self._fix_mongodb_for_reuse()
+
```
MiaAltieri: can we raise if we are a shard/config-server and the lock has doesn't match

src/charm.py
```diff
@@ -424,6 +453,83 @@ def _on_config_changed(self, event: ConfigChangedEvent) -> None:
                 f"Migration of sharding components not permitted, revert config role to {self.role}"
             )
 
+    def __fix_users_for_reuse(self, direct_mongo: MongoDBConnection) -> None:
+        """Fix the users in the DB for storage reuse."""
```
MiaAltieri: please add Juju/MongoDB context in doc string

src/charm.py
```diff
@@ -424,6 +453,83 @@ def _on_config_changed(self, event: ConfigChangedEvent) -> None:
...
+        users = direct_mongo.get_all_users()
+        breakpoint()
```
MiaAltieri:
```suggestion
```

src/charm.py
```diff
@@ -424,6 +453,83 @@ def _on_config_changed(self, event: ConfigChangedEvent) -> None:
                 f"Migration of sharding components not permitted, revert config role to {self.role}"
             )
 
+    def __fix_users_for_reuse(self, direct_mongo: MongoDBConnection) -> None:
+        """Fix the users in the DB for storage reuse."""
+        users = direct_mongo.get_all_users()
```
MiaAltieri, line 471: please see the comment/warning in the mongodb lib about `get_all_users` - I am worried that there could be uninteded consequences

src/charm.py
```diff
@@ -424,6 +453,83 @@ def _on_config_changed(self, event: ConfigChangedEvent) -> None:
...
+                OperatorUser.get_username(), direct_mongo.config.password
+            )
+        # Operator user not present in users, need to create it.
+        elif self.unit.is_leader():
+            logger.info("[Recovery] Create missing operator user (leader).")
+            direct_mongo.create_user(self.mongodb_config)
+        # Not leader and operator present, drop it.
+        elif OperatorUser.get_username() in users:
+            logger.info("[Recovering] Dropping operator user (not leader).")
+            direct_mongo.drop_user(OperatorUser.get_username())
```
MiaAltieri, line 485: should we allow non-leaders to execute this?

What happens if a leader and a non-leader execute this code at the same time or in dfferent order

src/charm.py
```diff
@@ -424,6 +453,83 @@ def _on_config_changed(self, event: ConfigChangedEvent) -> None:
...
+        # Drop backup and monitor user if present.
+        logger.info("[Recovering] Dropping users backup and monitor.")
+        if BackupUser.get_username() in users:
+            direct_mongo.drop_user(BackupUser.get_username())
+        if MonitorUser.get_username() in users:
+            direct_mongo.drop_user(MonitorUser.get_username())
+        pass
```
MiaAltieri:
```suggestion
```

src/charm.py
```diff
@@ -424,6 +453,83 @@ def _on_config_changed(self, event: ConfigChangedEvent) -> None:
...
+        This is done before starting (so with a downtime).
+        We need to:
+         * Start mongod in degraded mode (no auth validation, no replicaset
+         specified, bind only to local IP).
+         * Update operator password (on leader) and delete it on other units.
+         * Remove backup and monitor users on all units.
+         * Delete the local database (contains the replica set local information) on all units.
+        """
+        logger.debug("Fixing MongoDB for reuse.")
+        self.status.set_and_share_status(MaintenanceStatus("Updating MongoDB for storage reuse."))
```
MiaAltieri:
```suggestion
        self.status.set_and_share_status(MaintenanceStatus("Reusing provided storage."))
```

reason: status messages should be abstract

src/charm.py
```diff
@@ -424,6 +453,83 @@ def _on_config_changed(self, event: ConfigChangedEvent) -> None:
...
+        try:
+            # Update with degraded configuration.
+            # Dangerous, this is only allowed here because we need to fix the DB configuration.
```
MiaAltieri, line 508: please add a log with the content in the comment, perhaps as a "warning?"

src/charm.py
```diff
@@ -424,6 +453,83 @@ def _on_config_changed(self, event: ConfigChangedEvent) -> None:
...
+            update_mongod_service(
+                machine_ip="127.0.0.1",
+                config=self.mongodb_config,
+                role=self.role,
+                degraded=True,
+            )
+            # Start the charm services in degraded mode.
```
MiaAltieri, line 518:
```suggestion
            # Start the charm services in degraded mode.
            logger.info("Starting the charm services in degraded mode.")
```

src/charm.py
```diff
@@ -424,6 +453,83 @@ def _on_config_changed(self, event: ConfigChangedEvent) -> None:
...
+            self.start_charm_services()
+            # Perform data modification
+            if self.unit.is_leader():
+                # We can create only if we are the leader.
```
MiaAltieri:
```suggestion
                # We can only create operator user if we are the leader.
```

src/charm.py
```diff
@@ -424,6 +453,83 @@ def _on_config_changed(self, event: ConfigChangedEvent) -> None:
...
+                for attempt in Retrying(
+                    stop=stop_after_attempt(10),
+                    wait=wait_fixed(5),
+                    reraise=True,
+                    before=before_log(logger, logging.DEBUG),
+                ):
+                    with attempt:
+                        if not direct_mongo.is_ready:
+                            raise NotReadyError
+                self.__fix_users_for_reuse(direct_mongo)
```
MiaAltieri, line 537: is it necessary for all units to do this?

src/charm.py
```diff
@@ -424,6 +453,83 @@ def _on_config_changed(self, event: ConfigChangedEvent) -> None:
...
+                # Drop local database
+                logger.info("Dropping local database.")
+                direct_mongo.drop_local_database()
+        except Exception as err:
+            logger.error(f"Error encountered while updating database config: {err}")
+            raise
+        finally:
+            # Whatever happens here, we stop the charm services.
```
MiaAltieri:
```
# it is important that the charm services are stopped to prevent that application running in the degraded mode
 self.stop_charm_services()
logger.info("stop running charm in degraded mode")
```

src/charm.py
```diff
@@ -424,6 +453,83 @@ def _on_config_changed(self, event: ConfigChangedEvent) -> None:
...
+            self.stop_charm_services()
+
```
MiaAltieri, line 551: I could have missed something, but I think this function updates the env var MONGODB_ARGS to run the service in degraded mode. I believe if the service gets accidentally started again, that it will run in degraded mode. could this be dangerous?

regardless, it might be nice to reset the start args

src/charm.py
```diff
@@ -519,46 +625,53 @@ def _on_relation_handler(self, event: RelationEvent) -> None:
...
         self._connect_pbm_agent()
 
+        if (
+            self.db_initialised
+            and self.is_role(Config.Role.REPLICATION)
+            and self.lock_hash != self.databag_lock_hash
+        ):
+            # We write the lock file if it wasn't written yet.
+            self.lock_hash = self.databag_lock_hash
+
```
MiaAltieri: now that we have several operations that need to be done by all-units in this handler , I think it might make sense to add a dedicated function for them

src/charm.py
```diff
@@ -519,46 +625,53 @@ def _on_relation_handler(self, event: RelationEvent) -> None:
...
             return
 
+        # If needed, reconfigure replicaset (eg : network cut)
+        try:
+            self._reconfigure_replica_set()
+            self.__add_members_to_replicaset(event)
+        except NotReadyError:
+            self.status.set_and_share_status(WaitingStatus("waiting to reconfigure replica set"))
+            logger.error("Deferring reconfigure: another member doing sync right now")
+            event.defer()
```
MiaAltieri, line 663: add a return here - reasoning is that if we expand this function to have more after this try catch, that we will have to add it later (or worse we forget)

src/charm.py
```diff
@@ -519,46 +625,53 @@ def _on_relation_handler(self, event: RelationEvent) -> None:
...
+        except PyMongoError as e:
+            self.status.set_and_share_status(WaitingStatus("waiting to reconfigure replica set"))
+            logger.error("Deferring reconfigure: error=%r", e)
+            event.defer()
+
+    def __add_members_to_replicaset(self, event: RelationEvent):
```
MiaAltieri, line 671: please as a doc string along with what exceptions this function raises

src/charm.py
```diff
@@ -1234,6 +1353,25 @@ def _run_diagnostic_command(self, cmd) -> None:
         except subprocess.CalledProcessError as e:
             logger.error(f"Exception occurred running '{cmd}'\n {e}")
 
+    def _reconfigure_replica_set(self) -> None:
+        """Optionally reconfigure replicaset if needed.
+
+        This should happen only when we are reusing storage from one cluster to another cluster,
+        because all IPs have changed and all configurations are broken.
```
MiaAltieri, line 1396: please include in the docstring the exceptions this raises

QQ: does this raise NotReadyException?

src/charm.py
```diff
@@ -1234,6 +1353,25 @@ def _run_diagnostic_command(self, cmd) -> None:
...
+        """
+        if not self.db_initialised:
```
MiaAltieri, line 1401: should we let non-leaders execute this?

src/charm.py
```diff
@@ -1266,21 +1413,27 @@ def _initialise_replica_set(self, event: StartEvent) -> None:
...
                 event.defer()
                 self.status.set_and_share_status(
-                    WaitingStatus("waiting to initialise replica set")
+                    WaitingStatus("[ERR2] waiting to initialise replica set")
                 )
                 return
 
             # replica set initialised properly and ready to go
             self.db_initialised = True
+            self.lock_hash = generate_lock_hash()
```
MiaAltieri, line 1467: I haven't read this yet, but I assume it sets the lock hash across all units a a file?

src/charm.py
```diff
@@ -1266,21 +1413,27 @@ def _initialise_replica_set(self, event: StartEvent) -> None:
                 )
                 event.defer()
                 self.status.set_and_share_status(
-                    WaitingStatus("waiting to initialise replica set")
+                    WaitingStatus("[ERR1] waiting to initialise replica set")
```
MiaAltieri: I assume ERR1 + ERR2 were from local debugging? 

Let's remove them here. But also I can see why they are useful for debugging, perhaps we can have some more informative logs to help debug in future?

src/charm.py
```diff
@@ -1507,6 +1660,7 @@ def _generate_relation_departed_key(rel_id: int) -> str:
     @property
     def _is_removing_last_replica(self) -> bool:
         """Returns True if the last replica (juju unit) is getting removed."""
+        logger.error(f"{self.app.planned_units() = } | {len(self.peers_units) = }")
```
MiaAltieri: I think this can be removed. If you disagree can you add some more context to it and make it a logger.info?

src/lock_hash.py
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/env python3
+"""Charm code for MongoDB service."""
+# Copyright 2024 Canonical Ltd.
+# See LICENSE file for licensing details.
+
```
MiaAltieri: nit
```
#!/usr/bin/env python3
# Copyright 2024 Canonical Ltd.
# See LICENSE file for licensing details.
"""TODO something about this file and lock hash"""
```

src/lock_hash.py
```diff
@@ -0,0 +1,41 @@
...
+logger = getLogger(__name__)
+
+HASH_KEY = "lockhash"
+LOCK_PATH = Config.MONGODB_DATA_DIR / f".{HASH_KEY}"
+UNDEFINED = "UNDEFINED"
+
+
+class LockHashHandler:
+    """Descriptor class for the lock hash stored in the file."""
```
MiaAltieri: I think it would be VERY useful in the docstring for this had context on why (in the charming sense) this is necessary

src/lock_hash.py
```diff
@@ -0,0 +1,41 @@
...
+
+    def __set__(self, obj: ops.CharmBase, value: str):
```
MiaAltieri, line 34: noob q: why obj and not unit? I would assume this could only be done on the unit?

tests/integration/ha_tests/test_storage.py
```diff
@@ -0,0 +1,343 @@
...
+OTHER_APP_NAME = "mongodb-new"
+TIMEOUT = 1000
+
+
+@pytest.mark.runner(["self-hosted", "linux", "X64", "jammy", "large"])
+@pytest.mark.group(1)
+@pytest.mark.skipif(
+    os.environ.get("PYTEST_SKIP_DEPLOY", False),
+    reason="skipping deploy, model expected to be provided.",
+)
```
MiaAltieri: I believe this can be removed

tests/integration/ha_tests/test_storage.py
```diff
@@ -0,0 +1,343 @@
...
+from .helpers import (
+    APP_NAME,
+    count_writes,
+    reused_storage,
+    stop_continous_writes,
+    storage_id,
+    storage_type,
+)
+
+OTHER_APP_NAME = "mongodb-new"
```
MiaAltieri: `OTHER_MONGODB_APP_NAME = "mongodb-new"`

tests/integration/ha_tests/test_storage.py
```diff
@@ -0,0 +1,343 @@
...
+    app_name = await get_app_name(ops_test)
+    if storage_type(ops_test, app_name) == "rootfs":
+        pytest.skip(
+            "reuse of storage can only be used on deployments with persistent storage not on rootfs deployments"
+        )
+
+    # removing the only replica can be disastrous
+    if len(ops_test.model.applications[app_name].units) < 2:
+        await ops_test.model.applications[app_name].add_unit(count=1)
+        await ops_test.model.wait_for_idle(apps=[app_name], status="active", timeout=TIMEOUT)
```
MiaAltieri: I think this case be removed since the previous test scales the provided deployment?

tests/integration/ha_tests/test_storage.py
```diff
@@ -0,0 +1,343 @@
...
+
+    # check if previous volumes are attached to the units of the new cluster
+    new_storage_ids = []
+    for unit in ops_test.model.applications[new_app_name].units:
+        new_storage_ids.append(storage_id(ops_test, unit.name))
+
+    assert sorted(storage_ids.values()) == sorted(new_storage_ids), "Storage IDs mismatch"
+
+    actual_writes = await count_writes(ops_test, app_name=new_app_name)
+    assert writes_results["number"] == actual_writes
```
MiaAltieri, line 333: same checks as suggested above

MiaAltieri (changes requested):
Excellent work Neha.

You have added a huge value to our project! Left some questions, requested changes, and nits

lib/charms/mongodb/v0/mongo.py
Gu1nness: I need to check but this function is essentially here to get users to update/remove the charm users (`operator`, `backup` and `monitor`) to see which ones were created already or not.
This is here to make it more reliable in case some users were removed/not yet re-introduced in a storage.

lib/charms/mongodb/v0/mongo.py
Gu1nness: I'm not sure.
TBH dropping the local database is not really something that should be done, but it does work.
The big issue behind dropping the local DB is that it stores also the oplog.
I can try to drop less information than that in this call in order to be safer but I'm unsure it will work.
Should I give it a try?

lib/charms/mongodb/v1/helpers.py
Gu1nness: Yet let's do this.

lib/charms/mongodb/v1/helpers.py
Gu1nness: This has been split as two functions.

lib/charms/mongodb/v1/helpers.py
Gu1nness: Yes it does.
This is not something that can be done on sharded clusters because it would need some updates on both sides (config server and shard).
I thought I had specified this in the PR description but I might have forgotten and I'll add it :)

lib/charms/mongodb/v1/helpers.py
Gu1nness: This has been split as two functions.

lib/charms/mongodb/v1/helpers.py
Gu1nness: Sure

src/charm.py
Gu1nness: This was mainly to explain that it's a python descriptor.
If this seems obvious, I can remove the comment, if not, I can add some more comment.
What do you think?

src/charm.py
Gu1nness: Updated.

src/charm.py
Gu1nness: Done.

src/charm.py
Gu1nness: This is the exact same function as before, I'll remove one of the two :)
Miss from my side.

src/charm.py
Gu1nness: This function has been removed :)

src/charm.py
Gu1nness: Sure.
Should I raise or set a blocked status and return to avoid more damage?

src/charm.py
Gu1nness: This call is controlled here since we use it only to know which users are already present in the database.
We're only changing/dropping the administrative users that are already in DB (at most `operator`, `backup` and `monitor`)

src/charm.py
Gu1nness: The leader cannot execute this, the decision tree is:
 * If leader and Operator exists : update password
 * Elif leader : create Operator user
 * elif Operator exists (means we cannot be leader here): drop Operator user.
 
 Keep in mind this is executed locally on the node and not in the replicaset as this function is called while in degraded performance.

src/charm.py
Gu1nness: Yes it is otherwise the users will not be propagated and patched on other nodes.
I tried to avoid doing this but it would end up in failures because credentials were not updated on other nodes.

src/charm.py
Gu1nness: It should be safe because the configuration gets updated afterwards and the only binding IP allowed is localhost in this configuration but you are right, I'll update the code to reset the starting args here.

src/charm.py
Gu1nness: Very nice catch!

src/charm.py
Gu1nness: It does not raise NotReadyException:
 * We don't want to check for sync in progress here as we'll only update in the case none of the IPs in the config are part of the IPs in the databag.

src/charm.py
Gu1nness: Yes and we have to.
This code allows each node to fix it's local replica set configuration.
This is stored in the local database which is not shared among the replica set and this code updates this configuration to add nodes to which it should speak.
If we do not allow non-leader to do that, each node in the replica set will find itself isolated and will not be able to communicate with other nodes as it will not accept incoming configuration from the new primary as it would not be able to decide if there's a primary granted its knowledge of the configuration. This function acts as an oracle that gives knowledge of the other nodes IPs and then allows to reelect a leader in the replicaset.

src/charm.py
Gu1nness: Yes and it's a miss here, I indeed have forgotten to remove those lines which helped me understand what was going on here.

src/charm.py
Gu1nness: The lock_hash attribute is a descriptor.
Writing to `self.lock_hash` will execute the `__set__` function of the `LockHashHandler` class which does two things:
 * Write in the file of the unit
 * Write the lock hash in the secret if we're the leader.

So what happens here is that the leader will generate the lock hash once and write it to it's fs and in the secret.
And then the other nodes will fetch it from the secret and write it on their fs during the `RelationJoined` event.

This works because the relation joined event will fail until `db_initialised` is set to True which means the secret will also be created!

src/charm.py
Gu1nness: Same it's a miss from my side :)

src/lock_hash.py
Gu1nness: Added some basic module documentation

src/lock_hash.py
Gu1nness: This is how data descriptors work in python.
They have to define at least two functions :  `__set__` and `__get__` which are called when accessing the descriptor or setting a value to it. They can define two other methods which are `__delete__` and  `__set_name__`.
The `__set__` function shall have three parameters which are `self`, the object owning the descriptor object in use and the value which is being written. Here it's the charm code that owns this objects (see `self.lock_hash = LockHashHandler()` in `src/charm.py` so it's the `obj` parameter.

`properties` are a very basic usage of descriptors (in the sense that they define a very basic descriptor.

If you want more information, I let you [read this descriptor guide](https://docs.python.org/3/howto/descriptor.html) from Python documentation :)

It seemed easier for me to export this in a descriptor class here in order to reduce the size of the charm class and make it easier to understand.

tests/integration/ha_tests/test_storage.py
Gu1nness: I don't think you have suggested a check above, maybe a comment got lost?

PR #496, opened 2024-10-02
fix: clients need to be able to read + write to socket file

MiaAltieri:
## Issue
clients need to be able to read + write to socket file, but they cannot 

## Solution
update perms

### DIFF
--- a/lib/charms/mongodb/v1/helpers.py
+++ b/lib/charms/mongodb/v1/helpers.py
@@ -23,7 +23,7 @@ LIBAPI = 1
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 11
+LIBPATCH = 12
 
 # path to store mongodb ketFile
 KEY_FILE = "keyFile"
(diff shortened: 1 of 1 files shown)


lib/charms/mongodb/v1/helpers.py
```diff
@@ -118,7 +118,7 @@ def get_mongos_args(
     binding_ips = (
         "--bind_ip_all"
         if external_connectivity
-        else f"--bind_ip {MONGODB_COMMON_DIR}/var/mongodb-27018.sock  --filePermissions 755"
+        else f"--bind_ip {MONGODB_COMMON_DIR}/var/mongodb-27018.sock  --filePermissions 766"
```
Gu1nness:
```suggestion
        else f"--bind_ip {MONGODB_COMMON_DIR}/var/mongodb-27018.sock  --filePermissions 0766"
```

Gu1nness (changes requested):

Mehdi-Bendriss (approved):

Gu1nness (approved):

PR #498, opened 2024-10-02
bump snap to handle issues with perms in mongosh

MiaAltieri:
## Issue
This PR bumps the snap revision to the latest to fix the permission issues on mongosh


Mehdi-Bendriss (dismissed):

Gu1nness:
We need to use stable here instead of edge, snap has been promoted

Mehdi-Bendriss (dismissed):

Mehdi-Bendriss (approved):

PR #507, opened 2024-10-10
[DPE-3911] Update lib so that correct mongos port is used in k8s

MiaAltieri:
## Issue
port in mongos k8s is incorrect - since we don't query the actual charm for its port 

## Solution
add the port in mongos clients


Gu1nness (approved):

welpaolo (approved):

MiaAltieri:
merging - tests are failing due to timeout - and the inserted code is not ran by this charm

PR #508, opened 2024-10-10
feat(backup): Go to blocked if invalid integration

Gu1nness:
## Issue

If integrating s3-integrator to shard, the shard would end up in error state starting from credentials_changed event.

## Solution

In case of invalid integration, we want to go to Blocked state not try to fail the event.

### DIFF
--- a/lib/charms/mongodb/v1/mongodb_backups.py
+++ b/lib/charms/mongodb/v1/mongodb_backups.py
@@ -61,6 +61,10 @@ MONGODB_SNAP_DATA_DIR = "/var/snap/charmed-mongodb/current"
 BACKUP_RESTORE_MAX_ATTEMPTS = 10
 BACKUP_RESTORE_ATTEMPT_COOLDOWN = 15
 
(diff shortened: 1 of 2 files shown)


lib/charms/mongodb/v1/mongodb_backups.py
```diff
@@ -61,6 +61,10 @@
 BACKUP_RESTORE_MAX_ATTEMPTS = 10
 BACKUP_RESTORE_ATTEMPT_COOLDOWN = 15
 
+INVALID_INTEGRATION_STATUS = BlockedStatus(
```
MiaAltieri, line 64: bump libpatch

MiaAltieri (changes requested):

lib/charms/mongodb/v1/mongodb_backups.py
Gu1nness: Done :)

MiaAltieri (approved):

skourta (approved):

PR #509, opened 2024-10-14
fix: Do not integrate an invalid relation

Gu1nness:
## Issue
It is needed to check first if the relation is valid. 

## Solution
The reusable check works with proxy relations (so config-server can get data proxified by mongos), so we first check here that the basic relation is valid.

This should be improved in the future to make it more straightforward.

### DIFF
--- a/lib/charms/mongodb/v1/mongodb_provider.py
+++ b/lib/charms/mongodb/v1/mongodb_provider.py
@@ -16,7 +16,13 @@ from typing import List, Optional, Set
 from charms.data_platform_libs.v0.data_interfaces import DatabaseProvides
 from charms.mongodb.v0.mongo import MongoConfiguration, MongoConnection
 from charms.mongodb.v1.helpers import generate_password
-from ops.charm import CharmBase, EventBase, RelationBrokenEvent, RelationChangedEvent
+from ops.charm import (
+    CharmBase,
+    EventBase,
+    RelationBrokenEvent,
+    RelationChangedEvent,
+    RelationEvent,
+)
 from ops.framework import Object
 from ops.model import Relation
 from pymongo.errors import PyMongoError
@@ -88,10 +94,11 @@ class MongoDBProvider(Object):
         if not self.charm.db_initialised:
             return False
 
-        # This is tricky: the sanity hook checks can pass when the call is
+        # Warning: the sanity_hook_checks can pass when the call is
         # issued by a config-sever because the config-server is allowed to manage the users
         # in MongoDB. This is not well named and does not protect integration of a config-server
-        # to a client application
+        # to a client application. The mongos charm however doesn't care
+        # because it supports only one integration that uses MongoDBProvider.
         if not self.charm.is_role(Config.Role.MONGOS) and not self.charm.is_relation_feasible(
             self.get_relation_name()
         ):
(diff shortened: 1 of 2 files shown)


Mehdi-Bendriss (dismissed):

lib/charms/mongodb/v1/mongodb_provider.py
```diff
@@ -99,8 +103,12 @@ def pass_sanity_hook_checks(self) -> bool:
 
         return True
 
-    def pass_hook_checks(self, event: EventBase) -> bool:
+    def pass_hook_checks(self, event: RelationChangedEvent) -> bool:
```
MiaAltieri: this should be RelationEvent or something of that sort, from what I can tell it also gets called on `relation_broken` events

lib/charms/mongodb/v1/mongodb_provider.py
```diff
@@ -99,8 +103,12 @@ def pass_sanity_hook_checks(self) -> bool:
...
         """Runs the pre-hooks checks for MongoDBProvider, returns True if all pass."""
+        # First, ensure that the relation is valid, useless to do anything else otherwise
+        if not self.charm.is_relation_feasible(event.relation.name):
```
MiaAltieri: I don't think this will work for other charms that use this (namely) `mongos-k8s`, since it does not have this function.

We should either:
1. add a check here that checks if we aren't a mongos charm
OR
2. add a `is_relation_feasible ` - although I don't believe it would have any use in that charm
3.

lib/charms/mongodb/v1/mongodb_provider.py
```diff
@@ -88,6 +88,10 @@ def pass_sanity_hook_checks(self) -> bool:
         if not self.charm.db_initialised:
             return False
 
+        # This is tricky: the sanity hook checks can pass when the call is
+        # issued by a config-sever because the config-server is allowed to manage the users
+        # in MongoDB. This is not well named and does not protect integration of a config-server
+        # to a client application
```
MiaAltieri: I believe the reason of this is that mongos charms do not have a `is_relation_feasible` attribute. So we must protect the call, by first checking what the role is.

I think the comment here should be
```
# it is important to check if the relation is feasible for mongodb charms which support a number of interfaces.
# However, the mongos charm which supports this does not need to check if the relationship is feasible
```

do you think that would improve this check? If not, would adding a function `is_relation_feasible_for_mongodb_charm` improve this bit of code?

```
def is_relation_feasible_for_mongodb_charm(self) -> bool:
         # base case, do not proceed if running on a mongos charm
         if self.charm.is_role(Config.Role.MONGOS) :
                return True
        
         return self.charm.is_relation_feasible( self.get_relation_name())
```

MiaAltieri (changes requested):

lib/charms/mongodb/v1/mongodb_provider.py
Gu1nness: Oh my bad, I'll add the same check as in the other function.

lib/charms/mongodb/v1/mongodb_provider.py
Gu1nness: The issue here is not about mongos in what I talk about in the comment: What I'm saying is that the regular hook check allows config-server to run some `mongodb_provider` code and hence the `get_relation_name` function returns a relation that does not make `is_relation_feasible` check fail. Here, at integration time (because it's a relation we're talking about and not some protected code) we need to allow only the right roles to bind on this relation. I can rephrase however a bit of the comment to make it more clear.

This will be refactored in the single kernel charm because there's an issue with split responsibilities here between what the relation does (lifecycle) and the interactions with mongodb (workload).

lib/charms/mongodb/v1/mongodb_provider.py
Gu1nness: Oh my bad, it's an error in typing indeed.

lib/charms/mongodb/v1/mongodb_provider.py
MiaAltieri: thank you!

MiaAltieri (approved):

Mehdi-Bendriss (approved):

PR #520, opened 2024-12-04
[DPE-6089] Update Terraform to new scheme 

MiaAltieri:
## Issue
We would like an easy way to deploy MongoDB using terraform on lxd and openstack

## Solution
Add 

## Manual Testing
```
terraform init
terraform apply 
```
done on both LXD and on OpenStack


.github/workflows/ci.yaml
```diff
@@ -32,6 +32,54 @@ jobs:
...
+          pushd ./terraform
+          terraform init
+          terraform fmt
+          terraform validate
+          pushd ./tests
+          terraform init
+          terraform fmt
+          terraform validate
+          popd
+          popd
```
Gu1nness: Could this be split in two steps maybe?
Easier to know what broke I think.
You could do something like this in order to avoid also the pushd/popd
```suggestion
      - name: lint charm module
        working-directory: ./terraform
        run: |
          terraform init
          terraform fmt
          terraform validate
      - name: lint test charm module
         working-directory: ./terraform/tests
         run: |
          terraform init
          terraform fmt
          terraform validate
```

.github/workflows/ci.yaml
```diff
@@ -32,6 +32,54 @@ jobs:
...
+      - name: Juju setup
+        run: |
+          sg 'lxd' -c "juju bootstrap 'localhost' --config model-logs-size=10G"
+          juju model-defaults logging-config='<root>=INFO; unit=DEBUG'
+          juju add-model test
+      - name: Terraform deploy
+        run: |
+          pushd ./terraform/tests/
+          terraform apply  -var "model=test" -target null_resource.simple_deployment_juju_wait_deployment -auto-approve
+          popd
```
Gu1nness:
```suggestion
      - name: Terraform deploy
        working-directory: ./terraform/tests/
        run: |
          terraform apply  -var "model=test" -target null_resource.simple_deployment_juju_wait_deployment -auto-approve
```

terraform/main.tf
```diff
@@ -2,15 +2,40 @@
...
-  model = var.model_name
 
   charm {
-    name    = "mongodb"
-    channel = var.channel
-    base    = "ubuntu@22.04"
+    name     = "mongodb"
+    channel  = var.channel
+    revision = var.revision
+    base     = var.base
```
Gu1nness: Should we enforce the base for now since I'm not sure we're supporting more recent bases?

terraform/myplan
Gu1nness, line 1: This is supposed to be here? Or is it a miss?

MiaAltieri: sigh yes 🤦‍♀️

terraform/tests/preamble.tf
```diff
@@ -0,0 +1,10 @@
+resource "null_resource" "preamble" {
+  provisioner "local-exec" {
+    command = <<-EOT
+    sudo snap install juju-wait --classic || true
+    sudo sysctl -w vm.max_map_count=262144 vm.swappiness=0 net.ipv4.tcp_retries2=5
+    EOT
+  }
+
+}
```
marcoppenheimer: **question:** Do you need to do this? I was under the impression this was an opensearch specific requirement.

terraform/outputs.tf
```diff
@@ -33,22 +28,6 @@ output "cluster_endpoint" {
...
-
-output "grafana_dashboard_endpoint" {
-  description = "Name of the endpoint to provide the grafana_dashboard interface."
-  value       = "grafana-dashboard"
-}
-
-output "logging_endpoint" {
-  description = "Name of the endpoint to provide the loki_push_api relation interface."
-  value       = "logging"
-}
```
marcoppenheimer, line 49: **question:** These were there, but are obviously replaced by the `cos-agent` endpoint. 
Why were these there in the first place? For K8s?

terraform/main.tf
```diff
@@ -2,15 +2,40 @@
...
+  storage_directives = var.storage
+
+  lifecycle {
+    precondition {
+      condition     = length(var.machines) == 0 || length(var.machines) == var.units
+      error_message = "Machine count does not match unit count"
+    }
+    precondition {
+      condition     = length(var.storage) == 0 || lookup(var.storage, "count", 0) <= 1
+      error_message = "Only one storage is supported"
```
marcoppenheimer, line 38: **question:** Is this condition true for Mongo? Do you only support 1 storage or can you have multiple?

MiaAltieri: In MongoDB-VM we only support one storage, but in K8s we support two storages.

The reason we don't yet support that in MongoDB VM is because we aren't sure how to support that in the case of an upgrade

terraform/outputs.tf
MiaAltieri: If I am being honest I am not totally sure. The file in the first place was written by a different team

terraform/tests/preamble.tf
MiaAltieri: good eye!

terraform/main.tf
MiaAltieri: Good question! I have removed it for now

terraform/outputs.tf
MiaAltieri: If I had to guess I would say they were there earlier since the grafana-subordinate now integrates to those end points?

terraform/main.tf
marcoppenheimer: Is that also true? We can't have more than one storage PVC on K8s. Or do you mean multiple storage PVCs, but only 1 of each 'named' one?

MiaAltieri: @marcoppenheimer  I am not sure if I follow 🫣 - perhaps a 1:1 would be useful :)

But if I do understand correctly, [we do have two separate named PVCs](https://github.com/canonical/mongodb-k8s-operator/blob/3fc74007efd913ea8a46805df5a70c8f16f8cb26/metadata.yaml#L69-L74) and [this is how we tested it](https://matrix.to/#/!xzmWHtGpPfVCXKivIh:ubuntu.com/$z0zLuT5fasTDqcw22msgBSUYc5dSlD6kwm-JufyiPfA?via=ubuntu.com&via=matrix.org&via=xentonix.net)

Gu1nness (approved):

marcoppenheimer (approved):

Issue #517, opened 2024-11-27
COS / More alerts


gustavosr98:
## Steps to reproduce
Follow tutorial for COS integration

## Expected behavior
A few alerts I would like to see
- Machine is down
- Machine is up but service is down
- Cluster is not writable
- Cluster will not be writable if I lose one more node
- Query latency goes over X ms
- Number of connections is close to max connections limit

And any similar related to **degradation** of to **prevent the system stops working** as expected  

## Actual behavior
No errors in the alerts, just two alerts
![Pasted image 20241107183950](https://github.com/user-attachments/assets/d58569bc-b45a-469b-9a14-cdd94dd8acbf)

## Versions
Juju 3.5.4

```
data-integrator                    active      1  data-integrator           latest/stable   41  no
grafana-agent                      active      3  grafana-agent             latest/edge    299  no       tracing: off
mongodb                            active      3  mongodb                   6/stable       199  no
self-signed-certificates           active      1  self-signed-certificates  latest/stable  155  no
```

MiaAltieri:
Hi @gustavosr98 I have just opened a PR(#521) for these desired alerts. Thanks for taking the time to outline your wishlist! <3

I've gone ahead and added most of them. The ones I did not add in that PR are:
- Machine is down
- Machine is up but service is down
and
- Query latency goes over X ms

The first two I did not add since they are already there (screenshots 1+2) the third I do not believe is possible since I don't think grafana supports alerts based on user provided input (i.e. X), if you have a specific latency that you have in mind let me know and I can implement that for you ASAP

Please note that #1 will be further improved on o11y end since [they are currently undergoing work for it](https://warthogs.atlassian.net/browse/OPENG-2604)

<img width="1190" alt="Screenshot 2024-12-09 at 14 18 40" src="https://github.com/user-attachments/assets/d71550a0-e219-4701-98eb-ba38f6e764fa">


<img width="1188" alt="Screenshot 2024-12-09 at 14 20 04" src="https://github.com/user-attachments/assets/3d66b218-186c-42f1-9111-b28723d299d0">

MiaAltieri:
Seeing that this was merged and solved I will close this issue. Thanks for reporting!

PR #521, opened 2024-12-09
[DPE-6073] add requested alerts

MiaAltieri:
Addressing #517 by adding the following requested alerts:

- Cluster is not writable
- Cluster will not be writable if I lose one more node
- Number of connections is close to max connections limit

along with a few others from the Percona alert rules

## testing

- Cluster is not writable
Screenshot 2024-12-09 at 14 22 58

- Cluster will not be writable if I lose one more node - note this is firing because it was deployed with a single replica, when the replica set is scaled up it goes back to green
Screenshot 2024-12-09 at 14 22 04

- Number of connections is close to max connections limit (80%)
Screenshot 2024-12-09 at 14 23 32

### DIFF
--- a/src/alert_rules/prometheus/percona-mongodb-exporter.yml
+++ b/src/alert_rules/prometheus/percona-mongodb-exporter.yml
@@ -21,10 +21,64 @@ groups:
           description: "MongoDB replica member is not healthy\n  VALUE = {{ $value }}\n  LABELS = {{ $labels }}"
 
       - alert: MongodbReplicationLag
-        expr: 'mongodb_rs_members_optimeDate{member_state="PRIMARY"} - on (set) group_right mongodb_rs_members_optimeDate{member_state="SECONDARY"} > 10'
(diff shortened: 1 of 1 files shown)


src/alert_rules/prometheus/percona-mongodb-exporter.yml
```diff
@@ -21,10 +21,64 @@ groups:
           description: "MongoDB replica member is not healthy\n  VALUE = {{ $value }}\n  LABELS = {{ $labels }}"
 
       - alert: MongodbReplicationLag
-        expr: 'mongodb_rs_members_optimeDate{member_state="PRIMARY"} - on (set) group_right mongodb_rs_members_optimeDate{member_state="SECONDARY"} > 10'
+        expr: '(mongodb_rs_members_optimeDate{member_state="PRIMARY"} - on (set) group_right mongodb_rs_members_optimeDate{member_state="SECONDARY"}) / 1000 > 10'
```
MiaAltieri, line 24: this was updated in the upstream file frpm Percona alerts

delgod (approved):

Gu1nness (approved):

PR #537, opened 2025-01-16
[DPE-6322] add sharding tf

MiaAltieri:
## Issue
SQA would like a repeatable way to deploy our sharded mongodb

## Solution
Deploy a complete sharded cluster with all available integrations.


terraform/modules/sharded_module/preamble.tf
```diff
@@ -0,0 +1,52 @@
...
+}
+
+resource "juju_application" "mongos" {
+  charm {
+    name    = "mongos"
+    channel = "6/stable"
+  }
+  model      = var.model_name
+  depends_on = [null_resource.preamble]
+}
```
MiaAltieri: this file is near identical the the replica set module except for the addition of mongos

.github/workflows/ci.yaml
```diff
@@ -79,11 +79,18 @@ jobs:
...
+        working-directory: ./terraform/modules/replica_set_module
         run: |
-          terraform apply  -var "model_name=test" -target null_resource.simple_deployment_juju_wait_deployment -auto-approve
+          terraform init
+          terraform apply  -var "model_name=test-rs" -target null_resource.simple_deployment_juju_wait_deployment -auto-approve
+      - name: Terraform deploy sharded tests
+        working-directory: ./terraform/modules/sharded_module
+        run: |
+          terraform init
+          terraform apply  -var "model_name=test-sc" -target null_resource.simple_deployment_juju_wait_deployment -auto-approve
```
Gu1nness: Can we make a matrix out of this?

terraform/modules/sharded_module/preamble.tf
```diff
@@ -0,0 +1,52 @@
...
+
+resource "juju_application" "data-integrator" {
+  charm {
+    name    = "data-integrator"
+    channel = "latest/stable"
+  }
+  model      = var.model_name
+  depends_on = [null_resource.preamble]
+}
```
Gu1nness: This will be related to mongos right?
How do we deploy without using data-integrator?

terraform/modules/sharded_module/simple_deployment.tf
```diff
@@ -0,0 +1,165 @@
...
+      app_name = var.shard_one_app_name
+      units    = var.shard_one_units
+      role     = "shard"
+    }
+    "shard-two" = {
+      app_name = var.shard_two_app_name
+      units    = var.shard_two_units
+      role     = "shard"
+    }
+  }
```
Gu1nness: Noob question: If someone wants more shards, they have to edit this file?
There is no way to provide an array of shards instead?

MiaAltieri: Correct, they would have to edit this file. I am not sure if it is possible to provide an array in this case or not.

terraform/modules/sharded_module/preamble.tf
MiaAltieri: > This will be related to mongos right?

Yes

> How do we deploy without using data-integrator?

We can't since mongos-k8s is a subordinate charm. We can use a client app, but thats out of scope here

terraform/modules/sharded_module/simple_deployment.tf
marcoppenheimer: Neha's right, I think here we want to add a `var` in `variables.tf`, let's call it 'topology' or something:

```tf
variable "topology" {
  description = "mongo large deployment topology"
  type        = optional(list(map(string)), {})
  default     = [
    	"config-server" = {
      		app_name = var.config_server_app_name
      		units    = var.config_server_units
      		role     = "config-server"
    	}
}
```

This is really messy, syntax is 100% wrong, but hopefully you get the gist.
Useful examples here - https://discuss.hashicorp.com/t/why-not-arrays-of-objects/54481/4

MiaAltieri: TIL! Thank you both!

Gu1nness (dismissed):

marcoppenheimer (dismissed):
Much more solid, great stuff!

terraform/modules/sharded_module/simple_deployment.tf
Mehdi-Bendriss, line 1: nits: 
1. we should rename this file (`sharded_deployment.tf`)
2. I believe we can get rid of the `_module` suffix in both `replica_set_module, sharded_module`

Mehdi-Bendriss (commented):
Thanks Mia! great work! 
Left a couple of minor comments on files / folders renaming

Gu1nness (approved):

Mehdi-Bendriss (approved):

PR #543, opened 2025-01-29
[DPE-6355] - lib updates for mongos and TLS

MiaAltieri:
## Issue
This solves #87 

The issue was that a race condition was occurring and the mongos charm. Where the mongos charm went into the blocked state when trying to enable TLS and mongos at the same time.

## Solution

1. Do not request a certificate until we have a config server since a configs-server is needed for the subject name in the cert
2. Update the relation event for cluster relation to wait until the sans are correct and the cert is requested, then add the router

### DIFF
--- a/lib/charms/mongodb/v0/config_server_interface.py
+++ b/lib/charms/mongodb/v0/config_server_interface.py
@@ -40,9 +40,7 @@ KEYFILE_KEY = "key-file"
 KEY_FILE = "keyFile"
 HOSTS_KEY = "host"
 CONFIG_SERVER_DB_KEY = "config-server-db"
-MONGOS_SOCKET_URI_FMT = (
-    "%2Fvar%2Fsnap%2Fcharmed-mongodb%2Fcommon%2Fvar%2Fmongodb-27018.sock"
-)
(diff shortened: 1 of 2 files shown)


lib/charms/mongodb/v0/config_server_interface.py
```diff
@@ -426,12 +427,12 @@ def pass_hook_checks(self, event):
             event.defer()
             return False
 
-        if not self.is_ca_compatible():
+        # race condition where mongos cannot start without TLS certificates, but mongos cannot
+        # request TLS certificates without knowing the name of the config-server.
+        if self.is_waiting_to_request_certs():
```
Gu1nness, line 432: We do lose the check on ca compatibility here however.
Shouldn't we do something like  this?
```
if self.is_waiting_to_request_certs:
…
if not self.is_compatible():
…
```

MiaAltieri: Good eye @Gu1nness 

Its actually not useful here, infact **it is harmful**. It blocks the mongos router from requesting its certs (until it requests the certs it won't know the CA) 

This check is done elsewhere in the charm to ensure that mongos reports on a mismatched CA. So the user won't lose this context :)

Gu1nness: Ok, thanks :)

Gu1nness (approved):

Mehdi-Bendriss (approved):
Thank you Mia!

Issue #536, opened 2025-01-16
Immediate relations on sharding + certificates interface breaks cluster


MiaAltieri:
When deploying a sharded cluster of mongodb using terraform, integrations happen very quickly. This can result in race conditions and a broken cluster. 

Specifically these two issues seem to be related:

1. Shards don't update their TLS certs when immediately integrated to both config-server and self-singed certificates 
2. Config-servere is unable to add shards due to issues with TLS

PR #544, opened 2025-01-30
[DPE-6340] fix relation messaging

MiaAltieri:
## Issue
solves #536 

Initially we thought that there was a race condition between shards + config server + tls. But the issue was:
1. the TF didn't correctly specify the role for shards
2. the real bug: the status didn't say "replica cannot use sharding interface"

## Solution
1. fix TF to use correct role for shards
2. update status checking to check for "replica cannot use sharding interface"


Gu1nness (approved):

Mehdi-Bendriss (approved):
Good catch, thanks Mia!

Issue #533, opened 2025-01-14
PBM status is unhelpful/incorrect when missing configurations


MiaAltieri:
When deploying a sharded cluster of mongodb using terraform, integrations happen very quickly. 

During this period the s3 integration is made (but its configurations are not set) , when the PBM status is retrieved inside the charm with `charmed-mongodb.pbm status -o json`, it results in `{"Error":"get status of pitr: unable check PITR config status: get config: missed config"}`, this error  results in the charm showing a status of `PBM error`

when in fact it should result in something more informative. Perhaps a `BlockedStatus("Missing configurations")`

PR #545, opened 2025-01-31
[DPE-6326] pbm status fix

MiaAltieri:
## Issue
Solves #533 

TL;DR - Rather than PBM showing a useful error when it is missing *required* configurations - it just shows `PBM error`

## Solution
Detect these cases and show a useful message

### DIFF
--- a/lib/charms/mongodb/v1/mongodb_backups.py
+++ b/lib/charms/mongodb/v1/mongodb_backups.py
@@ -837,7 +837,7 @@ class MongoDBBackups(Object):
             logger.info("Missing s3 credentials")
             return False
 
-        if provided_configs.get("storage.s3.bucket"):
+        if not provided_configs.get("storage.s3.bucket"):
             logger.info("Missing bucket")
             return False
 
--- a/lib/charms/mongodb/v1/mongodb_backups.py
+++ b/lib/charms/mongodb/v1/mongodb_backups.py
@@ -123,8 +123,12 @@ class MongoDBBackups(Object):
         self.framework.observe(
             self.s3_client.on.credentials_changed, self._on_s3_credential_changed
         )
-        self.framework.observe(self.charm.on.create_backup_action, self._on_create_backup_action)
-        self.framework.observe(self.charm.on.list_backups_action, self._on_list_backups_action)
+        self.framework.observe(
+            self.charm.on.create_backup_action, self._on_create_backup_action
+        )
+        self.framework.observe(
+            self.charm.on.list_backups_action, self._on_list_backups_action
+        )
         self.framework.observe(self.charm.on.restore_action, self._on_restore_action)
 
     def on_s3_relation_joined(self, event: RelationJoinedEvent) -> None:
@@ -224,7 +228,9 @@ class MongoDBBackups(Object):
                 MaintenanceStatus(f"backup started/running, backup id:'{backup_id}'")
             )
             self._success_action_with_info_log(
-                event, action, {"backup-status": f"backup started. backup id: {backup_id}"}
+                event,
+                action,
+                {"backup-status": f"backup started. backup id: {backup_id}"},
             )
         except (subprocess.CalledProcessError, ExecError, Exception) as e:
             self._fail_action_with_error_log(event, action, str(e))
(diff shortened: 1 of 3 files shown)


lib/charms/mongodb/v1/mongodb_backups.py
```diff
@@ -806,3 +819,35 @@ def get_backup_error_status(self, backup_id: str) -> str:
                 return backup.get("error", "")
 
         return ""
+
+    def are_s3_configurations_provided(self) -> bool:
+        """Returns True if not all necessary configurations for s3 are provided.
```
Mehdi-Bendriss:
```suggestion
        """Returns True if all necessary configurations for s3 are provided.
```

lib/charms/mongodb/v1/mongodb_backups.py
```diff
@@ -806,3 +819,35 @@ def get_backup_error_status(self, backup_id: str) -> str:
...
+        https://docs.percona.com/percona-backup-mongodb/reference/configuration-options.html#s3-type-storage-options
+        """
+        if not self.model.get_relation(S3_RELATION):
+            logger.info("No configurations for backups, not relation to s3-charm.")
+            return False
+
+        provided_configs = self._get_pbm_configs()
+        if (
+            "storage.s3.credentials.access-key-id" not in provided_configs
+            or "storage.s3.credentials.secret-access-key" not in provided_configs
```
Mehdi-Bendriss: Safer with the following as we're also testing on the value 
```suggestion
            not provided_configs.get("storage.s3.credentials.access-key-id")
            or not provided_configs.get("storage.s3.credentials.secret-access-key")
```

lib/charms/mongodb/v1/mongodb_backups.py
```diff
@@ -806,3 +819,35 @@ def get_backup_error_status(self, backup_id: str) -> str:
...
+        ):
+            logger.info("Missing s3 credentials")
+            return False
+
+        if "storage.s3.bucket" not in provided_configs:
```
Mehdi-Bendriss: similar comment as above for this and subsequent conditions

lib/charms/mongodb/v1/mongodb_backups.py
```diff
@@ -806,3 +819,35 @@ def get_backup_error_status(self, backup_id: str) -> str:
...
+            logger.info("Missing bucket")
+            return False
+
+        # since we cannot determine whether the user has an AWS or GCP bucket or Minio bucket
+        # send them an info
+        if "storage.s3.region" not in provided_configs:
+            logger.info("Missing region - this is required for AWS and GCP")
+
+        if "storage.s3.endpointUrl" not in provided_configs:
+            logger.info("Missing endpointUrl - this is required for MinIO and GCP")
```
Mehdi-Bendriss, line 850: Can this info not be inferred / deducted from the previous config options?

Mehdi-Bendriss (commented):
Thanks Mia! I have a couple of questions

lib/charms/mongodb/v1/mongodb_backups.py
MiaAltieri: now that I think of it, I think it might be possible to determine which provider we use by parsing the format of the access key and see if its format is that of AWS or GCP. (but that assumes that the user has provided correct credentials - and would be a little hacky.) Lmk what you think here, I trust your vision.

Gu1nness (approved):

lib/charms/mongodb/v1/mongodb_backups.py
Mehdi-Bendriss: Let's address it in another PR

Mehdi-Bendriss (dismissed):

lib/charms/mongodb/v1/mongodb_backups.py
```diff
@@ -806,3 +819,34 @@ def get_backup_error_status(self, backup_id: str) -> str:
...
+            return False
+
+        provided_configs = self._get_pbm_configs()
+        if not provided_configs.get(
+            "storage.s3.credentials.access-key-id"
+        ) or not provided_configs.get("storage.s3.credentials.secret-access-key"):
+            logger.info("Missing s3 credentials")
+            return False
+
+        if provided_configs.get("storage.s3.bucket"):
```
Gu1nness: This should be a `not provided_configs.get…` I guess.

Gu1nness (changes requested):

Mehdi-Bendriss (approved):

Gu1nness (approved):

Issue #525, opened 2024-12-13
Feature Request: PBM does not support TLS with backups


MiaAltieri:
When using an s3 bucket with tls PBM agent fails with the message:
```
  - mongodb/192.168.100.82:27017 [S]: pbm-agent v2.4.0 FAILED status:
      > ERROR with storage: storage check failed with: get S3 object header: RequestError: send request failed
caused by: Head "https://radosgw.pc6a.canonical.com:443/mybucket/.pbm.init": tls: failed to verify certificate: x509: certificate signed by unknown authority
```

This is a known issue and outlined in SSDLC. We should support TLS with PBM, **this is a requirement from field.**

Additionally Charmed MongoDB should provide an error message when PBM is missing TLS configuration

## Steps to reproduce
[Deploy MognoDB in OpenStack](https://docs.google.com/document/d/1Oz4tyBIrFuXVGHFVwpnf7vKPFLSycOHhtGrX9JDu_ok/edit?usp=sharing)

```
openstack endpoint list | grep -E "s3|swift"

# curl the public URL 
curl  https://radosgw.pc6a.canonical.com:443/

# did not work + did not understand one of the steps

# create credentials and save the fields access and secret for later steps 
openstack ec2 credentials create



# configure MC using access + secret from openstack credentials 
 HTTP_PROXY="http://10.17.2.1:3128" HTTPS_PROXY="http://10.17.2.1:3128"  curl https://dl.min.io/client/mc/release/linux-amd64/mc \
  --create-dirs \
  -o $HOME/minio-binaries/mc

chmod +x $HOME/minio-binaries/mc
export PATH=$PATH:$HOME/minio-binaries/

mc --help
mc config host add my_project [https://radosgw.pc6a.canonical.com:443/](https://radosgw.pc6a.canonical.com/) b4ea84b3154e49f6ae1bd9fd07617c0d 215674a0fa64408389aa9cfa81d938a0

# now we need to create a bucket
sudo cp  vault-root-ca.pem /usr/local/share/ca-certificates/vault.crt

mc mb my-project/mybucket

# MY GUESS: you will need to generate a cert file from the pem file used from the novarc
openssl x509 -outform der -in vault-root-ca.pem -out vault-root-ca.crt

# use URL from earlier
[...]

PR #546, opened 2025-02-05
[DPE-6236] pbm tls

MiaAltieri:
## Issue
This solves the issue #525 

## Solution
We should save certificates to the trust store

## Testing
Tested by hand , by setting up [minio with TLS](https://gist.github.com/MiaAltieri/41a78253722cc018f2b82ae6de30da2e) and running various backup actions

### DIFF
--- a/lib/charms/mongodb/v1/mongodb_backups.py
+++ b/lib/charms/mongodb/v1/mongodb_backups.py
@@ -18,7 +18,7 @@ from typing import Dict, List, Optional, Union
 from charms.data_platform_libs.v0.s3 import CredentialsChangedEvent, S3Requirer
 from charms.mongodb.v1.helpers import current_pbm_op, process_pbm_status
 from charms.operator_libs_linux.v2 import snap
-from ops.charm import RelationJoinedEvent, RelationBrokenEvent
+from ops.charm import RelationBrokenEvent, RelationJoinedEvent
 from ops.framework import Object
 from ops.model import BlockedStatus, MaintenanceStatus, StatusBase, WaitingStatus
 from ops.pebble import ExecError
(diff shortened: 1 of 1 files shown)


phvalguima:
Hi @MiaAltieri, for my clarity, are you testing PBM vs. AWS S3? If so, then you are using their endpoint which is TLS.

The microceph action provides self-signed certs.

MiaAltieri:
@phvalguima - ah I should have looked at the minio work you had done more carefully! thank you for the heads up, I will remember this and update the message

Gu1nness:
@MiaAltieri We need also to have the same `write cert + update store` on the start event of the charm when we write the tls certificates on fs I think to ensure that we don't lose it accidentally upon restart since it's not written in the mongodb storage.

MiaAltieri:
@Gu1nness is that true even for VM? I thought this was only necessary for K8s?

Gu1nness:
> @Gu1nness is that true even for VM? I thought this was only necessary for K8s?

We do it on [VM](https://github.com/canonical/mongodb-operator/blob/6/edge/src/charm.py#L440)
And as we restart PBM right after, maybe we need to do it as well? But then it would also mean we have to `_set_pbm_config` actually? We should check to ensure that

MiaAltieri:
@Gu1nness - I think this conversation would be better for a call, but I still am sick and don't know if my throat can handle a conversation at the moment. We can also have a call on Monday :)

Until then, let me try to explain! 

First off, great find and thank you for pointing me to this and clearing my confusion. It is certainly correct that we need those tls files when running the start hook! But why?

Because when we run the start hook we aim to start the primary services of the charm (mongod+mongos)- when a charm scales its replicas they **will not** be able to start (or add replicas) if they do not have their internal membership certs. So in order to prevent this hook from repeatably failing we must provide the internal certs (especially since this hook sets `db_initialised`) - it helps with this race condition. i.e. this is a special case. Usually we should try where we can to honor the order of juju events.

But for s3-relations, we let all of these necessary configurations get set with the s3-relation-changed event - thereby honoring juju events. Note that TLS cert is not the only other configuration required for the pbm agent to start.

As for the file persisting on a container restart, it does persist on restart. :) 

[Storage](https://canonical-juju.readthedocs-hosted.com/en/latest/user/reference/storage/#storage) I believe is primary used so that you can save your data and transfer it to a new juju deployment, i.e. juju remove-application mongodb , juju deploy mongodb --storage old-storage). In this case we probably wouldn't want to put a configuration in the storage since it would mean that certain integrations are required. 

I hope that helps!

Gu1nness:
Very clear, thank you !

Gu1nness (dismissed):

Mehdi-Bendriss (commented):
Great fix, Thank you Mia! 
We should perhaps bump the libpatch?

Mehdi-Bendriss (approved):

Gu1nness (approved):

PR #553, opened 2025-04-30
[DPE-6955] Integrate Advanced Statuses in MongoDB VM

Gu1nness:
## Issue

We have developed a new Advanced Statuses library.
Let's use it in the charm

## Solution

Bump the mongo single kernel library to the latest version.

### DIFF
--- a/tests/integration/helpers.py
+++ b/tests/integration/helpers.py
@@ -450,7 +450,7 @@ async def check_all_units_blocked_with_status(
         ), f"unit {unit.name} not in blocked state, in {unit.workload_status.value}"
         if status:
             assert (
-                unit.workload_status.message == status
+                status in unit.workload_status.message
             ), f"unit {unit.name} not in blocked state, in {unit.workload_status.value}"
 
 
--- a/tests/integration/helpers.py
+++ b/tests/integration/helpers.py
@@ -472,6 +472,20 @@ async def wait_for_mongodb_units_blocked(
         await ops_test.model.set_config({hook_interval_key: old_interval})
 
 
+async def check_status_detail(ops_test: OpsTest, app_name: str, status: str, message: str) -> None:
+    """Checks that the first status returned by status-detail is the one expected."""
+    for unit in ops_test.model.applications[app_name].units:
+        action = await unit.run_action("status-detail")
+        action = await action.wait()
+        result = action.results["json-output"]
+
+        # juju messes up the string formatting here.
+        unit_statuses = json.loads(result["unit"].replace("'", '"'))
+
+        assert unit_statuses[0]["Status"].lower() == status
+        assert unit_statuses[0]["Message"] == message
+
+
 def is_relation_joined(ops_test: OpsTest, endpoint_one: str, endpoint_two: str) -> bool:
     """Check if a relation is joined.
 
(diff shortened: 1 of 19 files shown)


tests/integration/helpers.py
```diff
@@ -450,7 +450,7 @@ async def check_all_units_blocked_with_status(
         ), f"unit {unit.name} not in blocked state, in {unit.workload_status.value}"
         if status:
             assert (
-                unit.workload_status.message == status
+                status in unit.workload_status.message
```
Mehdi-Bendriss, line 453: Why are we checking on the inclusion instead of the equality?

tests/integration/sharding_tests/test_sharding.py
```diff
@@ -216,6 +228,7 @@ async def test_sharding(ops_test: OpsTest) -> None:
 
 @pytest.mark.runner(["self-hosted", "linux", "X64", "jammy", "large"])
 @pytest.mark.group(1)
+@pytest.mark.abort_on_fail
```
Mehdi-Bendriss, line 231: nice

tests/integration/sharding_tests/test_sharding_relations.py
```diff
@@ -264,7 +264,7 @@ async def test_shard_mongos_relation(ops_test: OpsTest) -> None:
     await wait_for_mongodb_units_blocked(
         ops_test,
         SHARD_ONE_APP_NAME,
-        status="Relation to mongos not supported, config role must be config-server",
+        status="Relation to mongos not supported",
```
Mehdi-Bendriss, line 271: it seems like we're losing a useful info here, by removing the second part of the status

tests/integration/sharding_tests/test_sharding_relations.py
```diff
@@ -289,7 +289,7 @@ async def test_shard_s3_relation(ops_test: OpsTest) -> None:
     await wait_for_mongodb_units_blocked(
         ops_test,
         SHARD_ONE_APP_NAME,
-        status="Relation to s3-integrator is not supported, config role must be config-server.",
+        status="Relation to s3-integrator is not support",
```
Mehdi-Bendriss, line 302: `supported` -- and Similar comment on the loss of info

Mehdi-Bendriss (commented):
Thanks Neha! I really like how this is taking form! I left a few comments

tests/integration/helpers.py
Gu1nness: We might have some other concurrent statuses so the string may or may not be extended with the extra info of additional statuses

tests/integration/sharding_tests/test_sharding_relations.py
Gu1nness: There are some concurrent statuses in that case iirc.
Maybe we should extend this function (`wait_for_mongodb_units_blocked`) to be smarted and parse `status-detail` action to get the exact status that we want?

Gu1nness: I added the status-detail check as we said

tests/integration/sharding_tests/test_sharding_relations.py
Gu1nness: I added the status-detail check as we said

tests/integration/helpers.py
Gu1nness: After this morning's discussion, I'll leave it as is and I added the status-detail check as we said

Mehdi-Bendriss (dismissed):

Mehdi-Bendriss (approved):

PR #566, opened 2025-08-04
Unify / Improve terraform modules

Mehdi-Bendriss:
## Issue
Terraform charm and product modules need to be unified and improved for both replicaset and sharded deployments.


terraform/charm/sharded/outputs.tf
```diff
@@ -0,0 +1,44 @@
...
+
+# Required integration endpoints
+output "requires" {
+  description = "Map of all \"requires\" endpoints"
+  value = {
+    sharding                  = "sharding"
+    certificates              = "certificates"
+    s3_credentials            = "s3-credentials"
+    ldap                      = "ldap"
+    ldap_certificate_transfer = "ldap-certificate-transfer"
```
Gu1nness, line 34: This is not yet on 6/stable, so I don't know if we should add it right after release ? Or keep it already ?

Mehdi-Bendriss: I think it's okay given we'll soon promote to stable and it's "just" in the outputs

Gu1nness (approved):
Thanks a lot for the rework!

PR #575, opened 2025-08-29
patch: remove the auto-delete config option

patriciareinoso:
## Issue
The `auto-delete` config option is no longer a desired config option in release 8
The terraform checks throw and ambiguous relation error.

## Solution
This PR removes the `auto-delete` config option and its logic.
Now, the databases are never dropped.
This prevents unintended data loss in case of misconfiguration.
Adds the endpoint to the certificates relation in the TF modules


terraform/product/replica_set/integrations.tf
```diff
@@ -27,9 +27,11 @@ resource "juju_integration" "mongodb_tls_same_model_integration" {
   model = var.mongodb.model
   application {
     name = var.mongodb.app_name
+    endpoint = "certificates"
```
Mehdi-Bendriss, line 30: nice, thank you

Mehdi-Bendriss (dismissed):

Gu1nness:
You also need to bump the library version to the new version containing your patch in your PR to remove the workflows for that option.

Gu1nness (approved):

Gu1nness (approved):

Mehdi-Bendriss (approved):

PR #577, opened 2025-09-26
patch: migrate from action-based password management to config-option-based passwords 

patriciareinoso:
This PR propagate the changes in https://github.com/canonical/mongo-single-kernel-library/pull/83 to the mongodb-operator.

Now the internal user passwords will be managed with a config option (system-users) instead of using actions.

### DIFF
--- a/actions.yaml
+++ b/actions.yaml
@@ -4,28 +4,6 @@
 get-primary:
   description: Report primary replica
 
-get-password:
-  description:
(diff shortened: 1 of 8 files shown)


Gu1nness (dismissed):

Mehdi-Bendriss (approved):

Gu1nness (approved):
Just fix the PR description to use mongodb-operator instead of mongodb-k8s-operator please

PR #669, opened 2026-04-20
fix: release workflow

Gu1nness:
Wrong parameter to workflow


sinclert-canonical (approved):
Optional improvement: quote the value, in case you need to target a value having a dot (`.`) sometime in the future. Otherwise, it will not parse correctly.

akram09 (approved):

PR #722, opened 2026-09-14
patch: (fix) mark optional endpoints

### LINKED ISSUE #708: Optional endpoints are not marked as optional
The Charm QA program run by @canonical/solutions-qa attempts to test every endpoint of every public charm on Charmhub. We rely on charm metadata to determine what is required to deploy a charm.

Edit: I'm leaving my original description below, but I think this issue should ultimately reflect that the metadata should be updated to reflect optionality for all optional endpoints.

Original description:
The `mongodb` charm offers/`provides` the `cluster` endpoint and `config-server` interface, but this endpoint is not marked as optional, and there are no publicly listed charms that `require` this endpoint, so our tests can't do anything.

1. Is the `cluster` endpoint required, or could it be marked optional in the metadata?
   ```
   provides:
     cluster:
       interface: config-server
       optional: true
   ```
2. Is there a private charm somewhere that does `require` the `cluster` endpoint, but that private charm should actually be made public?

patriciareinoso:
Fix https://github.com/canonical/mongodb-operator/issues/708


metadata.yaml
Gu1nness, line 86:
```suggestion
    limit: 1
    optional: true
```

Gu1nness (changes requested):

Gu1nness (approved):

